PR #4, opened 2017-10-23
factor in min option for maxIdle value

callmehiphop:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2685

- [x] Tests and linter pass


src/session-pool.js
```diff
@@ -59,7 +59,7 @@ function SessionPool(database, options) {
   options = options || {};
 
   this.database = database;
-  this.maxIdle = options.maxIdle || 1;
+  this.maxIdle = options.maxIdle || options.min || 1;
```
stephenplusplus: Is this a place where `Math.min` would help?

stephenplusplus: Or `max`?

stephenplusplus:
@lukesneeringer should we wait to merge this for any reason?

src/session-pool.js
lukesneeringer:
```
> Math.min(undefined, 1)
<- NaN
```

So, no. :-)

stephenplusplus:
```js
this.maxIdle = options.maxIdle /*0*/ || options.min /*10*/ || 1
// this.maxIdle = 10
```

Is this scenario realistic and assignment correct?

callmehiphop: @stephenplusplus I just pushed a commit that checks to see if `options.maxIdle` is a number first.

lukesneeringer:
Yeah, I want to merge repo-migration into it first so the tests will pass.

PR #12, opened 2017-10-26
use retry info 

callmehiphop:
Closes https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2673

- [x] Tests and linter pass

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -18,7 +18,7 @@
 
 var common = require('@google-cloud/common');
 var extend = require('extend');
-var grpc = require('grpc');
+var gax = require('google-gax');
 var is = require('is');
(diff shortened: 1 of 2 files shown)


src/transaction.js
```diff
@@ -18,7 +18,9 @@
 
 var common = require('@google-cloud/common');
 var extend = require('extend');
+var grpc = require('grpc');
```
stephenplusplus: `grpc` isn't our own dependency, so I don't think we should load it up through whichever other module brings it in.

callmehiphop: Doh, I hadn't even noticed that. I just added a commit to pull it from gax instead.

PR #29, opened 2017-11-06
use different pooling library

### LINKED ISSUE #13: KeepAlive does not work on SessionPool
cc @lukesneeringer @callmehiphop @stephenplusplus 
I am using the latest version of the node.js client. It seems that KeepAlive logic does not work.
To reproduce I wrote a simple code which set KeepAlive time to 1 minutes and then it makes a single query to ensure that the session is created. After that it just waits for a while on prompt. I would expect that after a few minutes I would see at least some ExecuteSql calls (which is how session is kept alive). I enabled grpc tracing and all I saw was a single CreateSession followed by a single ExecuteStreamingSql (which was due to the query I made) and nothing else.
To reproduce turn on grpc tracing
```
export GRPC_TRACE=http
export GRPC_VERBOSITY=INFO
```

Then run the sample code given below and direct the grpc logs to a file
```
node keepalive.js 2> grpc.log
```

This will make a query and display a prompt. Wait for few minutes and take a look at grpc.log.
This is the code
```

'use strict';
  const Spanner = require('@google-cloud/spanner');
  const spanner = Spanner();
  const instanceId = 'myinstance';
  const databaseId = 'mydb';
  const instance = spanner.instance(instanceId);
  const database = instance.database(databaseId);
  const query = {
    sql: 'SELECT 1 as c'
  };
  database.run(query)
    .then((results) => {
      const rows = results[0];
      rows.forEach((row) => {
        const json = row.toJSON();
        console.log(`${json.c.value}`);
      });
      var prompt = require('prompt');
      prompt.start();
      prompt.get(['done'], function(err, result){
        database.close();
        console.log('Bye');
      });
    });
```

### LINKED ISSUE #19: Cloud Spanner Session Pool should rate limit initial session creations.
_From @vkedia on June 14, 2017 20:25_

Cloud Spanner Session pool upon construction would start creating sessions eagerly upto the minimum number of sessions. It will essentially fire all the create requests in parallel. We should instead rate limit the session creation so that it creates at most X number of sessions at a time (X=10 would be a reasonable value).
I realize that this behavior is manged by the underlying pooling library which does not provide a way to rate limit. One option can be to set `autostart=false` so that the pooling library does not eagerly create sessions. Then in the SessionPool we ourselves can call `pool.acquire` X times so that X number of sessions would be eagerly created.
Another option can be to file a feature request against the pooling library to support this use case.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2388_

### LINKED ISSUE #25: Cloud Spanner: Session being deleted and created on each call
_From @vkedia on October 20, 2017 18:45_

A customer reported seeing higher latencies on ExecuteStreamingSql. Debugging further they realised that after every ExecuteStreamingSql call, a DeleteSession and CreateSession is being called which is unexpected.
This is the information from customer:

Our App is currently deployed with-in GCP in a Docker container running Node V8 and we are using Spanner module 0.7.1
When establishing a connection to Spanner we are using a singleton pattern where we retain a global reference to the spanner, instance and database objects returned by the Spanner API.
 
When requesting the database connection from the spanner instance, we are using the following connection options:

```es6
database = spannerInstance.database(‘database’, {min: 20, keepAlive: 10});
```

After we enabled Google Stackdriver Trace we noticed that on essentially every request made to our deployed service that involved the execution of a Spanner query:

```es6
database.run(‘select stuff from table thing where column = value’).then(results => {
  …do stuff…
}).catch(err=>{
  ..uh oh..
});
 ```

That we see the following gRPC time spans logged in the following order:
```
grpc:/google.spanner.v1.Spanner/ExecuteStreamingSql
grpc:/google.spanner.v1.Spanner/DeleteSession
grpc:/google.spanner.v1.Spanner/CreateSession
```

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2685_

### PR DESCRIPTION
**Pull request by callmehiphop**
Fixes #25 
Fixes #19
Fixes #13

- [x] Tests and linter pass

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -72,6 +71,7 @@
     "lodash.snakecase": "^4.1.1",
     "merge-stream": "^1.0.1",
     "methmeth": "^1.1.0",
+    "pewl": "^0.1.0",
```
vkedia: where is this library?

callmehiphop: This is something I put together specifically for Spanner - https://github.com/callmehiphop/pewl.

vkedia: I see. Why not put it here itself?

callmehiphop: I wasn't able to find a pooling library that met all our requirements so I decided it might be beneficial to the Node community if I opened sourced it in case other developers were searching for similar functionality.

lukesneeringer: Can we put it in `nodejs-spanner` for now, please?

callmehiphop: @lukesneeringer would giving you and @vkedia ownership of the repo suffice instead?

vkedia: I think it is more than ownership. Session pooling is a pretty core piece of cloud spanner library. It will be much more flexible if that library is specific to spanner and not open to the world to use. That will give us much more leeway in changing it/enhancing it as per our requirements.

vkedia: Any updates here?

callmehiphop: Between the request to move the pooling logic to this repo and a few work re-prioritizations I've been a little set back. This is my main focus now and I should have something for you to review later today.

callmehiphop: @vkedia I've made a sizable commit with the pooling refactor, I still need to write tests but if you'd be kind enough to review while I'm working on them, that would be awesome!

src/session-pool.js
```diff
@@ -76,189 +88,76 @@ function SessionPool(database, options) {
...
-
-  return pool;
-};
+util.inherits(SessionPool, EventEmitter);
 
 /**
- * Build `genericPool` options from the user's options.
- *
- * @private
+ * Closes and the pool and destroys all sessions within it.
```
vkedia: "Closes the pool"

src/session-pool.js
```diff
@@ -16,31 +16,50 @@
...
  * @property {number} [max=100] Maximum number of resources to create at any
  *     given time.
  * @property {number} [maxIdle=1] Maximum number of idle resources to keep in
  *     the pool at any given time.
  * @property {number} [min=0] Minimum number of resources to keep in the pool at
  *     any given time.
- * @property {number} [keepAlive=59] How often to ping idle sessions, in
+ * @property {number} [keepAlive=50] How often to ping idle sessions, in
  *     minutes. Must be less than 1 hour.
  * @property {number} [writes=0] Pre-allocate transactions for the number of
```
vkedia: instead of fixed number of sessions, can this be made into a percentage of sessions they want to be prepared for write. That ways it will scale with how many open sessions there are.

src/session-pool.js
```diff
@@ -435,252 +366,370 @@ SessionPool.prototype.requestStream = function(config) {
...
+  var maxIdle = this.options.maxIdle;
+  var min = this.options.min;
 
-    var transaction = self.createTransaction_(session);
+  var idle = this.getIdleSessions_();
+  var count = idle.length;
 
-    session.isWriteSession_ = true;
-    session.transaction_ = transaction;
+  while (count-- > maxIdle && count >= min) {
```
vkedia: I think you should compare size() to min?

src/session-pool.js
```diff
@@ -435,252 +366,370 @@ SessionPool.prototype.requestStream = function(config) {
...
-    return;
   }
 
-  // Use a regular read session.
-  if (this.pool.free) {
-    this.getSession(callback);
-    return;
+  var session;
+
+  for (session of this.available_) {
```
vkedia: why not maintain two lists, one for read sessions and one for read write sessions? Then you do not need to iterate through the list. You can just pop the head of the relevant list if non empty otherwise pop the other list.

src/session-pool.js
```diff
@@ -435,252 +366,370 @@ SessionPool.prototype.requestStream = function(config) {
...
-      return self.writePool.destroy(session);
-    });
+  var sessions = this.getIdleSessions_();
+  var requests = sessions.map(function(session) {
+    return self.requestQueue_
+      .add(function() {
+        return session.keepAlive();
+      })
+      .catch(function(err) {
+        self.destroySession_(session);
```
vkedia: You should only destroy is the error is "NOT_FOUND"

src/session-pool.js
```diff
@@ -435,252 +366,370 @@ SessionPool.prototype.requestStream = function(config) {
...
 
-  var needsKeepAlive = this.evictor.evict(config, resource, available);
+  return transaction.begin().then(
+    function() {
+      session.txn = transaction;
+      return transaction;
+    },
+    function(err) {
+      transaction.end();
+      self.destroySession_(session);
```
vkedia: only destroy if the error is "NOT_FOUND", otherwise put the session back in the read only state.

src/session-pool.js
```diff
@@ -304,57 +192,100 @@ SessionPool.prototype.getSession = function(callback) {
...
+  this.emit('open');
+
+  var neededReadWrite = this.options.writes;
+  var neededReadOnly = this.options.min - neededReadWrite;
+
+  var requests = [];
+  var i;
+
+  for (i = 0; i < neededReadWrite; i++) {
+    requests.push(this.createSessionInBackground_(READWRITE));
```
vkedia: what happens if these creations fail? For eg if the user gave an incorrect database name, all session creations would fail.

src/session-pool.js
```diff
@@ -435,252 +366,370 @@ SessionPool.prototype.requestStream = function(config) {
...
-        return;
-      }
+/**
+ * Attempts to get a session of a specific type. If the type is unavailable it
+ * may try to use a different type.
+ *
+ * @private
+ *
+ * @param {string} type The desired session type.
+ * @returns {Promise<Session>}
```
vkedia: if we are never able to create a session, will this promise ever be resolved?

src/session-pool.js
```diff
@@ -435,252 +366,370 @@ SessionPool.prototype.requestStream = function(config) {
...
-      return self.writePool.release(session);
-    })
-    .then(null, function() {
-      return self.writePool.destroy(session);
-    });
+  var sessions = this.getIdleSessions_();
+  var requests = sessions.map(function(session) {
+    return self.requestQueue_
+      .add(function() {
+        return session.keepAlive();
```
vkedia: we should rest session.txn when this returns. That transaction is no longer valid because for doing keep alive a new implicit transaction was started.

test/session-pool.js
```diff
@@ -17,734 +17,489 @@
...
-      var error = new Error('Error.');
+      assert.strictEqual(callCounts.readonly, min - writes);
+      assert.strictEqual(callCounts.readwrite, writes);
+    });
 
-      function callback(err) {
-        assert.strictEqual(err, error);
-        done();
-      }
+    it('should settle once all sessions are created', function() {
```
vkedia: Does this mean that session pool is only usable after all of the min sessions have been created. If yes then that is not idle. It should be usable immediately and any getSession would block till there is any session available.

src/session-pool.js
callmehiphop: In the event of a create failure we would emit an error that the Database object would catch and then re-emit said error to the user.

src/session-pool.js
callmehiphop: But wouldn't just about any error mean the keep alive request failed and the session is likely to expire on the server?

src/session-pool.js
callmehiphop: We wrap any requests in a `race_` function which essentially races against the pool closing and a timeout occurring. So the only way it would not resolve is if the pool never closed, the user chose to opt out of timing out and a session never became available.

test/session-pool.js
callmehiphop: Nope! We don't want for the pool when using a Database object, they can use it immediately and unless configured to fail, it would wait for a session to free up.

src/session-pool.js
callmehiphop: Good call, I've pushed a commit that does this.

src/session-pool.js
callmehiphop: Done! PTAL

src/session-pool.js
vkedia: This is still pending.

src/session-pool.js
vkedia: I see. But what would happen if they call something like `run_in_transaction` on this database? Will there callback never get invoked or will it get invoked with an error?

src/session-pool.js
vkedia: This is still pending.

src/session-pool.js
callmehiphop: It should get invoked with an error, basically our flow looks like this

* Is there a session available?
  * Yes! Return it!
* Did the user disable waiting for a session?
  * Yes! Return an error.
* Is the pool full?
  * No! Create a session and return that. If an error occurs while creating, we'll return this error to the user.
  * Yes! Set a timeout (if one is configured) and race between the pool closing, the timeout occurring and a session being released back to the pool.

vkedia: Perfect. Thanks for clarifying.

alexander-fenster:
@callmehiphop Can you please take a look at two failing tests:
      1) should evict the sessions
      2) should respect the min value

vkedia:
Any updates here?

callmehiphop:
@vkedia sorry, I was making the updates you had mentioned about switching a session state if we're unable to begin a transaction - it ended up being a little more in depth than I originally thought. But I think we should be good for a final review if you'd like to take another look.

vkedia:
Ok. This looks ok. Can you also do some manual testing to make sure that keep alives are being called, idle sessions are being closed etc etcYou can turn on grpc logging like I had mentioned in my  issue report and then see what all rpcs are being made.
You can do something like
min=1
maxIdle=2
max=5
keepAliveInterval = 5 minutes

Then you can create a database object and do not make any query. 1 session should still be created. After that you can send a burst of 5 queries in parallel, so 5 session should be created. Then do not do anything for a while. 3 sessions should be deleted. 2 sessions should always remain in the pool even if you are not doing anything and every 5 mintues keep alive should be sent on them.

vkedia:
Can we try to get this merged in next day or two?

callmehiphop:
@vkedia sure can, I just finished testing your plan and I'm happy to report everything functions as expected.

vkedia:
That is good to know. Thanks for testing this out.

PR #54, opened 2017-12-06
add benchmarks

callmehiphop:


benchmark/workload.js
```diff
@@ -0,0 +1,121 @@
...
+    const end = timeSpan();
+
+    for (let i = 0; i < operationCount; i++) {
+      let randomWeight = Math.random() * this.totalWeight;
+
+      this.weights.forEach((weight, j) => {
+        const operation = this.operations[j];
+
+        if (randomWeight <= weight) {
+          this.queue.add(() => this.runOperation(operation));
```
vkedia: you should break after this. Each iteration of the for loop should do exactly one operation based on the random weight.

benchmark/workload.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  read() {
+    const table = this.options.get('table');
+    const id = this.getRandomKey();
+    const query = `SELECT u.* FROM ${table} u WHERE u.id="${id}"`;
+
+    return this.database.getTransaction({ readonly: true }).then(data => {
+      const txn = data[0];
+      return txn.run(query).then(() => txn.end());
```
vkedia: when does the promise returned by `txn.run` resolve? Does it resolve when all the data has been returned?

benchmark/workload.js
```diff
@@ -0,0 +1,121 @@
...
+    });
+  }
+
+  update() {
+    const table = this.database.table(this.options.get('table'));
+    const id = this.getRandomKey();
+    const field = `field${random(9)}`;
+    const value = crypto.randomBytes(100).toString('hex');
+
+    return table.update({ id, [field]: value });
```
vkedia: does this use a single use transaction?

benchmark/ycsb.js
```diff
@@ -0,0 +1,134 @@
...
+
+const Spanner = require('../src');
+const Workload = require('./workload');
+
+require('yargs')
+  .version(false)
+  .strict()
+  .command(
+    'run [args]',
+    'Run the workload',
```
vkedia: we also want to test with multiple instance of workload running in parallel. So can you add a parameter for that and based on that start n numbers of workload?

benchmark/workload.js
callmehiphop: Yep! Under the hood we use the streaming RPC, this particular method calls that rpc and buffers the results, the promise won't resolve until it has completely finished.

benchmark/workload.js
callmehiphop: It does - should it not be?

benchmark/workload.js
vkedia: ok. Does this create a single use transaction? If not then we should use the API which creates a single use transaction since that is what we are doing in other language benchmarks.

benchmark/workload.js
vkedia: For writes, we want to use a multi use transaction, so that the BeginTransaction path is exercised.

vkedia:
What is pending here?

benchmark/workload.js
callmehiphop: Done!

benchmark/workload.js
callmehiphop: Done!

benchmark/workload.js
callmehiphop: Done!

benchmark/ycsb.js
callmehiphop: Done!

callmehiphop:
@vkedia I think we should be good to go if you'd be kind enough to review.

PR #56, opened 2017-12-06
do not promisify txn crud methods

### LINKED ISSUE #55: Transaction.insert generates unhandled promise rejections
#### Environment details

  - OS: hmm... Google Cloud Platform?
  - Node.js version: 8.9.x
  - npm version: yarn 1.1
  - @google-cloud/spanner version: 0.8.1

#### Steps to reproduce

  1. Create a read-write transaction
  2. Use transaction.insert, but insert data which do not match the destination table's schema
  3. Run transaction.commit()

This will generate an unhandled promise rejection, which does **not** come from `commit()`.

My code (simplified) looks like this:
```js
try {
   transaction.insert('tableName', {fakeData: 'rejected!'});
   await transaction.commit();
} catch(error){
  ...
}
```

[According to the docs](https://cloud.google.com/nodejs/docs/reference/spanner/0.8.x/Transaction#insert), `insert` does not seem to accept a callback. However, it [looks like it accepts a callback parameter in the source code](https://github.com/googleapis/nodejs-spanner/blob/master/src/transaction-request.js#L450).

So I tried adding a `catch` after the insert:
```js
transaction.insert('tableName', {fakeDate: 'rejected!'}).catch(...); // No unhandled promise rejection
```
This does seem to catch the error. However, the rest of the code is still executed synchronously.
So I tried this:
```js
try {
   await transaction.insert('tableName', {fakeData: 'rejected!'}); // Use await instead of catch
   await transaction.commit();
} catch(error){
  ...
}
```

The error is caught, but if there's no error, await (or the associated promise) never resolves!
```js
try {
   await transaction.insert('tableName', {validData: 'should be good!'}); // hangs forever
   await transaction.commit();
} catch(error){
  ...
}
// Unhandled promise rejection
```

It seems like:
- `insert()` in fact accepts a callback
- A `resolve` is missing somewhere if the insert is a success

Thanks for your help!

callmehiphop:
Fixes #55

- [x] Tests and linter pass

### DIFF
--- a/system-test/spanner.js
+++ b/system-test/spanner.js
@@ -3070,6 +3070,34 @@ describe('Spanner', function() {
     });
 
     describe('read/write', function() {
+      it('should throw an error for mismatched columns', function(done) {
+        database.runTransaction(function(err, transaction) {
+          assert.ifError(err);
+
+          var rows = [
+            {
+              Key: 'k1',
+              StringValue: 'hi',
+            },
+            {
+              Key: 'k2',
+              NumberValue: 4,
+            },
+          ];
+
+          var expectedMessage = [
+            'Row at index 0 does not contain the correct number of columns.',
+            'Missing columns: ["NumberValue"]',
+          ].join('\n\n');
+
+          assert.throws(function() {
+            transaction.insert(table.name, rows);
+          }, expectedMessage);
+
+          transaction.end(done);
+        });
+      });
+
       it('should commit a transaction', function(done) {
         database.runTransaction(function(err, transaction) {
           assert.ifError(err);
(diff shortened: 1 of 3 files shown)


WaldoJeffers:
If possible, this should be merged & released as soon as possible, as it fixes a critical bug :)

vkedia:
Is it possible to write a test which reproduces the scenario mentioned by @WaldoJeffers and assert that is passes (and it failed without this fix).

callmehiphop:
@vkedia I've created a new system test if you'd be kind enough to review.

system-test/spanner.js
```diff
@@ -3070,6 +3070,34 @@ describe('Spanner', function() {
     });
 
     describe('read/write', function() {
+      it('should throw an error for mismatched columns', function(done) {
```
vkedia: So this test would fail without the fix introduced in this PR?

callmehiphop: Correct, it would return a rejected promise instead of throwing an error.

PR #74, opened 2017-12-18
track session leaks

### LINKED ISSUE #18: Spanner library should have a way to track leaked sessions.
_From @vkedia on May 10, 2017 17:56_

Due to a bug in user code or client library code, it is possible that a session is checked out from the pool but never released. We should make it easier for users to be able to detect such leaks and debug where in there code the leak occurred. One way can be to record a trace in a list or a map whenever a session is checked out and remove that trace when the session is released back to the pool. User should be able to look at this list and figure out the location in the code where all the unreleased sessions were checked out at. We can also log these stack traces when the pool is closed.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2292_

callmehiphop:
Fixes #18

- [x] Tests and linter pass

### DIFF
--- a/src/session-pool.js
+++ b/src/session-pool.js
@@ -21,6 +21,7 @@ var delay = require('delay');
 var extend = require('extend');
 var is = require('is');
 var PQueue = require('p-queue');
+var stackTrace = require('stack-trace');
 var streamEvents = require('stream-events');
 var through = require('through2');
 var util = require('util');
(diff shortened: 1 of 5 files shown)


src/session-pool.js
```diff
@@ -112,11 +113,26 @@ function SessionPool(database, options) {
...
 }
 
 util.inherits(SessionPool, EventEmitter);
 
+/**
+ *
+ */
+SessionPool.formatTrace_ = function(trace) {
+  var formatted = trace.slice(2).map(function(t) {
+    return `  at ${
```
stephenplusplus: I checked a native-thrown error, and it uses 4 spaces:

```js
$ node -e "throw new Error('no')"
[eval]:1
throw new Error('no')
^

Error: no
    at [eval]:1:7
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at Object.runInThisContext (vm.js:152:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:641:30)
    at evalScript (bootstrap_node.js:470:27)
    at startup (bootstrap_node.js:167:9)
    at bootstrap_node.js:613:3
```

Not sure if that's just for me / my terminal / my Node version, etc. Can you see what you get If you get the same, and then if you do, let's follow suit.

callmehiphop: SGTM

PR #75, opened 2017-12-18
increase transaction timeout

callmehiphop:
Closes https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2644

- [x] Tests and linter pass

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -125,11 +125,18 @@ function Transaction(session, options) {
    */
   this.transaction = true;
 
+  options = options || {};
+
   this.queuedMutations_ = [];
   this.runFn_ = null;
 
-  this.timeout_ = DEFAULT_TRANSACTION_TIMEOUT;
   this.beginTime_ = null;
+  this.timeout_ = DEFAULT_TRANSACTION_TIMEOUT;
+
+  if (options.timeout) {
+    this.timeout_ = options.timeout;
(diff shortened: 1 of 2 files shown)


src/transaction.js
```diff
@@ -125,11 +125,18 @@ function Transaction(session, options) {
...
+  options = options || {};
+
   this.queuedMutations_ = [];
   this.runFn_ = null;
 
-  this.timeout_ = DEFAULT_TRANSACTION_TIMEOUT;
   this.beginTime_ = null;
+  this.timeout_ = DEFAULT_TRANSACTION_TIMEOUT;
+
+  if (options.timeout) {
```
stephenplusplus: Will anyone ever want `0`?

src/transaction.js
```diff
@@ -125,11 +125,18 @@ function Transaction(session, options) {
...
+    this.timeout_ = options.timeout;
+    delete options.timeout;
```
stephenplusplus: We should probably not delete the user-held property (unless `options` is a cloned object already?)

src/transaction.js
callmehiphop: I don't think so, but I've added a `is.number()` just to be safe.

src/transaction.js
callmehiphop: Good call, I've made an update PTAL

PR #79, opened 2018-01-02
refactor acquire queuing

### LINKED ISSUE #78: SessionPool acquireQueue concurrency limit causes sessions to pile up under load
#### Environment details

  - OS: MacOS
  - Node.js version: 9.2.0
  - @google-cloud/spanner version: 0.10.0

#### Steps to reproduce

1. Run a simple load test at a rate that fills up the pool of available sessions (40 req/s was enough for me, with current session pool settings)
2. After a little while you'll start to see requests taking a disproportionate amount of time (dozens of seconds) to respond, or even timeouts.

Not sure I have the right analysis of what is happening but here's what I understand:

- With the current session pool option defaults, the pool starts with 0 (min) open sessions, so it fills up pretty quickly and the session pool starts acquiring new sessions (https://github.com/googleapis/nodejs-spanner/blob/master/src/session-pool.js#L660)

- `requestQueue_` has a configurable rate limit with a sensible default, which is fine, but calls to `createSession` are queued in `acquireQueue_` (https://github.com/googleapis/nodejs-spanner/blob/master/src/session-pool.js#L778), which has a limit of 1 concurrent operation.

- So in effect, at load, `acquireQueue_` seems to become a bottleneck, where requests to create new sessions pile up. 

What's worrying is that queued `createSession` calls stay there piling up, even when previously busy sessions are now available. (Can even become a memory leak problem)

The session pool will wait and resolve each session creation one at a time, and because creating a new session takes time, clearing the queue takes a very long time (assuming request load has dropped, otherwise it will never get cleared), causing a good amount of requests to timeout. 

When I tested, I couldn't get over **30 req/s**, after which I started seing connection timeouts and requests responding after several dozen seconds

After setting `concurrency` for the `acquireQueue_` to 10, I could get the load to about **80 req/s**.
I suspect increasing `min` and `acquireQueue_` concurrency would allow to go even higher.

Hope it helps !

callmehiphop:
Fixes #78

- [x] Tests and linter pass

### DIFF
--- a/src/session-pool.js
+++ b/src/session-pool.js
@@ -622,7 +622,12 @@ SessionPool.prototype.getIdleSessions_ = function() {
 };
 
 /**
+ * Grabs the next available session.
  *
+ * @private
+ *
+ * @param {string} type The desired session type.
+ * @returns {Promise<Session>}
  */
 SessionPool.prototype.getNextAvailableSession_ = function(type) {
   var self = this;
@@ -670,9 +675,11 @@ SessionPool.prototype.getSession_ = function(type) {
     return Promise.reject(new Error('No resources available.'));
   }
 
-  var promises = [self.waitForNextAvailable_()];
+  var promises = [this.waitForNextAvailable_(type)];
+  var shouldCreate = !this.isFull() &&
+    available + this.pendingCreates_ < acquires + 1
 
-  if (!this.isFull() && available + this.pendingCreates_ < acquires + 1) {
+  if (shouldCreate) {
     var createPromise = new Promise(function(resolve, reject) {
       self.createSession_(type).then(function() {
         self.emit('available');
@@ -682,9 +689,7 @@ SessionPool.prototype.getSession_ = function(type) {
     promises.push(createPromise);
   }
 
-  return this.race_(promises).then(function() {
-    return self.getNextAvailableSession_(type);
-  });
+  return this.race_(promises);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


vkedia:
I have to admin I do not understand how this is fixing the issue. Can you please explain?
Also can there be some unit test which would demonstrate the original issue and which passes with this fix?

callmehiphop:
I think the major difference is that before if no sessions were available we would wait on a new one to be created and would not listen for sessions being freed up. If a session was created, all acquires would have to wait for that session to finish being created even if sessions were being freed up by other means. The big difference now is observing a race between creating a session and one being freed up.

> Also can there be some unit test which would demonstrate the original issue and which passes with this fix?

Absolutely, I was waiting to unit test this once https://github.com/googleapis/nodejs-spanner/pull/54 was merged and I could benchmark the changes.

vkedia:
Thanks for the explanation. From the original report it also seems that we only create one session at a time. Can we change that as well?We can set it to something like 10 by default and give users option to configure that.

callmehiphop:
That was a side effect of bottlenecking acquires with creating sessions, so that should also be resolved with this.

vkedia:
So now we would be creating multiple sessions in parallel?

callmehiphop:
@vkedia yep!

vkedia:
Ok. What about the sessions which are being prepared for write? Do you also race with those so that if one of those get ready you can return those to a waiting request?

callmehiphop:
@vkedia yes, basically when a session is requested and none are available we will listen for an `available` event that is used internally. The available event fires when:

* A session is created (if it is a write session, we wait until after the txn is ready)
* A session is released into the pool
  * Write sessions are not completely released until a new transaction is prepared

vkedia:
Ok. makes sense. I will let @stephenplusplus or @alexander-fenster review the code. Overall approach looks good to me.

callmehiphop:
@vkedia also I've been comparing against the benchmarks and I'm seeing some considerable gains. Do you know what kind of numbers I should expect/hope to see when I run them?

vkedia:
If you can share the numbers before and after, I can take a look.

PacciniBruno:
This is great ! :) 

Just re-ran my tests. It works flawlessly since https://github.com/googleapis/nodejs-spanner/pull/79/commits/7ef84b1f6d01dedf001199bf26c71879aa5b07db 
This should be a huge step forward in terms of performance. Really excited to see this released.

Also very happy to see how quick you are at answering to customer's issues.

stephenplusplus:
@callmehiphop is this safe to merge?

vkedia:
I think this can be merged. We can run the benchmark after merging.

callmehiphop:
@stephenplusplus @vkedia yes this should be safe to merge in that case.

PR #94, opened 2018-01-16
Support toJSON() on read methods.

### LINKED ISSUE #27: Spanner table.read() returns a list of column name-value pairs. Why not an object?
_From @jgeewax on March 17, 2017 15:52_

The response of calling `table.read(query)` is a list of rows, but the row isn't a native object, instead it's a key-value pair-style object:

```js
[ [ { name: 'id', value: { [Number: 1] value: '1' } } ],
 [ { name: 'id', value: { [Number: 2] value: '2' } } ],
 [ { name: 'id', value: { [Number: 3] value: '3' } } ]
```

Why wouldn't we match these up as a proper object? Since we don't customers end up having to write a helper-snippet just so they can pass around a useful object... That is, it'd be nice if the rows returned were...

```js
[ {id: '1'}, {'id': '2'}, {'id': '3'} ]
```

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2103_

stephenplusplus:
Fixes #27 

Provide `options.toJSON = true` to `database.run`, `table.read`, and `table.createReadStream`, and all returned rows will already have had `toJSON` called on them.

```js
table.read({
  keys: [...],
  columns: [...],
  toJSON: true
}, function(err, rows) {
  var row = rows[0]
  // row.Name = 'value'
})
```
```js
table.createReadStream({
    keys: [...],
    columns: [...],
    toJSON: true
  })
  .on('data', row => {
    // row.Name = 'value'
  })
```

`row.toJSON()` supports options, such as `wrapNumbers`. To get these through, the user has to provide `toJSONOption: {...}` alongside `toJSON: true`.

### DIFF
--- a/src/partial-result-stream.js
+++ b/src/partial-result-stream.js
@@ -19,6 +19,8 @@
 var checkpointStream = require('checkpoint-stream');
 var chunk = require('lodash.chunk');
 var eventsIntercept = require('events-intercept');
+var exec = require('methmeth');
+var extend = require('extend');
 var is = require('is');
 var mergeStream = require('merge-stream');
 var split = require('split-array-stream');
@@ -43,9 +45,11 @@ var RowBuilder = require('./row-builder.js');
  *     receive one argument, `resumeToken`, which should be used however is
  *     necessary to send to the API for additional requests.
  */
-function partialResultStream(requestFn) {
+function partialResultStream(requestFn, options) {
   var lastResumeToken;
   var activeRequestStream;
+
+  options = extend({toJSON: false}, options);
 
   // mergeStream allows multiple streams to be connected into one. This is good
   // if we need to retry a request and pipe more data to the user's stream.
@@ -103,6 +107,10 @@ function partialResultStream(requestFn) {
       }
 
       rowChunks = [];
+
+      if (options.toJSON) {
+        formattedRows = formattedRows.map(exec('toJSON'));
+      }
 
       split(formattedRows, userStream, function() {
         next();
--- a/src/partial-result-stream.js
+++ b/src/partial-result-stream.js
@@ -109,7 +109,9 @@ function partialResultStream(requestFn, options) {
       rowChunks = [];
 
       if (options.toJSON) {
-        formattedRows = formattedRows.map(exec('toJSON'));
+        formattedRows = formattedRows.map(
+          exec('toJSON', options.toJSONOptions)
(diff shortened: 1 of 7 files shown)


stephenplusplus:
@WaldoJeffers want to take a look? :)

src/partial-result-stream.js
```diff
@@ -104,6 +108,10 @@ function partialResultStream(requestFn) {
 
       rowChunks = [];
 
+      if (options.toJSON) {
+        formattedRows = formattedRows.map(exec('toJSON'));
```
WaldoJeffers: Although having `toJSON` available on `row` instances is great, having a static `toJSON` method would be cool too, since it's not dependent on the row's content. It would save you the trouble of using the `methmeth` library 😉 `formattedRows.map(Spanner.toJSON)`

WaldoJeffers:
As for passing arguments to `toJSON` from `table.createReadStream({...})`, I can't think of anything other than:
- accepting a `toJSONoptions`: `table.createReadStream({ toJSON: true, toJSONoptions: {...} })`
- making toJSON accept a function instead of a Boolean `table.createReadStream({ toJSON: row => toJSON(row, {...}) })`, but the downside is you always have to a function even if you're interested in giving it custom options

walshie4:
Didn't dig in too deeply to see if `run` would end up using the modified `partial-result-stream`, but it looks like `database.run` would _not_ support this option per the current state of this PR.  Is that correct?

stephenplusplus:
I've added `options.toJSON` and `options.toJSONOptions` to:

- `database.run()` (thanks, @walshie4!)
- `table.createReadStream()`
- `table.run()`

Tests added. Please take another look!

src/partial-result-stream.js
stephenplusplus: I tried going that route, but now that `options.toJSONOptions` is supported, it forced an inline function to wrap the `toJSON()` call anyway. So, now we have an easier to look at:

```js
formattedRows.map(exec('toJSON', options.toJSONOptions));
```

stephenplusplus:
@alexander-fenster would you mind taking another look?

stephenplusplus:
Since we haven't released this yet, I have a PR to rename `query.toJSON` to `query.json` and `query.toJSONOptions` to `query.jsonOptions`. Please take a look!: #122

PR #97, opened 2018-01-17
Changes RowBuilder.formatValue to return null in case value is 'NULL_VALUE'

### LINKED ISSUE #95: RowBuilder.formatValue breaks if an array field has NULL_VALUE
When selecting an Array type field with null value `RowBuilder.formatValue` breaks on line 64 throwing the exception:
```
TypeError: value.map is not a function
    at Function.RowBuilder.formatValue (/var/app/node_modules/@google-cloud/spanner/src/row-builder.js:65:18)
```
The code:
``` nodejs
RowBuilder.formatValue = function(field, value) {
  if (field.code === 'ARRAY') {
    // Execution enters here and value is equals 'NULL_VALUE' throwing an Exception
    return value.map(function(value) {
      return RowBuilder.formatValue(field.arrayElementType, value);
    });
  }

  if (field.code !== 'STRUCT') {
    return value;
  }

  return field.structType.fields.reduce(function(struct, field, index) {
    struct[field.name] = RowBuilder.formatValue(field.type, value[index]);
    return struct;
  }, {});
};
```

The workaround I'm using locally is:

```
if (value == 'NULL_VALUE') {
  return RowBuilder.formatValue(field.arrayElementType, null);
}
```

This way, the field serializes as `'field_name': null` which I think is the expected behavior.

#### Environment details

  - OS: Linux 14.04.1-Ubuntu
  - Node.js version: v8.9.4
  - npm version: 5.6.0
  - @google-cloud/spanner version: 1.0.0

#### Steps to reproduce
  1. Create table with ARRAY type column
  2. Insert record with null value in this field
  3. Run a query on method `database.run(query)` selecting this field

hugoluchessi:
Fixes #95

Simple validation. In case Rowbuilder.formatValue receives 'NULL_VALUE' as value, which applies to all types, return null as value for that field.
 
# Tests
Unit test were also included in this PR

### DIFF
--- a/src/row-builder.js
+++ b/src/row-builder.js
@@ -61,6 +61,10 @@ RowBuilder.getValue = function(obj) {
  * object.
  */
 RowBuilder.formatValue = function(field, value) {
+  if (value == 'NULL_VALUE') {
+    return null;
+  }
+  
   if (field.code === 'ARRAY') {
     return value.map(function(value) {
       return RowBuilder.formatValue(field.arrayElementType, value);
--- a/src/row-builder.js
+++ b/src/row-builder.js
@@ -61,10 +61,10 @@ RowBuilder.getValue = function(obj) {
  * object.
  */
 RowBuilder.formatValue = function(field, value) {
-  if (value == 'NULL_VALUE') {
+  if (value === 'NULL_VALUE') {
     return null;
   }
-  
+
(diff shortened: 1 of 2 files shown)


callmehiphop:
@hugoluchessi awesome, thanks for this!

hugoluchessi:
My pleasure!

PR #122, opened 2018-02-15
Rename `query.toJSON` to `query.json`.

stephenplusplus:
The original PR, #94, made a poor choice in variable naming. We haven't released to npm, so we'd like to change the name before it's permanent.

Instead of:

```js
table.read({
  keys: [...],
  columns: [...],
  toJSON: true
}, function(err, rows) {
  var row = rows[0]
  // row.Name = 'value'
})
```

This seems better:

```js
table.read({
  keys: [...],
  columns: [...],
  json: true // <--
}, function(err, rows) {
  var row = rows[0]
  // row.Name = 'value'
})
```

### DIFF
--- a/src/database.js
+++ b/src/database.js
@@ -648,9 +648,9 @@ Database.prototype.getTransaction = function(options, callback) {
  *     // Error handling omitted.
  *   }
  *
- *   const row1 = rows[0];
+ *   const firstRow = rows[0];
  *
- *   // row1 = [
+ *   // firstRow = [
  *   //   {
  *   //     name: 'SingerId',
  *   //     value: '1'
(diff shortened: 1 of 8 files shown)


src/database.js
```diff
@@ -674,6 +680,23 @@ Database.prototype.getTransaction = function(options, callback) {
...
  * //-
+ * // Alternatively, set `query.json` to `true`, and this step will be performed
+ * // automaticaly.
+ * //-
+ * database.run(query, function(err, rows) {
+ *   if (err) {
+ *     // Error handling omitted.
+ *   }
+ *
+ *   const row1 = rows[0];
```
alexander-fenster: looks really weird even for a sample code, can we call it `row0`, or `firstRow`, or just `row` ? :)

WaldoJeffers:
@stephenplusplus Good call 👍

src/database.js
```diff
@@ -862,8 +903,8 @@ Database.prototype.runStream = function(query, options) {
     };
   }
 
-  delete reqOpts.toJSON;
-  delete reqOpts.toJSONOptions;
+  delete reqOpts.json;
```
eric4545: Hi, why need `delete reqOpts.json;`? if the query didn't pass `params` or `types` it will return same reference of `query` object, which cause `query.json` also being deleted.

PR #132, opened 2018-02-26
feature: batch queries

callmehiphop:
- [x] Tests and linter pass

### DIFF
--- a/protos/google/spanner/v1/spanner.proto
+++ b/protos/google/spanner/v1/spanner.proto
@@ -1,4 +1,4 @@
-// Copyright 2017 Google Inc.
+// Copyright 2018 Google Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ option go_package = "google.golang.org/genproto/googleapis/spanner/v1;spanner";
 option java_multiple_files = true;
 option java_outer_classname = "SpannerProto";
 option java_package = "com.google.spanner.v1";
+option php_namespace = "Google\\Cloud\\Spanner\\V1";
 
 
 // Cloud Spanner API
@@ -159,6 +160,30 @@ service Spanner {
   rpc Rollback(RollbackRequest) returns (google.protobuf.Empty) {
     option (google.api.http) = { post: "/v1/{session=projects/*/instances/*/databases/*/sessions/*}:rollback" body: "*" };
   }
+
+  // Creates a set of partition tokens that can be used to execute a query
+  // operation in parallel.  Each of the returned partition tokens can be used
+  // by [ExecuteStreamingSql][google.spanner.v1.Spanner.ExecuteStreamingSql] to specify a subset
+  // of the query result to read.  The same session and read-only transaction
+  // must be used by the PartitionQueryRequest used to create the
+  // partition tokens and the ExecuteSqlRequests that use the partition tokens.
+  // Partition tokens become invalid when the session used to create them
+  // is deleted or begins a new transaction.
+  rpc PartitionQuery(PartitionQueryRequest) returns (PartitionResponse) {
+    option (google.api.http) = { post: "/v1/{session=projects/*/instances/*/databases/*/sessions/*}:partitionQuery" body: "*" };
+  }
+
+  // Creates a set of partition tokens that can be used to execute a read
+  // operation in parallel.  Each of the returned partition tokens can be used
+  // by [StreamingRead][google.spanner.v1.Spanner.StreamingRead] to specify a subset of the read
+  // result to read.  The same session and read-only transaction must be used by
+  // the PartitionReadRequest used to create the partition tokens and the
+  // ReadRequests that use the partition tokens.
+  // Partition tokens become invalid when the session used to create them
+  // is deleted or begins a new transaction.
+  rpc PartitionRead(PartitionReadRequest) returns (PartitionResponse) {
+    option (google.api.http) = { post: "/v1/{session=projects/*/instances/*/databases/*/sessions/*}:partitionRead" body: "*" };
+  }
 }
 
 // The request for [CreateSession][google.spanner.v1.Spanner.CreateSession].
@@ -172,7 +197,8 @@ message CreateSessionRequest {
 
 // A session in the Cloud Spanner API.
 message Session {
-  // The name of the session.
+  // The name of the session. This is always system-assigned; values provided
+  // when creating a session are ignored.
   string name = 1;
 
   // The labels for the session.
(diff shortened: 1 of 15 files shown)


protos/google/spanner/v1/spanner.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 Google Inc.
+// Copyright 2018 Google Inc.
```
alexander-fenster: should be Google LLC here and everywhere

samples/batch.js
```diff
@@ -0,0 +1,246 @@
+/**
+ * Copyright 2018, Google, Inc.
```
alexander-fenster: I think it should be Google LLC here and everywhere.

callmehiphop: Is this something we need to change across all files? If so, does something of that scale belong in a PR like this?

alexander-fenster: I meant only new `*.js` files introduced within this PR, of course. But probably it's OK to merge as is.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+};
+
+/**
+ * Generic create partition method. Handles common parameters used in both
+ * {@link BatchTransaction#createQueryPartitions} and {@link BatchTransaction#createReadPartitions}
+ *
+ * @private
+ *
+ * @param {object} config The request config.
+ * @param {function} Callback function.
```
stephenplusplus: @param {function} callback Callback function.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+ *       //     value: 'Eddie Wilson'
+ *       //   }
+ *       // ]
+ *     })
+ *     .on('end', function() {
+ *       // All results retrieved
+ *     });
+ * });
+ */
+BatchTransaction.prototype.executeStream = function(partition) {
```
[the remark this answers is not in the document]
callmehiphop: @stephenplusplus the naming here was actually at the request of the Spanner team.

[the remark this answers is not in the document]
stephenplusplus: @vkedia -- we have existing methods named `run` and `runStream`. To deviate and have `run` here and `execute` there is probably not good. Should we be consistent for now, then rename the existing `run` and `runStream` to `execute` and `executeStream` for 2.0?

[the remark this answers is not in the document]
callmehiphop: @stephenplusplus it should probably be noted that execute doesn't just execute a query, but instead a partition that could be for a query or a read request. I think its fine to call it execute here, because its not really the same as run/runStream.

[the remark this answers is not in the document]
stephenplusplus: Alright, whatever you guys think.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+ * @param {?Error} err Request error, if any.
+ * @param {ReadPartition[]} partitions List of read partitions.
+ * @param {object} apiResponse The full API response.
+ */
+/**
+ * Creates a set of read partitions that can be used to execute a read
+ * operation in parallel. Partitions become invalid when the transaction used
+ * to create them is closed.
+ *
+ * @param {ReadRequestOptions} options Configuration object, describing what to
```
stephenplusplus: We should document `gaxOptions`.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+ *
+ * @param {ReadPartition|QueryParition} partition The partition object.
+ * @param {TransactionRequestReadCallback|RunCallback} callback Callback
+ *     function.
+ * @returns {Promise<RunResponse>|Promise<TransactionRequestReadResponse>}
+ *
+ * @example <caption>include:samples/batch.js</caption>
+ * region_tag:execute_partition
+ */
+BatchTransaction.prototype.execute = function(partition, callback) {
```
stephenplusplus: Can we use gaxOptions here?

src/batch-transaction.js
stephenplusplus: Can we use gaxOptions here?

src/database.js
```diff
@@ -270,6 +309,56 @@ Database.prototype.close = function(callback) {
...
+ * Create a transaction that can be used for batch querying.
+ *
+ * @param {TransactionOptions} [options] [Transaction options](https://cloud.google.com/spanner/docs/timestamp-bounds).
+ * @param {CreateTransactionCallback} [callback] Callback function.
+ * @returns {Promise<CreateTransactionResponse>}
+ *
+ * @example <caption>include:samples/batch.js</caption>
+ * region_tag:create_batch_transaction
+ */
+Database.prototype.createBatchTransaction = function(options, callback) {
```
stephenplusplus: Is `gaxOptions` possible here?

src/database.js
```diff
@@ -524,6 +613,121 @@ Database.prototype.getSchema = function(callback) {
...
+ *     - **`labels.key`** where key is the name of a label
+ *
+ *     Some examples of using filters are:
+ *     - **`name:*`** The instance has a name.
+ *     - **`name:Howl`** The instance's name is howl.
+ *     - **`labels.env:*`** The instance has the label env.
+ *     - **`labels.env:dev`** The instance's label env has the value dev.
+ *     - **`name:howl labels.env:dev`** The instance's name is howl and it has
+ *       the label env with value dev.
+ * @property {number} [maxApiCalls] Maximum number of API calls to make.
```
stephenplusplus: We should document `gaxOptions`.

src/database.js
```diff
@@ -524,6 +613,121 @@ Database.prototype.getSchema = function(callback) {
...
+/**
+ * @typedef {array} GetSessionsResponse
+ * @property {Session[]} 0 Array of {@link Session} instances.
+ * @property {object} 1 The full API response.
+ */
+/**
+ * @callback GetSessionsCallback
+ * @param {?Error} err Request error, if any.
+ * @param {Session[]} instances Array of {@link Session} instances.
+ * @param {object} apiResponse The full API response.
```
stephenplusplus: I don't think we get an apiResponse from GAPIC.

src/database.js
```diff
@@ -524,6 +613,121 @@ Database.prototype.getSchema = function(callback) {
...
+ * @property {number} [maxApiCalls] Maximum number of API calls to make.
+ * @property {number} [maxResults] Maximum number of items to return.
+ * @property {number} [pageSize] Maximum number of results per page.
+ * @property {string} [pageToken] A previously-returned page token
+ *     representing part of the larger set of results to view.
+ */
+/**
+ * @typedef {array} GetSessionsResponse
+ * @property {Session[]} 0 Array of {@link Session} instances.
+ * @property {object} 1 The full API response.
```
stephenplusplus: I don't think we get an apiResponse from GAPIC.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * @typedef {array} CreateQueryPartitionsResponse
+ * @property {QueryPartition[]} 0 List of query partitions.
+ * @property {object} 1 The full API response.
+ */
+/**
+ * @callback CreateQueryPartitionsCallback
+ * @param {?Error} err Request error, if any.
+ * @param {QueryPartition[]} partitions List of query partitions.
+ * @param {object} apiResponse The full API response.
```
stephenplusplus: No apiResponse from GAPIC here.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * @see [`ExecuteSqlRequest`](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest)
+ * @typedef {object} QueryPartition
+ * @property {string} partitionToken The partition token.
+ */
+/**
+ * @typedef {array} CreateQueryPartitionsResponse
+ * @property {QueryPartition[]} 0 List of query partitions.
+ * @property {object} 1 The full API response.
```
stephenplusplus: No apiResponse from GAPIC here.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+ */
+/**
+ * Creates a set of query partitions that can be used to execute a query
+ * operation in parallel. Partitions become invalid when the transaction used
+ * to create them is closed.
+ *
+ * @param {string|object} query A SQL query or
+ *     [`ExecuteSqlRequest`](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest)
+ *     object.
+ * @param {object} [query.params] A map of parameter name to values.
```
stephenplusplus: We should document `gaxOptions` here.

src/batch-transaction.js
stephenplusplus: Just realized this was a private function. Like, who cares, @stephenplusplus?

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * @typedef {array} CreateReadPartitionsResponse
+ * @property {ReadPartition[]} 0 List of read partitions.
+ * @property {object} 1 The full API response.
+ */
+/**
+ * @callback CreateReadPartitionsCallback
+ * @param {?Error} err Request error, if any.
+ * @param {ReadPartition[]} partitions List of read partitions.
+ * @param {object} apiResponse The full API response.
```
stephenplusplus: No API response from GAPIC.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * @typedef {object} ReadPartition
+ * @mixes ReadRequestOptions
+ * @property {string} partitionToken The partition token.
+ */
+/**
+ * @typedef {array} CreateReadPartitionsResponse
+ * @property {ReadPartition[]} 0 List of read partitions.
+ * @property {object} 1 The full API response.
```
stephenplusplus: No API response from GAPIC.

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    callback(null, partitions, resp);
+  });
+};
+
+/**
+ * @typedef {object} ReadPartition
+ * @mixes ReadRequestOptions
+ * @property {string} partitionToken The partition token.
```
stephenplusplus: This would normally be a place to do auto-unwrapping for the user, so they don't have to dig through an object with only one property (obviously, they want the token). As long as we add an `apiResponse` argument with the full, original object, it should be more convenient to the user:

```diff
var options = {
  table: 'Singers',
  keys: ['1'],
  columns: ['SingerId'],
}

- createReadPartitions(options, function(err, readPartitions) {
-   const readPartitionTokens = readPartitions.map(partition => partition.partitionToken)
- })
+ createReadPartitions(options, function(err, readPartitionTokens) {})
```

src/batch-transaction.js
```diff
@@ -0,0 +1,343 @@
...
+ * @param {ReadPartition|QueryPartition} partition The partition object.
+ * @returns {ReadableStream} A readable stream that emits rows.
+ *
+ * @example
+ * const Spanner = require('@google-cloud/spanner');
+ * const spanner = new Spanner();
+ *
+ * const instance = spanner.instance('my-instance');
+ * const database = instance.database('my-database');
+ * const transaction = database.batchTransaction(identifier);
```
stephenplusplus: Is `identifier` supposed to be defined?

src/batch-transaction.js
callmehiphop: Yes!

src/batch-transaction.js
callmehiphop: Yes

src/batch-transaction.js
callmehiphop: Yes

src/database.js
callmehiphop: I'm going to go ahead and say no since we call multiple rpcs here.

src/batch-transaction.js
callmehiphop: This isn't a paginated list, so there is a response.

src/batch-transaction.js
callmehiphop: See https://github.com/googleapis/nodejs-spanner/pull/132#discussion_r171338900

src/batch-transaction.js
callmehiphop: See https://github.com/googleapis/nodejs-spanner/pull/132#discussion_r171338900

src/batch-transaction.js
callmehiphop: See https://github.com/googleapis/nodejs-spanner/pull/132#discussion_r171338900

src/batch-transaction.js
callmehiphop: I can't speak as to why its done this way, but in order to execute a partition you need the original query/read request as well. We actually deliberately extend the original query to include the partition token.

src/batch-transaction.js
stephenplusplus: Oops, then the way it is currently, the user callback receives what GAPIC returns which is:

https://github.com/callmehiphop/nodejs-spanner/blob/df0116e5f450cc8bf4759e49494b08b92dc83634/src/v1/spanner_client.js#L1202

> The function which will be called with the result of the API call.
> The second parameter to the callback is an object representing [PartitionResponse]{@link google.spanner.v1.PartitionResponse}.

src/batch-transaction.js
stephenplusplus: https://github.com/callmehiphop/nodejs-spanner/blob/df0116e5f450cc8bf4759e49494b08b92dc83634/src/v1/spanner_client.js#L1288

```
  * @param {function(?Error, ?Object)} [callback]
   *   The function which will be called with the result of the API call.
   *
   *   The second parameter to the callback is an object representing [PartitionResponse]{@link google.spanner.v1.PartitionResponse}.
```

src/batch-transaction.js
callmehiphop: No, the idea behind this API is that you create a single transaction to be shared amongst several workers. A user can create an identifier via `transaction.identifier()`. How they pass the identifier to other workers is an implementation detail, so I chose to omit it.

stephenplusplus: I know we don't isolate examples and test them anymore, so runnable JS isn't exactly important. Still, is it possible to save users from going "wait, where is `identifier` defined?" and just put something fake in? If it's too complicated, I suppose even:

```js
const identifier = {};
const transaction = database.batchTransaction(identifier);
```

src/database.js
callmehiphop: I just checked and we do, it contains a key for both the `sessions` and `nextPageToken`.

src/database.js
callmehiphop: See https://github.com/googleapis/nodejs-spanner/pull/132#discussion_r171342654

src/batch-transaction.js
stephenplusplus: > We actually deliberately extend the original query to include the partition token.

Where does that happen?

stephenplusplus: Wait, sorry, pause this convo. I just realized this is the options argument and not the response from the API.

src/batch-transaction.js
callmehiphop: So then are you asking me to rename api response to partition response?

src/batch-transaction.js
callmehiphop: I think stubbing the identifier is more confusing, so I just rewrote a section of the example to avoid using it entirely.

src/batch-transaction.js
stephenplusplus: I might just not get JSDocs, but it looks like this says that there are two responses that come back, an array of "ReadPartition"s, and an API response. But in the code, we only return what GAPIC returns, which is just "PartitionResponse".

callmehiphop: Oh, the method we call here isn't a gapic method, this is a private one we control. https://github.com/googleapis/nodejs-spanner/pull/132/files/df0116e5f450cc8bf4759e49494b08b92dc83634#diff-2577c1ac4653d72633571d9a02d071d8R180

src/batch-transaction.js
stephenplusplus: */mark as resolved*

src/batch-transaction.js
stephenplusplus: Impressive patience you have with me.

*/mark as resolved*

src/batch-transaction.js
stephenplusplus: */mark as resolved*

src/batch-transaction.js
stephenplusplus: */mark as resolved*

src/batch-transaction.js
stephenplusplus: */mark as resolved*

PR #133, opened 2018-02-28
Remove autoCreate for objects that return operations.

### LINKED ISSUE #127: `options.autoCreate` not working
Tried with `database.get({ autoCreate: true })` and `instance.get({ autoCreate: true })`. They both return the standard 404 error for the resource not existing.

stephenplusplus:
Fixes #127 

This adds get-or-create behavior to database.get() and instance.get().


stephenplusplus:
Please discuss over in #127 if you think this is the wrong solution.

stephenplusplus:
@callmehiphop ready for review.

PR #141, opened 2018-03-05
fix: Clone query passed to encodeQuery

stephenplusplus:
Fixes https://github.com/googleapis/nodejs-spanner/pull/122/files#r172219962


stephenplusplus:
cc @eric4545

PR #148, opened 2018-03-08
bugfix: promisify database#delete

callmehiphop:
I think we used to exclude it because we returned the gapic request, but never included it after changing that.


walshie4:
This was mentioned in the `v1.3.0` release with the following note:

> Promisify database.close() (#148)

Is that a typo?  Was it supposed to refer to `delete` instead of `close` or does this indirectly impact `close`?

callmehiphop:
@walshie4 it is a typo! Whoops, sorry about that.

PR #149, opened 2018-03-08
remove circleci fix

alexander-fenster:
Since @callmehiphop found the __real__ issue with samples, we don't need this anymore (it did not really help anyway).

- [x] Tests and linter pass


samples/system-test/util/circleci_wait.js
```diff
@@ -1,85 +0,0 @@
-/**
- * @fileoverview Samples tests work with the same project in Google Cloud and
- * can fail if more than one test job is running at the same time.
- * This script will help CircleCI job wait until all the similar tasks are
- * completed.
- */
-
-const axios = require('axios');
```
callmehiphop: Does this mean we can uninstall axios from our deps too?

PR #150, opened 2018-03-09
test: make sure tests can run in parallel

alexander-fenster:
Since the problem of failing spanner tests was resolved in #148, the fix from #147 is not needed so let's put it back: we still want to delete our test data. Also, since we now allow multiple samples tests to run simultaneously (#149), we need to make sure that they are working with different instances, otherwise, when two CI tasks run simultaneously, the `before` hook of the second run will delete the instance that the first run is still using.

[Also, found out that we cannot just make the test run in parallel, as they depend on their order. Ideally, they should be rewritten :)]

- [x] Tests and linter pass
- [x] And samples tests pass as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "power-assert": "^1.4.4",
     "prettier": "^1.7.4",
     "proxyquire": "^2.0.0",
+    "request": "^2.83.0",
     "stats-lite": "^2.1.0",
     "time-span": "^2.0.0",
     "uuid": "^3.2.1",
--- a/package.json
+++ b/package.json
@@ -107,7 +107,6 @@
     "power-assert": "^1.4.4",
     "prettier": "^1.7.4",
     "proxyquire": "^2.0.0",
-    "request": "^2.83.0",
     "stats-lite": "^2.1.0",
     "time-span": "^2.0.0",
     "uuid": "^3.2.1",
--- a/samples/package.json
+++ b/samples/package.json
@@ -21,6 +21,7 @@
     "@google-cloud/nodejs-repo-tools": "2.1.0",
     "ava": "0.22.0",
     "proxyquire": "1.8.0",
+    "request": "^2.83.0",
     "sinon": "3.2.1"
   }
 }
--- a/samples/system-test/spanner.test.js
+++ b/samples/system-test/spanner.test.js
@@ -16,7 +16,7 @@
 'use strict';
 
 const path = require(`path`);
-const request = require('request');
+const request = require(`request`);
 const Spanner = require(`@google-cloud/spanner`);
 const test = require(`ava`);
 const tools = require(`@google-cloud/nodejs-repo-tools`);
@@ -87,7 +87,7 @@ test.before(async () => {
 
         return instanceCreated < yesterday;
       })
-      .forEach(async operation => await instance.delete());
+      .forEach(async () => await instance.delete());
   });
 });
 
--- a/samples/system-test/spanner.test.js
+++ b/samples/system-test/spanner.test.js
@@ -38,7 +38,7 @@ const spanner = new Spanner({
   projectId: PROJECT_ID,
 });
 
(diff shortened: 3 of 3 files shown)


stephenplusplus:
> Also, since we now allow multiple samples tests to run simultaneously (#149), 

When does Circle run the sample tests simultaneously?

callmehiphop:
> When does Circle run the sample tests simultaneously?

I think when multiple builds fire off? If you checked the console output when #149 was still in effect, it would poll the server and wait for there to be no more pending jobs before running.

stephenplusplus:
I didn't know we had multiple builds going at once, I thought we had only one. Where do we configure that?

alexander-fenster:
@stephenplusplus We have 6 CircleCI tasks running in parallel for googleapis org, and since samples test take more than 5 minutes, it's possible that when 2 PRs are merged one after one, two samples test tasks start running at the same time and one of them would fail because they both use the same instance (but different databases though).  The output @callmehiphop is talking about was from my script (introduced in #113, removed in #149) - I'm pretty confident now that running two samples test in parallel is safe if they use different instances, so there is no need to serialize tasks anymore.

stephenplusplus:
Ah, gotcha. So we aren't doing any before/after catch all clean up, right? I guess if we weren't already, then this isn't the place to add it. Although, since we're on the subject, is that what we want? In the event the after all hook doesn't run (I guess a rare scenario), we would have a leftover instance, never to be picked up for deletion again.

alexander-fenster:
@stephenplusplus That's a good point. I would say, we have three options here:
- unique instance created in `before` and deleted in `after` hook. If a test breaks in the middle (especially possible when someone runs it manually and presses Ctrl+C), the leftover instance will never be deleted.
- same instance, different databases. Need to put back the protection I removed in #149. In case of a test stopping in the middle, the instance will be deleted by the following test run.
- same instance, different databases, no protection (#149) - current state (and the same state as before #113). Simultaneously running samples test will fail.

Note: for system tests we do use different instance names. We use the following function to generate truly unique name:
```js
function generateName(resourceType) {
  return (  
    PREFIX +  
    resourceType +
    '-' +     
    uuid        
      .v1()   
      .split('-')
      .shift()
  );    
}
```

and yet I don't see many leftover instances left. It's probably OK to delete them manually time to time. I would vote for the first option (implemented in this PR) as the most straightforward one. What's your opinions @stephenplusplus and @callmehiphop ?

stephenplusplus:
Alright, here is an option #4!

- List all instances
- Make sure they have our `gcloud-tests` label
- If they are more than 24 hours old, delete 'em

```js
function apiRequest(reqOpts) {
  return new Promise((resolve, reject) => {
    spanner.auth.authorizeRequest(reqOpts, (err, reqOpts) => {
      if (err) {
        reject(err);
        return;
      }

      request(reqOpts, (err, response) => {
        if (err) {
          reject(err);
          return;
        }

        try {
          resolve(JSON.parse(response.body));
        } catch (e) {
          reject(e);
          return;
        }
      });
    });
  });
}

function getOperations(instanceName) {
  return apiRequest({
    uri: `https://spanner.googleapis.com/v1/${instanceName}/operations`
  });
}

const [instances] = await spanner.getInstances({
  filter: 'labels.gcloud-tests:true',
});

instances.forEach(async (instance) => {
  const {operations} = await getOperations(instance.metadata.name);

  operations
    .filter(operation => {
      return operation.metadata['@type'].includes('CreateInstance')
    })
    .filter(operation => {
      const yesterday = new Date();
      yesterday.setHours(-24);

      const instanceCreated = new Date(operation.metadata.startTime);

      return instanceCreated > yesterday;
    })
    .forEach(async operation => await instance.delete());
});
```

alexander-fenster:
Confirming that both authors are OK with merging (for CLA bot)

stephenplusplus:
I'm okay!

PR #159, opened 2018-03-20
Update google-auto-auth

stephenplusplus:


alexander-fenster:
I'm pretty sure you need to regenerate `samples/package-lock.json` as well otherwise something will break (as it happened with @callmehiphop before). I just have the following in my bash history:
```sh
rm -rf node_modules samples/node_modules package-lock.json samples/package_lock.json
npm install && npm link && cd samples && npm link ../ && npm install && cd .. && npm run lint
git add package-lock.json samples/package-lock.json
```

PR #176, opened 2018-04-22
docs: fix table

JoshFerge:
the docs in the `Table` class seem to have a consistent mistake of referring to `instance.database` instead of `database.table` to create a table object. This PR corrects. If there's more to this, an issue can be opened instead.
- [x] Tests and linter pass


stephenplusplus:
Woops. Thanks!

PR #178, opened 2018-04-24
feat: struct query parameter

callmehiphop:
## [Do not merge until feature is public](https://github.com/googleapis/nodejs-spanner/pull/178#issuecomment-393341776)

- [x] Tests and linter pass

### DIFF
--- a/src/codec.js
+++ b/src/codec.js
@@ -69,6 +69,39 @@ Int.prototype.valueOf = function() {
 
 codec.Int = Int;
 
+var TYPE = Symbol();
+
+var Struct = {
+  TYPE: 'struct',
+  create: function() {
+    var struct = [];
+
+    struct[TYPE] = Struct.TYPE;
+
+    Object.defineProperty(struct, 'toJSON', {
+      enumerable: false,
+      value: generateToJSONFromRow(struct)
+    });
+
+    return struct;
+  },
+  createFromJSON: function(json) {
+    var struct = Struct.create();
+
+    Object.keys(json || {}).forEach(name => {
+      let value = json[name];
+      struct.push({name, value});
+    });
+
+    return struct;
+  },
+  isStruct: function(thing) {
+    return thing && thing[TYPE] === Struct.TYPE;
+  },
+};
+
+module.exports.Struct = Struct;
+
 /**
  * Wherever a row object is returned, it is assigned a "toJSON" function. This
  * function will create that function in a consistent format.
@@ -148,27 +181,18 @@ function decode(value, field) {
         break;
       }
       case 'STRUCT': {
-        var formattedRow = [];
-        var fields = type.structType.fields;
+        let struct = Struct.create();
+        let fields = type.structType.fields;
 
-        fields.forEach(function(field, index) {
-          var value = decoded[field.name] || decoded[index];
+        fields.forEach((field, index) => {
+          let name = field.name;
+          let value = decoded[name] || decoded[index];
 
-          var column = {
-            name: field.name,
-            value: decodeValue_(value, field.type),
-          };
-
-          formattedRow.push(column);
+          value = decodeValue_(value, field.type);
+          struct.push({name, value});
         });
 
-        Object.defineProperty(formattedRow, 'toJSON', {
-          enumerable: false,
-          value: codec.generateToJSONFromRow(formattedRow),
-        });
-
-        decoded = formattedRow;
-
+        decoded = struct;
         break;
       }
     }
@@ -204,6 +228,8 @@ function encode(value) {
       value = value.value;
     } else if (Buffer.isBuffer(value)) {
       value = value.toString('base64');
+    } else if (Struct.isStruct(value)) {
+      value = value.map(field => preEncode(field.value));
     } else if (is.array(value)) {
       value = value.map(preEncode);
     } else if (is.object(value) && is.fn(value.hasOwnProperty)) {
@@ -270,10 +296,22 @@ function getType(field) {
     return 'date';
   }
 
-  if (is.array(field)) {
-    var child;
+  if (Struct.isStruct(field)) {
+    let fields = field.map(({name, value}) => {
+      let type = getType(value);
+      return {name, type};
+    });
 
-    for (var i = 0; i < field.length; i++) {
+    return {
+      type: 'struct',
+      fields,
+    };
+  }
+
+  if (is.array(field)) {
+    let child;
+
+    for (let i = 0; i < field.length; i++) {
       child = field[i];
 
       if (!is.nil(child)) {
@@ -308,6 +346,7 @@ var TYPES = [
   'string',
   'bytes',
   'array',
+  'struct',
 ];
 
 codec.TYPES = TYPES;
@@ -348,34 +387,8 @@ function encodeQuery(query) {
   if (query.types) {
     let formattedTypes = {};
 
-    for (let prop in query.types) {
-      let type = query.types[prop];
-      let childType;
-      let child;
-
-      // if a type is an ARRAY, then we'll accept an object specifying
-      // the type and the child type
-      if (is.object(type)) {
-        childType = type.child;
-        child = codec.TYPES.indexOf(childType);
-        type = type.type;
-      }
-
-      let code = codec.TYPES.indexOf(type);
-
-      if (code === -1) {
-        code = 0; // unspecified
-      }
-
-      formattedTypes[prop] = {code};
-
-      if (child === -1) {
-        child = 0; // unspecified
-      }
-
-      if (is.number(child)) {
-        formattedTypes[prop].arrayElementType = {code: child};
-      }
+    for (let field in query.types) {
+      formattedTypes[field] = createTypeObject(query.types[field]);
     }
 
     delete query.types;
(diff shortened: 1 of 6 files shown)


src/codec.js
```diff
@@ -69,6 +69,39 @@ Int.prototype.valueOf = function() {
 
 codec.Int = Int;
 
+var TYPE = Symbol();
+
+var Struct = {
```
callmehiphop: @stephenplusplus I know this isn't what we initially discussed, but I ran into some weirdness when subclassing Arrays, so I decided not to. LMK if you'd prefer I organize/structure this differently, I know singletons aren't really something we do.

stephenplusplus: I can go either way between this being the way it is now or if we went with a Class system. You can return the array from a class constructor, so it should be the same. `createFromJSON` and `isStruct` could be static methods. The reason I prefer that is for consistency with our typical usage, and the reason I don't is because you already didn't do it that way. So again, I can go either way :)

callmehiphop:
@stephenplusplus I think the unit tests are ready for you to look at!

src/codec.js
```diff
@@ -69,6 +69,42 @@ Int.prototype.valueOf = function() {
...
+  Object.keys(json || {}).forEach(function(name) {
+    var value = json[name];
+    struct.push({name, value});
+  });
+
+  return struct;
+};
+
+Struct.isStruct = function(thing) {
+  return !!(thing && thing[TYPE] === Struct.TYPE);
```
stephenplusplus: I think just `thing && thing[TYPE] === Struct.TYPE` will always return a bool, so the `!!()` wrapper isn't necessary?

test/codec.js
```diff
@@ -990,4 +996,117 @@ describe('codec', function() {
...
+    });
+
+    it('should set the arrayElementType', function() {
+      var type = codec.createTypeObject({
+        type: 'array',
+        child: 'bool',
+      });
+
+      assert.deepEqual(type, {
+        code: 8,
```
stephenplusplus: Can we do `TYPES.indexOf('array')` here? (Same for other places in the tests that we use the magic numbers)

src/codec.js
```diff
@@ -69,6 +69,42 @@ Int.prototype.valueOf = function() {
 
 codec.Int = Int;
 
+var TYPE = Symbol();
+
+codec.TYPE = TYPE;
+
+function Struct() {
```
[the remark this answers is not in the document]
callmehiphop: Sure thing, I didn't originally only because we don't document any of the internal custom types (Int, etc.)

src/codec.js
callmehiphop: That's not true actually, if `thing` is `null` then it would return `null` since the first expression is falsey.

src/codec.js
```diff
@@ -270,10 +299,22 @@ function getType(field) {
     return 'date';
   }
 
+  if (Struct.isStruct(field)) {
+    let fields = field.map(({name, value}) => {
```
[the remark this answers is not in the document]
callmehiphop: It does not! I thought we stopped caring about 4, but I'll make this 4 friendly!

system-test/spanner.js
```diff
@@ -2198,6 +2201,316 @@ describe('Spanner', function() {
...
+              assert.ifError(err);
+
+              var row = rows[0].toJSON();
+              assert.strictEqual(row.f1, null);
+
+              done();
+            });
+          });
+
+          it.skip('should allow equality checks', function(done) {
```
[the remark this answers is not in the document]
callmehiphop: I couldn't get this to pass originally, but it turns out that the params are flipped in the query, so the upstream API returns a super unhelpful error.

src/codec.js
```diff
@@ -69,6 +69,42 @@ Int.prototype.valueOf = function() {
 
 codec.Int = Int;
 
+var TYPE = Symbol();
```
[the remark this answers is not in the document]
callmehiphop: Done!

system-test/spanner.js
```diff
@@ -2198,6 +2201,316 @@ describe('Spanner', function() {
             });
           });
         });
+
+        describe.only('structs', function() {
```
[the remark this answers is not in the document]
callmehiphop: Yeah, we need to revert the config change when its available as well.

test/codec.js
callmehiphop: Done!

stephenplusplus:
I can't check the system tests, but... pre-approved!

system-test/spanner.js
```diff
@@ -2198,6 +2201,305 @@ describe('Spanner', function() {
...
+            });
+          });
+
+          it('should bind null nested structs', function(done) {
+            var query = {
+              sql: 'SELECT @structParam.structf.nestedf',
+              params: {
+                structParam: null,
+              },
+              types: {
```
vkedia: what happens if types is not specified and the value for structParams is null?

system-test/spanner.js
```diff
@@ -2198,6 +2201,305 @@ describe('Spanner', function() {
...
+            };
+
+            database.run(query, function(err, rows) {
+              assert.ifError(err);
+              assert.strictEqual(rows.length, 0);
+              done();
+            });
+          });
+        });
       });
```
vkedia: what about structs with duplicate field names and missing field names?

src/codec.js
```diff
@@ -69,6 +69,80 @@ Int.prototype.valueOf = function() {
 
 codec.Int = Int;
 
+/**
+ * We use this symbol as a means to identify if an array is actually a struct.
+ * We need to do this because changing Structs from an object to an array would
+ * be a major breaking change.
```
vkedia: I don't understand this comment. can you please elaborate?

system-test/spanner.js
callmehiphop: An exception would be thrown.

system-test/spanner.js
callmehiphop: Struct objects in the Node client are always Arrays containing objects with a name field, if the struct with duplicate/empty names comes back from the API we handle it for them. We can also add a `Struct.fromArray` method for users who want to build objects like that themselves.

src/codec.js
callmehiphop: When the types dictionary is omitted, we attempt to determine the type for the user. Because structs can have empty/duplicate field names, we've always used an Array to represent a struct. Now that we accept it as a param we need a way to distinguish between an actual array and a struct (assuming the types object is null).

src/index.js
```diff
@@ -270,6 +270,23 @@ Spanner.int = function(value) {
...
+ * @returns {object}
+ *
+ * @example
+ * const Spanner = require('@google-cloud/spanner');
+ * const struct = Spanner.struct({
+ *   user: 'bob',
+ *   age: 32
+ * });
+ */
+Spanner.struct = function(value) {
```
vkedia: why can't users directly pass in a json object as the value of struct parameter? Why do they have to wrap it in `Spanner.struct`

src/codec.js
vkedia: So to bind a struct parameter, users can either pass in a `Spanner.struct(object)` or an Array?

vkedia: Any updates?

src/index.js
vkedia: Any updates?

src/codec.js
danoscarmike: ping @callmehiphop

src/index.js
callmehiphop: I suppose we could allow them to pass objects directly, my thinking was that by default we don't give the users structs in the form of plain old objects by default, so accepting them by default might be confusing.

@stephenplusplus had brought up the same point and I thought it might be better to allow them to pass in an object by default once we defaulted to returning json.

src/codec.js
callmehiphop: @vkedia thats correct

system-test/spanner.js
```diff
@@ -2198,6 +2201,305 @@ describe('Spanner', function() {
...
+              var row = rows[0].toJSON();
+              assert.strictEqual(row.userf, 'bob');
+
+              done();
+            });
+          });
+
+          it('should bind null structs', function(done) {
+            var query = {
+              sql: 'SELECT @structParam.userf',
```
adirastogi: Perhaps change this test to `SELECT @structParam IS NULL` ?

system-test/spanner.js
adirastogi: It would be nice to add integration tests for these two scenarios.

callmehiphop: @adirastogi I went ahead and added the requested tests in [this commit](https://github.com/googleapis/nodejs-spanner/pull/178/commits/a988aa00cc8765c6d6e69f3da7804d848d3d1a65). PTAL!

system-test/spanner.js
callmehiphop: Done!

stephenplusplus:
Holding off on merging because of the `do not merge` label to be safe. @callmehiphop feel free to merge if that's stale.

adirastogi:
This should only be merged into the public repo after the server-side feature has become available (after 6/11). otherwise the integration tests will fail.

PR #185, opened 2018-05-01
test(sessionPool): check for out of bounds write value

### LINKED ISSUE #184: Spanner session count growing well beyond limit
#### Environment details

  - OS: CoreOS
  - Node.js version: v8.9.6
  - npm version: v5.6.0
  - `@google-cloud/spanner` version: v1.4.1

#### Steps to reproduce
Will create some if needed

I am running a replicated set of APIs using this library with the following session pool configuration:
```
max: 700
min: 300
concurrency: 200
idlesAfter: 4
maxIdle: 1
acquireTimeout: 4000
writes: 50
```

You may have noticed the `writes: 50` value down there as being a bit odd.  I had forgotten this value was a percentage and accidentally rolled out an update with this config.  From my understanding of this configuration if I have 10 API instances I should see a max session count for the Spanner instance (assuming these APIs are the only ones connecting) of 7000 sessions. 

However when I deployed this change I went from 2,401 sessions all the way up to 50,092 sessions which put me over the session limit for my 5 node Spanner instance.  I understand that doing 50% write sessions is overly-aggressive however I keep seeing this go over what I view as the theoretical session limit which is very concerning considering once that happens all requests start to fail and it is non-trivial to free up these sessions by bouncing the service or something similar.  Please let me know if I'm missing something in understanding the pooling logic.

callmehiphop:
Fixes #184

- [x] Tests and linter pass

### DIFF
--- a/src/session-pool.js
+++ b/src/session-pool.js
@@ -63,8 +63,9 @@ var READWRITE = 'readwrite';
  *     any given time.
  * @property {number} [keepAlive=50] How often to ping idle sessions, in
  *     minutes. Must be less than 1 hour.
- * @property {number} [writes=0] Percentage of session to be pre-allocated as
- *     write sessions.
+ * @property {number} [writes=0.0] Percentage of sessions to be pre-allocated as
+ *     write sessions represented as a float.
+ */
 
 /**
  * Class used to manage connections to Spanner.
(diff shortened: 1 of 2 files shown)


src/session-pool.js
```diff
@@ -93,6 +94,12 @@ function SessionPool(database, options) {
    */
   this.options = extend({}, DEFAULTS, options);
 
+  if (this.options.writes > 1) {
+    throw new TypeError(
+      'write percentage should be represented as a float between 0.0 and 1.0'
```
stephenplusplus: Just grammar stuff to match our usual throwing language.

```diff
- write percentage should be represented as a float between 0.0 and 1.0
+ Write percentage should be represented as a float between 0.0 and 1.0.
```

callmehiphop: Done!

PR #194, opened 2018-05-11
AB - New session pool code

### LINKED ISSUE #192: Session Pool Leaks & Investigation
#### Environment details

  - OS: CoreOS
  - Node.js version: v8.9.6
  - npm version: v6.0.0
  - `@google-cloud/spanner` version: v1.4.1 (kinda - see below)

#### The issue

As I've mentioned in my other issues (#183, #177) we're (@honeyscience) attempting to apply a fairly large amount of load to our new Spanner backed APIs but are running into a slew of session pooling issues.  The main issues are:

* Time spent waiting for a session is too high
* Session pool grows seemingly without bound (#177)
* When the session max is hit you either need to spin up even more nodes ($$$) or wait ~an hour for sessions to be deleted by the server

As per our previous issue (#134) & PR (#135) into the session pooling code the module is setup in such a way which makes it quite prone to race conditions.  On top of that the session pool management responsibilities are not 100% contained within the session pool module itself which has made understanding it a larger challenge.  Without knowing exactly what the authors were aiming for, and not sure we’d be able to confidently ensure no leaks with the current architecture, we decided to replace it rather than try to fix it.

Our first attempt at replacing it was simply a rudimentary create-upon-request and destroy after 60 seconds setup.  You can see this here (https://github.com/honeyscience/nodejs-spanner/commit/c76df11457e3b72fa0c1f32088af780f9e64e730#diff-0c4114554752f1c79d35d17a91563062R292).

Although this setup suffered from some write failures and increased latency overall it worked amazingly.  We were able to run the API overnight with 120 replicas of the API _in total_ never consuming more than 100 Spanner sessions.

image

This is of course non-optimal as you can see from the following chart where the purple line is session destruction and the blue line is session creation:
image

The impact of this approach is increased latency and some occasional failures.  However as a proof of concept we were more than pleased with the results.

We then decided we’d try to implement our own pool implementation leveraging this library for the main pool management functionality: https://github.com/coopernurse/node-pool

The code for this is available here https://github.com/honeyscience/nodejs-spanner/blob/869871e45a034af2ba9158466912dc4f17e8e32b/src/session-pool.js

It still has some issues (particularly with write sessions) but so far it has been doing fairly well.  

The main issue we’ve seen with our current setup is everything seems to be only using write sessions which are sometimes slow to create or be released back into the pool.  We suspect there is still a location not releasing back to the pool because with the new setup the pool properly enforces the max session count but all are marked borrowed so everything is queued and thus times out.

Session count chart:
image

Spanner API requests for the same period:
image

When the limit for sessions was much higher we saw failed writes causes spikes in session create requests (upwards of a few thousand per second - see screenshot) but does not solve write failures, they still timeout. Which seems like a similar issue if not the same one. 

Looking around we found this PR which addressed an issue with write sessions https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2561. Could be relevant to the write session speed issue.

We're currently looking into if this change https://github.com/googleapis/nodejs-spanner/commit/f5897d659c6300a31247d4d1e4a162a7f572d05f will resolve the issue with sessions getting stuck in the borrowed state.

If there is any other information you can give us on how sessions work which could be relevant to our efforts we're all ears.

We’re still digging at this from our side and will report back with any more findings.  We're more than happy to work with anyone on your side to get a speedy and solid resolution to this.

anupbaldawa:
Fixes #192

- [x] Tests and linter pass
     --- It is a major change. I have updated the documentation inside the code.


anup-honey:
All the tests are successful but the result of the test did not print. I was seeing the same problem when debugging locally.

walshie4:
Test issue is likely this https://boneskull.com/mocha-v4-nears-release/#mochawontforceexit

There's some leftover async code running not being cleaned up from one of the tests.

walshie4:
Ran the tests using this module [0] to see what's going on and it's output suggests it's the session pinging code.  Commenting that `setTimeout` call out resolves the problem.  I'm going to look into a fix for this.

[0] https://github.com/myndzi/wtfnode

walshie4:
Still in the midst of getting the tests to properly clean up after themselves (should be done by around noon).  However I have opened this PR against the @honeyscience master branch (which will add it to this PR) to resolve leaking the `pingSessions_` timeout. 

https://github.com/honeyscience/nodejs-spanner/pull/3

walshie4:
Tests are now passing without having to pass `--exit` 🎉 

https://github.com/honeyscience/nodejs-spanner/pull/4

We'll be reviewing this code this morning and should have it merged into this PR today.

walshie4:
👋 Hey, just checked the Corporate CLA and I'm covered on it.  Company name is `Honey Science Corporation`, was signed on March 5th, 2018.

Just let me know if I need to do some special linking or something to get this to go through.

test/session-pool.js
```diff
@@ -16,48 +16,40 @@
...
-var events = require('events');
-var extend = require('extend');
-var PQueue = require('p-queue');
-var proxyquire = require('proxyquire');
-var stackTrace = require('stack-trace');
-var through = require('through2');
-var timeSpan = require('time-span');
+const assert = require('assert');
+const common = require('@google-cloud/common');
+const events = require('events');
```
DuaneGarber: this var is no longer used in the file and it can be removed

walshie4: Huh thought my lint fix run caught that, I'll run that again.

src/session-pool.js
```diff
@@ -109,221 +97,71 @@ function SessionPool(database, options) {
...
-    sessions.map(function(session) {
-      return self.destroySession_(session);
-    })
+  this.readPool = Pool.createPool(
+    {
+      create: this.createReadSession.bind(this),
+      destroy: this.destroySession.bind(this),
+    },
+    {
+      max: this.options.maxReads,
```
DuaneGarber: Since you use this options key twice, you should destructure it above and just use the variables, would be easier to check for valid properties and improve performances trivially

alexander-fenster:
@walshie4 The CI is taking extremely long time for your latest commit - looks like it just hangs somewhere. Are you able to see the CircleCI logs of your CI tasks?

I will cancel the running tests as they already take more than 10 minutes each and it started creating a backup in our CI queue :)

alexander-fenster:
upd: they just failed (timeout probably?), I did not cancel anything.

walshie4:
@alexander-fenster Hmm that was the issue my changes were intending to fix.  I'll take another look.

src/session-pool.js
```diff
@@ -550,399 +288,178 @@ SessionPool.prototype.createTransaction_ = function(session, options) {
...
-    if (self.needsFill_()) {
-      self.fill();
-    }
-  });
+SessionPool.prototype.getStats = function() {
+  const readPool = this.readPool;
+  const writePool = this.writePool;
+  const stats = {
+    readPool: {
+      spareResourceCapacity: readPool.spareResourceCapacity,
```
DuaneGarber: How big is this object, why wouldnt you just do an `object.assign` here?

walshie4: That object is tiny, just a few numbers.

walshie4:
Oh derp, I see what happened, one of my branches never got merged in.

walshie4:
Coverage failures are (from what I can tell) from this line:

https://github.com/googleapis/nodejs-spanner/pull/194/files#diff-0c4114554752f1c79d35d17a91563062R445

@anup-honey is working on another PR which will change how pinging sessions works and as a side-effect remove that line of code, so we should probably wait to scrutinize this line too heavily.

stephenplusplus:
We might have a major bump coming soon, if we break support for Node v4 now that it's EOL. @crwilcox @alexander-fenster -- are we doing that?

alexander-fenster:
@ofrobots - here we are going to have a proposed breaking change in Spanner library. Do you feel we can drop Node 4 support any time soon so that we could release this breaking change as v2.0.0?

alexander-fenster:
(asking specifically about Node 4 support for Spanner, but I remember the conversation when you stated it would be great to have all libraries be on the same page here, so just reiterating)

anup-honey:
I have almost completed the code changes that will be backward compatible so that it does not have to be a major version. But then the changes that I am adding have to be removed when you guys do a major upgrade.

Let me know your thoughts so that I know whether to continue with pushing my changes.

Thanks a lot for the help

vkedia:
I think we should avoid making breaking changes in the library. What exactly are the breaking changes ans why are they needed?

anup-honey:
I have another PR opened and I will merge. But before that, please have a look at it.
https://github.com/honeyscience/nodejs-spanner/pull/6

This will be backward compatible and once we make a major update, we can update the repo to make a breaking change.

Also, it handles the idle sessions differently. We do not send a keep-alive message anymore.
 - Whenever we borrow a session, we check when the session was last used. Based on that, either we use that session or destroy it.
- The code also destroys any session that has been idle for a certain time (provided in parameters, the default is 50).

We will never have a session that has been open for days/months as sessions will be destroyed and created based on idle timeouts. If there is always load then the sessions will not be destroyed. 

Once we have internally reviewed the PR, I will update this PR with those changes.

vkedia:
What is the justification for not keeping the sessions alive?

callmehiphop:
Initially we used the same pooling library before, but it didn't quite accommodate all the features requested from the spanner team, which is where the decision to write our own came from. I'm ok with switching back if we can make it work, or if the Spanner team (@vkedia @snehashah16) is ok with changes in functionality being proposed.

anup-honey:
The code now sends KeepAlive messages as before. I have reverted those changes.

walshie4:
@anup-honey probably the same line missing coverage as previously (`setTimeout` function).  Should probably also rebase this now that `master` has moved

callmehiphop:
Where do we stand on this? I started to work through the code and it looks like this still has a few breaking changes and some missing functionality. 

- acquireTimeout
  - Default value has been changed from Infinity to 0
- concurrency
  - Concurrency no longer controls how many concurrent requests the pool makes, but instead acts as an acquire queue
- writes
  - Support appears to be missing in favor of minWrites/maxWrites
  - appears default went from 0 to 20%
- get next available session when desired type is unavailable
  - in the old code if a read session was requested but none were available, we would return a write session in its place (and vice versa). It looks like this is no longer the case and we simply wait for a session of the requested type to become available.

anup-honey:
@callmehiphop 

acquireTimeout
           The default value has been changed from Infinity to 0. Yes, but the behavior is same. Here 0, indicated infinity. 
           Ans - If you want, I can easily change that.

concurrency
            Concurrency no longer controls how many concurrent requests the pool makes, but instead acts as an acquire queue. 
             Ans - We need a session to make a request. So if we throttle getting a session, we indirectly throttle our requests. Also, in the previous code, requestQueue_ is used only in 3 places, createSession, destorySession and pingSession.

writes
           Support appears to be missing in favor of minWrites/maxWrites appears default went from 0 to 20%. 
          Ans - This is correct. To change this, we will have to rewrite session pool logic and it will get complicated. To make the code simpler, this decision was taken. 

get next available session when the desired type is unavailable in the old code if a read session was requested but none were available, we would return a write session in its place (and vice versa). It looks like this is no longer the case and we simply wait for a session of the requested type to become available.
          Ans - We can implement this functionality, when asked for a read session and no read session is available, we grab a write session. But when we ask for a write session, if we return a read session, it might not work as the code was expecting a write session. We can definitely convert a read session to a write session on the fly and convert it back when the task is completed.

anup-honey:
@callmehiphop 

acquireTimeout

           The default value has been changed from Infinity to 0. 
           Ans - Yes, but the behavior is same. Here 0, indicated infinity. If you want, I can easily change that.

concurrency

            Concurrency no longer controls how many concurrent requests the pool makes, but instead acts as an acquire queue. 
             Ans - We need a session to make a request. So if we throttle getting a session, we indirectly throttle our requests. Also, in the previous code, requestQueue_ is used only in 3 places, createSession, destorySession and pingSession.

writes

          Support appears to be missing in favor of minWrites/maxWrites appears default went from 0 to 20%. 
          Ans - This is correct. To change this, we will have to rewrite session pool logic and it will get complicated. To make the code simpler, this decision was taken. 

get next available session when the desired type is unavailable in the old code if a read session was requested but none were available, we would return a write session in its place (and vice versa). It looks like this is no longer the case and we simply wait for a session of the requested type to become available.

          Ans - We can implement this functionality, when asked for a read session and no read session is available, we grab a write session. But when we ask for a write session, if we return a read session, it might not work as the code was expecting a write session. We can definitely convert a read session to a write session on the fly and convert it back when the task is completed.

callmehiphop:
@anup-honey 

> Yes, but the behavior is same. Here 0, indicated infinity. If you want, I can easily change that.

I'm ok with adding support for 0, but lets just keep support for `Infinity` to make it backwards compatible.

> We need a session to make a request. So if we throttle getting a session, we indirectly throttle our requests. Also, in the previous code, requestQueue_ is used only in 3 places, createSession, destorySession and pingSession.

I don't think this is right. As the code is now even if acquires are throttled, that doesn't mean the user can't acquire more sessions than the specified concurrency value. If that happens there would be no mechanism to throttle all of the concurrent requests. The original request for a concurrency knob was to throttle the initial creation of sessions. IIRC creating too many sessions too quickly will result in a rate limiting error.

> This is correct. To change this, we will have to rewrite session pool logic and it will get complicated. To make the code simpler, this decision was taken.

That might be an issue because this would be a breaking change. I need a Googler to chime in here (@JustinBeckwith), but I think we want to avoid any breaking changes since this is client is already at 1.0.0.

> But when we ask for a write session, if we return a read session, it might not work as the code was expecting a write session. We can definitely convert a read session to a write session on the fly and convert it back when the task is completed.

Currently we do just that, so let's continue to do this moving forward.

anup-honey:
@callmehiphop 
I have addressed all the issues including the 'writes'.
Now the session pool will not create 20% of write session by default.

I need some help in fixing the test cases. All the test cases work fine on my local machine but once it is deployed, one of the test cases fails. I would like your help so that I do create a lot of commits to fix a test case.

callmehiphop:
@anup-honey sure thing! What can I do to help?

anup-honey:
@callmehiphop 
Thanks for the response. There is one test case that is failing in circleci. When I run it locally on my machine, all the test cases work fine. 

I feel there is some timing issue (race condition) that I cannot replicate on my local machine. If I can get access to the box where circleci is running, I can ssh into it and try to find the reason.

Let me know if you think of a better way.

anup-honey:
@callmehiphop 
Also, we ran the circleci script in our circleci environment and all the test cases passed.

anup-honey:
@callmehiphop @crwilcox @alexander-fenster 
We need to get this resolved soon. We need to either get this PR merged or you guys want to come up with a different solution. We have been running this code in our production environment for about a month now. This can/will affect our migration effort to spanner and we are running under very tight deadlines.

alexander-fenster:
cc: @JustinBeckwith

src/session-pool.js
```diff
@@ -377,151 +307,75 @@ SessionPool.prototype.requestStream = function(config) {
...
-      session.type = READWRITE;
-
-      return self.createTransaction_(session).catch(function() {
-        session.type = READONLY;
-      });
-    })
-    .then(function() {
-      return session;
-    });
+  return self.acquireQueue_.add(() => self.readPool.acquire());
```
callmehiphop: I think if we get this far we want to check that there are actually read sessions available before acquiring, if not then we want to watch for resources being freed up in either pool and acquire the first available. The same logic should be applied to `getWriteSession`.

src/database.js
```diff
@@ -251,12 +257,6 @@ Database.prototype.batchTransaction = function(identifier) {
...
   var leakError = null;
-  var leaks = this.pool_.getLeaks();
-
-  if (leaks.length) {
-    leakError = new Error(`${leaks.length} session leak(s) found.`);
-    leakError.messages = leaks;
-  }
 
   this.parent.databases_.delete(key);
   this.pool_.close().then(() => callback(leakError), callback);
```
callmehiphop: Removing this functionality would be a breaking change, can you please re-implement it?

src/session-pool.js
anup-honey: I have made these changes and will push my changes

src/database.js
anup-honey: This functionality is not required as genericPool manages everything for us. I can implement something to "make it work" but that will be counter productive.

anup-honey: I have implemented the leaks functionality.

anup-honey:
@callmehiphop 
I have fixed the test cases too.

src/database.js
```diff
@@ -250,16 +256,14 @@ Database.prototype.batchTransaction = function(identifier) {
...
-    leakError = new Error(`${leaks.length} session leak(s) found.`);
-    leakError.messages = leaks;
-  }
-
   this.parent.databases_.delete(key);
-  this.pool_.close().then(() => callback(leakError), callback);
+  this.pool_.close().then(count => {
+    let error = null;
+    if (count > 0) {
+      error = new Error(`${count} session leak(s) found.`);
```
callmehiphop: Is it possible to track the call stacks for the leaks in question?

walshie4: 👍  Should be simple to re-add this trace map setup for use here.

test/database.js
```diff
@@ -169,17 +169,6 @@ describe('Database', function() {
       assert.strictEqual(database.pool_.calledWith_[1], POOL_OPTIONS);
     });
 
-    it('should re-emit SessionPool errors', function(done) {
```
callmehiphop: I know generic-pool does not emit generic error events (I don't think at least). It *does* appear to emit `factoryCreateError` and `factoryDestroyError`, I think it would be in the users interest to emit these errors.

test/database.js
```diff
@@ -331,26 +331,6 @@ describe('Database', function() {
           done();
         });
       });
-
-      it('should report session leaks', function(done) {
```
callmehiphop: Looks like this test is still relevant.

src/database.js
anup-honey: I have added this functionality.

anup-honey:
@callmehiphop 
I have incorporated your comments in the PR. Please let me know if you have any more comments

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
-      assert.strictEqual(sessionPool.request_, DATABASE.request);
-      assert.strictEqual(sessionPool.requestStream_, DATABASE.requestStream);
-    });
+        it('should override DEFAULT for writes', function() {
+          return sessionPool.close().then(function() {
+            sessionPool = new SessionPool(DATABASE, {writes: 0.5});
+            assert.strictEqual(sessionPool.options.acquireTimeout, Infinity);
+            assert.strictEqual(sessionPool.options.concurrency, 10);
+            assert.strictEqual(sessionPool.options.maxWait, 50);
+            assert.strictEqual(sessionPool.options.idlesAfter, 50);
```
callmehiphop: Can we remove any assertions that do not pertain to the test? Seems like concurrency/etc. shouldn't be influenced by the `writes` option.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+            assert.strictEqual(sessionPool.options.minWrites, 0);
+          });
+        });
 
-    it('should localize database request functions', function() {
-      assert.strictEqual(sessionPool.request_, DATABASE.request);
-      assert.strictEqual(sessionPool.requestStream_, DATABASE.requestStream);
-    });
+        it('should override DEFAULT for writes', function() {
+          return sessionPool.close().then(function() {
```
callmehiphop: I just noticed that this seems to be a common pattern throughout the tests, I would prefer we remove the call to close. I don't think it does, but if for some reason in the future close were to modify option values, we might see a lot of failures and wouldn't understand why.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+              min: 70,
+            });
+            assert.strictEqual(sessionPool.options.acquireTimeout, Infinity);
+            assert.strictEqual(sessionPool.options.concurrency, 10);
+            assert.strictEqual(sessionPool.options.maxWait, 50);
+            assert.strictEqual(sessionPool.options.idlesAfter, 50);
+            assert.strictEqual(sessionPool.options.max, 100);
+            assert.strictEqual(sessionPool.options.writes, 0.5);
+            assert.strictEqual(sessionPool.options.min, 70);
+            assert.strictEqual(sessionPool.options.maxReads, 40);
```
callmehiphop: I understand why we change the value, but I think it might be better to throw an error in the event the user provides us with a bunch of wonky values.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+            assert.strictEqual(sessionPool.options.maxWrites, 60);
+            assert.strictEqual(sessionPool.options.minReads, 40);
+            assert.strictEqual(sessionPool.options.minWrites, 60);
+          });
+        });
 
-      sessionPool = new SessionPool(DATABASE);
-      assert.deepEqual(sessionPool.acquireQueue_, {
-        concurrency: 1,
+        it('should override DEFAULT for minReads and minWrites', function() {
```
callmehiphop: Looks like we have duplicates of this test (and perhaps a couple others). Is there a reason for it? If we're testing for something specific here can we alter the test description to reflect that?

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+        assert.strictEqual(sessionPool.getStats().readPool.size, 0);
+        assert.strictEqual(sessionPool.getStats().writePool.size, 0);
       });
-    });
 
-    it('should create handles for intervals', function() {
-      assert.strictEqual(sessionPool.evictHandle_, null);
-      assert.strictEqual(sessionPool.pingHandle_, null);
-    });
+      it('read and write pools should be initialized with the parameters passed in', function() {
```
callmehiphop: Can we break this up into 2 separate tests, one for writes and one for reads.

src/session-pool.js
```diff
@@ -109,221 +113,178 @@ function SessionPool(database, options) {
...
-  this.borrowed_ = [];
-
-  this.pendingCreates_ = 0;
-  this.acquireQueue_ = new PQueue({concurrency: 1});
   this.requestQueue_ = new PQueue({concurrency: this.options.concurrency});
-
-  this.evictHandle_ = null;
-  this.pingHandle_ = null;
-
+  this.acquireQueue_ = new PQueue({concurrency: 1});
```
callmehiphop: I think we're missing a test for this.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+        };
 
-      var expected = `Session leak detected!\n    at ${fakeFunction} (${file})`;
-      var actual = SessionPool.formatTrace_(fakeTrace);
+        sessionPool.session_ = function() {
+          return fakeSession;
+        };
 
-      assert.strictEqual(expected, actual);
+        return sessionPool.createReadSession().catch(error => {
```
callmehiphop: We need to handle the non-failure case here, otherwise the test can potentially give a false positive.

```js
return sessionPool.createReadSession(
  () => { throw new Error('Should not pass.') },
  err => { ... }
);
```

src/session-pool.js
```diff
@@ -550,399 +508,169 @@ SessionPool.prototype.createTransaction_ = function(session, options) {
...
  */
-SessionPool.prototype.getIdleSessions_ = function() {
-  var idlesAfter = this.options.idlesAfter * 60000;
-  var sessions = this.reads_.concat(this.writes_);
-
-  return sessions.filter(function(session) {
-    return Date.now() - session.lastUsed >= idlesAfter;
+SessionPool.prototype.createReadSession = function() {
+  const self = this;
+  return self.requestQueue_.add(() => {
```
callmehiphop: We're missing a test to ensure the request is queued.

src/session-pool.js
```diff
@@ -550,399 +508,169 @@ SessionPool.prototype.createTransaction_ = function(session, options) {
...
-  }
-
-  return Promise.all(
-    evicted.map(function(session) {
-      return self.destroySession_(session);
-    })
-  );
+SessionPool.prototype.createWriteSession = function() {
+  const self = this;
+  return self.requestQueue_.add(() => {
```
callmehiphop: Missing a test to assert that the request is being queued.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+        };
+
+        sessionPool.session_ = function() {
+          return fakeSession;
+        };
 
-    it('should emit a close event', function(done) {
-      sessionPool.on('close', done);
-      sessionPool.close();
+        return sessionPool.createWriteSession().catch(error => {
```
callmehiphop: Need to add a success callback that throws an error to prevent false positives.

src/session-pool.js
```diff
@@ -109,221 +113,178 @@ function SessionPool(database, options) {
...
+  return Promise.resolve(flag);
+}
 /**
  * Opens the pool, filling it to the configured number of read and write
  * sessions.
  *
- * @emits SessionPool#open
  * @return {Promise}
  */
 SessionPool.prototype.open = function() {
```
callmehiphop: This function looks to be largely untested. I only see a test for the `isOpen` flag. Can we add tests for starting the read/write pools and a test to ensure the correct values are passed to the `setTimeout` call

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+          clearTimeout(sessionPool.pingTimeoutHandle);
+          return sessionPool.close();
+        });
+      });
     });
 
-    it('should call stopHouseKeeping_', function(done) {
-      sessionPool.stopHouseKeeping_ = done;
-      sessionPool.close();
+    describe('close', function() {
```
callmehiphop: There's a lot going on in this test, can we break it down into multiple test cases?

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+          });
+        });
+      });
     });
 
-    it('should destroy all sessions', function() {
-      var fakeReads = [{}, {}];
-      var fakeWrites = [{}];
-      var fakeBorrowed = [{}, {}, {}];
+    describe('getReadSession and getWriteSession with minReads and minWrites', function() {
```
callmehiphop: Can we break this into multiple describe blocks?

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
     });
 
-    it('should settle once all sessions are destroyed', function() {
-      var delay = 500;
-
-      sessionPool.reads_ = [{}];
-      sessionPool.destroySession_ = function() {
-        return new Promise(function(resolve) {
-          setTimeout(resolve, delay);
+    describe('release', function() {
```
callmehiphop: Can we make a test for each type of resource (read, read/write) that ensures release is called against the correct underlying pool?

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+          ]);
         });
-      };
-
-      var end = timeSpan();
+      });
 
-      return sessionPool.close().then(function() {
-        assert(isAround(delay, end()));
+      it('should throw an error while release a write session', function() {
```
callmehiphop: Maybe rename this to something more descriptive like "should destroy the session when createTransaction fails"

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+        };
 
-      sessionPool.createSessionInBackground_ = function(type) {
-        created[type] += 1;
-      };
-    });
+        sessionPool.release = common.util.noop;
+      });
+
+      it('should get a session', function(done) {
```
callmehiphop: What is this testing? We don't assert that the session is returned or that the options are passed anywhere.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+          done();
+        });
+      });
 
-    it('should respect the current size of the pool', function() {
-      sessionPool.options.writes = 0.5;
-      sessionPool.reads_ = [{}];
-      sessionPool.writes_ = [{}, {}];
-      sessionPool.fill();
+      it('should call the method with the session', function(done) {
```
callmehiphop: Can we update the description? It looks like we're just testing that the session name is applied to the request options.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
-        assert.strictEqual(session, fakeSession);
+        sessionPool.release = common.util.noop;
       });
-    });
-  });
 
-  describe('getWriteSession', function() {
-    it('should call through to acquireSession_', function() {
-      var fakeSession = {};
+      it('should get a session when stream opens', function(done) {
```
callmehiphop: I think you can remove this test, it's kind of unclear what is being tested here and it will be inherently tested elsewhere.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
-        reqOpts: {},
-      };
-
-      sessionPool.getSession = function() {
-        return Promise.resolve(SESSION);
-      };
 
-      sessionPool.release = common.util.noop;
-    });
+      it('should create a txn object', function(done) {
```
callmehiphop: We should be asserting that the transaction object is returned.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
-      };
+        fakeSession.transaction = function(options) {
+          assert.strictEqual(options, fakeOptions);
+          setImmediate(done);
+          return fakeTxn;
+        };
 
-      sessionPool.request(CONFIG, assert.ifError);
-    });
+        sessionPool.createTransaction_(fakeSession, fakeOptions);
```
callmehiphop: We're also missing a test that checks to see if the transaction was started correctly via `begin()`

src/session-pool.js
```diff
@@ -550,399 +508,169 @@ SessionPool.prototype.createTransaction_ = function(session, options) {
...
  * @returns {Promise}
  */
-SessionPool.prototype.onAvailable_ = function() {
-  var self = this;
-
-  return new Promise(function(resolve) {
-    self.once('available', resolve);
-  });
+SessionPool.prototype.destroySession = function(session) {
+  return this.requestQueue_.add(() => session.delete());
```
callmehiphop: Need to test that this is being queued.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+        maxWrites: 5,
+        acquireTimeout: 50,
+      });
+    });
 
-  describe('instantiation', function() {
-    it('should localize the database instance', function() {
-      assert.strictEqual(sessionPool.database, DATABASE);
+    afterEach(function() {
+      clearTimeout(sessionPool.pingTimeoutHandle);
```
callmehiphop: Shouldn't this be handled by the call to `close()`?

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
-    sessionPool = new SessionPool(DATABASE);
-    sessionPool.onClose_ = new Promise(function() {});
+    SessionPool = require('../src/session-pool.js');
   });
 
-  afterEach(function() {
-    pQueueOverride = null;
-    fakeStackTrace.get = stackTrace.get;
-  });
+  describe('tests with no min sessions sessionPool before hook', function() {
```
callmehiphop: I'm not entirely sure I understand what this means, what is this block of tests trying to do here?

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
-      sessionPool.borrowed_ = [{}, {}, {}];
-
-      assert.strictEqual(sessionPool.size(), 5);
+    afterEach(function() {
+      return sessionPool.close();
     });
   });
 
-  describe('acquireSession_', function() {
+  describe('should convert session types', function() {
```
callmehiphop: Can we move these tests for the corresponding methods? Having them in a top level describe block makes them harder to navigate.

test/session-pool.js
```diff
@@ -16,2022 +16,1575 @@
...
+              sessionPool.writePool._config.acquireTimeoutMillis,
+              1
+            );
+          });
+        });
 
-    it('should set isOpen to false', function() {
-      assert.strictEqual(sessionPool.isOpen, false);
-    });
+        it('should apply defaults', function() {
```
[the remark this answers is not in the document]
anup-honey: This is fixed. The test case is required, just the description of the test case was wrong. I have updated it.

test/session-pool.js
anup-honey: We need the assertions to make sure that when we are trying to change one option then only that option is changed and the default value of other options are not affected

test/session-pool.js
anup-honey: Fixed it

test/session-pool.js
anup-honey: Fixed it. There are not duplicates. I updated the description

test/session-pool.js
anup-honey: This is fixed

test/session-pool.js
anup-honey: This is not a new test case. It is copied from existing test cases for session pool. If you want, I can remove them.

test/session-pool.js
anup-honey: This is not a new test case. I just copied it from existing test cases for session pool

test/session-pool.js
anup-honey: Fixed it

test/session-pool.js
anup-honey: Updated

src/session-pool.js
anup-honey: This is implicitly tested by other tests cases.

src/session-pool.js
anup-honey: Added the test case

src/session-pool.js
anup-honey: Added the test case

test/session-pool.js
anup-honey: fixed it

src/session-pool.js
anup-honey: update the test case

test/session-pool.js
anup-honey: This test case does the following:
1. open a session pool
2. check the readPool and writePool size
3. close the session pool
4. verify that the readPool and writePool size is 0

I don't think I can break down any more in this test case

test/session-pool.js
anup-honey: updated

test/session-pool.js
anup-honey: Fixed it

test/session-pool.js
anup-honey: updated the description

src/session-pool.js
anup-honey: added the test case

test/session-pool.js
anup-honey: fixed it

test/session-pool.js
anup-honey: updated the test case

test/session-pool.js
anup-honey: These test cases are not new. These are just copied from existing test cases

test/session-pool.js
anup-honey: Fixed it

anup-honey:
@callmehiphop: I have updated my PR with a bug fix and addressed all the comments.

callmehiphop:
Thanks @anup-honey, we really appreciate all the work you've put into this. I started making some local edits that I was going to submit as a PR to your branch and I stumbled across something that has me a bit concerned.

We've introduced a new option, `maxWait`, which allows the user to specify the maximum number of acquire calls to have at one time. We're also using this as a means to support the `fail` option by setting `maxWait` to 0. After reviewing the generic pool code, it actually looks as though this value must be a positive integer. That being said, I'm not sure there is actually a reliable way of supporting the `fail` option with this pooling library.

anup-honey:
@callmehiphop 

Actually setting the `maxWait` to `0` will work. 

https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L441

I have provided the link of the code in Generic Pool that checks for the condition and rejects when the `_waitingClientsQueue` is equal to or greater than 
`maxWaitingClients`. All these variables are defined and used in Generic Pool code

anup-honey:
@callmehiphop 

You are right, I somehow missed it. 

I tested by setting the value to `0.1` instead of `0` and all the other checks to work fine as required.

Do you think, it is an acceptable solution? Or I can create an issue with Generic Pool library.

anup-honey:
I have created an issue on the Github page for [Generic Pool](https://github.com/coopernurse/node-pool/issues/234)

anup-honey:
@callmehiphop 
I have created a PR to fix Generic Pool.
https://github.com/coopernurse/node-pool/pull/235

This should get merged

crwilcox:
PR on node-pool is merged.

anup-honey:
@callmehiphop 
I have a new PR to fix generic pool.
https://github.com/coopernurse/node-pool/pull/236

callmehiphop:
Closing in favor of #243

PR #203, opened 2018-05-23
Debugging Issue #180

### LINKED ISSUE #180: runStream returning incomplete and out of order rows
#### Environment details

  - OS: Ubuntu 16.04.4 LTS
  - Node.js version: 6.14.1
  - npm version: 5.8.0

#### Steps to reproduce

```
CREATE TABLE accounts (
    recordId STRING(36) NOT NULL,
    account_CID STRING(255),
    account_ok_tabs STRING(255),
    Address STRING(1024),
    Address1 STRING(1024),
    Address1_city STRING(255),
    Address_city STRING(255),
    Email STRING(255),
    fund_to_account STRING(36),
    Number FLOAT64,
    Number1 FLOAT64,
    Number_of_Payments FLOAT64,
    Phone_Number STRING(1024),
    Phone_Number1 STRING(1024),
    Short_Text STRING(255),
    Short_Text1 STRING(255),
    Short_Text4 STRING(255),
    user_to_accounts STRING(36),
) PRIMARY KEY (recordId)
```
```
```
My table has 17349  records

```js
database.runStream(query)
	.on('error', function (err) {
		console.error(err);
		return reject(err);
	})
	.on('data', function (row) {
		let rowJSON = row.toJSON();
			console.log(row);
		}
	})
	.on('end', function () {
		return resolve(records);
	});
```

stephenplusplus:
- [ ] Fixes #180

This is a reproduction script. You can run it with:

- node ./issue-180.js

### DIFF
--- a/src/partial-result-stream.js
+++ b/src/partial-result-stream.js
@@ -106,8 +106,6 @@ function partialResultStream(requestFn, options) {
         }
       }
 
-      rowChunks = [];
-
       if (options.json) {
         formattedRows = formattedRows.map(exec('toJSON', options.jsonOptions));
       }
@@ -165,6 +163,11 @@ function partialResultStream(requestFn, options) {
  */
 partialResultStream.formatRow_ = function(metadata, row) {
   var fields = metadata.rowType.fields;
+
+  if (row.values.length < fields.length) {
+    // This will only print when we end up with a faulty row.
+    console.log('something is wrong here')
+  }
 
   if (row.values.length > fields.length) {
     // More than one row exists. Return an array of formatted rows.
--- a/issue-180.js
+++ b/issue-180.js
@@ -1,0 +1,175 @@
+const assert = require('assert')
+const async = require('async')
+
+const Spanner = require('./src')
+const spanner = new Spanner()
+
+const instance = spanner.instance('issue-180-instance')
+const database = instance.database('issue-180-database');
+const table = database.table('accounts');
+
+const NUM_ROWS_TO_INSERT = 3000
+
+async function init() {
+  await prepareInstance()
+  await prepareDatabase()
+  await prepareTable()
+  await insertData()
+  await insertData()
+  await insertData()
+
+  async.times(5, runQueryBatches, err => {
+    if (err) throw err
+    console.log('Query was successful')
+  })
+
+  function runQueryBatches(_, callback) {
+    async.times(5, runQuery, callback)
+  }
+
+  let invalidRowReceived = false
+
+  function runQuery(_, callback) {
+    const query = {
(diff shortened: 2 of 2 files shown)


src/partial-result-stream.js
```diff
@@ -64,7 +64,7 @@ function partialResultStream(requestFn, options) {
   var batchAndSplitOnTokenStream = checkpointStream.obj({
     maxQueued: 10,
     isCheckpointFn: function(row) {
-      return is.defined(row.resumeToken);
+      return Buffer.isBuffer(row.resumeToken);
```
stephenplusplus: This may or may not have been a problem. It looked like we were sometimes getting empty arrays for "row.resumeToken". This changes to confirm it's an actual resume token.

crwilcox:
Thanks @stephenplusplus  for providing this repro. When I end up in the case of an unusual number of rows.

I am staring at a return that has two extra rows (not 5 as would be required to be a complete entry) and it saying no resume token, no chunkedValue.

A full row is `account_created_on, Short_Text, Short_Text1, account_CID, [0-9]+ (ID)`
```
    {
      "stringValue": "account_created_on",
      "kind": "stringValue"
    },
    {
      "stringValue": "Short_Text",
      "kind": "stringValue"
    }
  ],
  "metadata": null,
  "chunkedValue": false,
  "resumeToken": [],
  "stats": null
}
```

crwilcox:
Upon further looking it seems that we are getting a `row` that isn't `chunkedValue` and doesn't have a `resumeToken` but if you continue the other portion of the incomplete entry is in the next `row`.

So somehow it seem a row object is given that should be listed as chunked but it isn't.

snehashah16:
Java Client has a mechanism for Partial Result Set buffering as well resuming broken streams - https://github.com/GoogleCloudPlatform/google-cloud-java/blob/27c90f1298bb27a618ff59b3cd17321c577f3a6b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L2376

If this is not explicitly handled in the Node client, this could be possibly causing this error.

crwilcox:
Here is the code I added that seems to handle the odd case where we seem to have a chunked row, but chunkedValue isn't true. 

 https://github.com/googleapis/nodejs-spanner/commit/4c054954a79cb04c285ba78b06e421c6063e54e6

crwilcox:
@snehashah16 , as best I can tell, in Java, if we are missing the resumeToken, we know we aren't to th end? I also am not sure where the node client is getting `chunkedValue` on the PartialResultSet object as that doesn't seem to exist in Java.

crwilcox:
@snehashah16 @stephenplusplus my proposed fix is at https://github.com/googleapis/nodejs-spanner/pull/204. I copied some of the Java impl behavior to continue on absence of resumeToken. I also kept a one line fix from @stephenplusplus around detecting the resumeToken on checkpointFn

PR #204, opened 2018-05-23
Handle partial rows in PartialResultSet obj.

### LINKED ISSUE #180: runStream returning incomplete and out of order rows

crwilcox:
Fixes #180 by continuing to form rows if we do not receive a resumeToken. If a PartialResultSet exceeds allowable size it will not contain that value but need to be continued.

I have some tests to update to make this mergeable, but I wanted to stage this so I can point the pertinent threads to it.

- [ ] Tests and linter pass

### DIFF
--- a/src/partial-result-stream.js
+++ b/src/partial-result-stream.js
@@ -70,6 +70,7 @@ function partialResultStream(requestFn, options) {
 
   var rowChunks = [];
   var metadata;
+  var pendingRowValues;
 
   var userStream = streamEvents(
     through.obj(function(row, _, next) {
@@ -79,18 +80,9 @@ function partialResultStream(requestFn, options) {
         metadata = row.metadata;
       }
 
-      // A streamed result set consists of a stream of values, which might
-      // be split into many `PartialResultSet` messages to accommodate
-      // large rows and/or large values. If we are missing the resumeToken
-      // this is likely due to the PartialResultSet hitting size restrictions.
-      // In this case it is also necessary to combine this with the next obj.
-      if (
-        (!row.resumeToken || row.resumeToken.length === 0) &&
-        row.values.length % metadata.rowType.fields.length !== 0
-      ) {
-        rowChunks.push(row);
-        next();
-        return;
+      if (pendingRowValues) {
+        row.values = pendingRowValues.concat(row.values);
+        pendingRowValues = null;
       }
 
       if (row.chunkedValue) {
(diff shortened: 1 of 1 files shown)


src/partial-result-stream.js
```diff
@@ -96,6 +102,16 @@ function partialResultStream(requestFn, options) {
         formattedRows = formattedRows.concat(builder.toJSON());
         rowChunks.length = 0;
       } else {
+        // A streamed result set consists of a stream of values, which might
```
vkedia: You also need to handle the possibility of extra fields in the above if block since the last chunk in a list of chunks might have extra fields.

src/partial-result-stream.js
```diff
@@ -79,6 +80,11 @@ function partialResultStream(requestFn, options) {
         metadata = row.metadata;
       }
 
+      if (pendingRowValues) {
+        row.values = pendingRowValues.concat(row.values);
+        pendingRowValues = null;
+      }
```
vkedia: Unrelated to this change, but below we keep on buffering till chunkedValue is true. That can cause problems since you might buffer a lot. In the pathological case, all but the last PartialResultSet can have this as true. In that case you will buffer the whole stream before yielding anything at all even though you would have already received many complete rows.

stephenplusplus: The `else` clause of the `row.chunkedValue` check is where the rows that are pending are split from the rows that are ready to pass on to the user. Only the columns that were cut short are held back (and become `pendingRowValues`.

crwilcox: I think @vkedia  is referring to how the existing code will keep pushing row to rowChunks while row.chunkedValue is true. And in a pathological case that could be every object. Meaning we would collect all values client side, then combine, make rows, and return all rows at once.

What probably has to be done is we should merge chunks as we can and attempt to send rows as we have them. This is different than how this is currently done.

vkedia: @crwilcox yes that is what I meant and the proposed solution seems good.

Issue #198, opened 2018-05-17
Error: decoded.map is not a function


walshie4:
#### Environment details

  - OS: CoreOS
  - Node.js version: v8.9.4
  - npm version: v6.0.1
  - `@google-cloud/spanner` version: v1.4.1

#### Steps to reproduce

  1. ???
  2. ???

This error is occasionally happening when running the same query as in #197.  

```sql
SELECT * FROM <TABLE> ORDER BY id LIMIT 10000 OFFSET @offset
```
```javascript
const [rows] = await spannerDb.run({
  sql: 'SELECT * FROM <table> ORDER BY id LIMIT @limit OFFSET @offset',
  params: {
    limit: 10000,
    offset: 0,
  },
}, { readTimestamp: timestampBound });
return rows.map(row => row.toJSON()); // Causes error from scrambled data
```

```
TypeError: decoded.map is not a function

        at decodeValue_ (/opt/app/node_modules/@google-cloud/spanner/src/codec.js:145)
        at Object.decode (/opt/app/node_modules/@google-cloud/spanner/src/codec.js:179)
        at (/opt/app/node_modules/@google-cloud/spanner/src/partial-result-stream.js:183)
        at Array.map (<anonymous>)
        at Function.partialResultStream.formatRow_ (partial-result-stream.js:179)
        at (/opt/app/node_modules/@google-cloud/spanner/src/partial-result-stream.js:175)
        at Array.map (<anonymous>)
        at Function.partialResultStream.formatRow_ (partial-result-stream.js:173)
        at DestroyableTransform._transform (/opt/app/node_modules/@google-cloud/spanner/src/partial-result-stream.js:99)
        at DestroyableTransform.Transform._read (_stream_transform.js:184)
        at DestroyableTransform.Transform._write (_stream_transform.js:172)

```

stephenplusplus:
This is likely #180 as well, sadly.

walshie4:
Re-opening this following the discussion in #180.  The issue is currently resulting in an unhandled rejection with the above error which seems to cause the promise from the library to never resolve or reject causing our tasks to freeze up.

crwilcox:
@walshie4 
Sorry for a bit of delay between responses. I have rewritten the core logic of how we stream rows. It is possible it has fixed this issue also now (even though I believe this is not related to the other). If you want to try `npm install googleapis/nodejs-spanner#rewrite-partial-result-stream` you can.

walshie4:
[...]

PR #208, opened 2018-05-29
Rewrite PartialResultStream

### LINKED ISSUE #180: runStream returning incomplete and out of order rows

stephenplusplus:
Fixes #180 and #198 by reworking the way we process streamed rows from PartialResultStream on Spanner. The high-level difference is this

1) use RowBuilder for all streams. We had two code paths previously
2) do not block on returning until we have a non-chunked row. We can hold the final value of a chunked segment and begin returning other data
3) handle partial segments. Occasionally we will get a partial row and shouldn't try to send that to the user.

#### To Dos

- [x] Unit tests

### DIFF
--- a/src/codec.js
+++ b/src/codec.js
@@ -118,7 +118,7 @@ codec.generateToJSONFromRow = generateToJSONFromRow;
  * @private
  */
 function decode(value, field) {
-  function decodeValue_(decoded, type, options) {
+  function decodeValue_(decoded, type) {
     if (is.null(decoded)) {
       return null;
     }
--- a/src/codec.js
+++ b/src/codec.js
@@ -122,9 +122,6 @@ function decode(value, field) {
     if (is.null(decoded)) {
       return null;
     }
-    if (type === undefined) {
-      return decoded;
-    }
 
     switch (type.code) {
       case 'BYTES': {
--- a/src/codec.js
+++ b/src/codec.js
@@ -122,6 +122,9 @@ function decode(value, field) {
     if (is.null(decoded)) {
       return null;
     }
+    if (type === undefined) {
+      return decoded;
+    }
 
     switch (type.code) {
       case 'BYTES': {
--- a/src/codec.js
+++ b/src/codec.js
@@ -118,7 +118,7 @@ codec.generateToJSONFromRow = generateToJSONFromRow;
  * @private
  */
 function decode(value, field) {
-  function decodeValue_(decoded, type) {
+  function decodeValue_(decoded, type, options) {
     if (is.null(decoded)) {
       return null;
     }
@@ -176,7 +176,7 @@ function decode(value, field) {
     return decoded;
   }
 
-  return decodeValue_(commonGrpc.Service.decodeValue_(value), field.type);
+  return decodeValue_(value, field.type);
 }
 
 codec.decode = decode;
--- a/src/partial-result-stream.js
+++ b/src/partial-result-stream.js
@@ -104,7 +104,7 @@ function partialResultStream(requestFn, options) {
         return;
       }
 
-      formattedRows = builder.toJSONnobuild(formattedRows);
+      formattedRows = builder.toJSON(formattedRows);
       if (options.json) {
         formattedRows = formattedRows.map(exec('toJSON', options.jsonOptions));
       }
--- a/src/partial-result-stream.js
+++ b/src/partial-result-stream.js
@@ -74,7 +74,6 @@ function partialResultStream(requestFn, options) {
 
   var userStream = streamEvents(
     through.obj(function(row, _, next) {
-      //console.log("rowcount:" + row.values.length + " firstid:" + row.values[0].stringValue)
       var formattedRows = [];
 
       if (row.metadata) {
@@ -86,15 +85,12 @@ function partialResultStream(requestFn, options) {
         return;
       }
 
-      if (row.values[0].stringValue.endsWith('9197')) {
-        console.log('magic off error')
-      }
       // Use RowBuilder to construct rows returning rows as complete.
-      if (builder == undefined){
+      if (builder === undefined) {
         builder = new RowBuilder(metadata, rowChunks.concat(row));
       } else {
-        builder.chunks = builder.chunks.concat(row)
-        builder.metadata = metadata
+        builder.chunks = builder.chunks.concat(row);
+        builder.metadata = metadata;
       }
       // Build the chunks to rows.
       builder.build();
@@ -103,12 +99,12 @@ function partialResultStream(requestFn, options) {
       formattedRows = formattedRows.concat(builder.collectRows());
 
       // This isn't likely, but we could have zero rows. Continue if so.
-      if (is.empty(formattedRows)){
+      if (is.empty(formattedRows)) {
         next();
         return;
       }
 
-      formattedRows = builder.toJSONnobuild(formattedRows)
+      formattedRows = builder.toJSONnobuild(formattedRows);
       if (options.json) {
         formattedRows = formattedRows.map(exec('toJSON', options.jsonOptions));
       }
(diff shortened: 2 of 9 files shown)


stephenplusplus:
@crwilcox I made some changes, but my tests still pass. If this looks okay to you, I'll move onto unit tests.

src/partial-result-stream.js
```diff
@@ -155,41 +134,4 @@ function partialResultStream(requestFn, options) {
...
-      row.values = valueSet;
-      return partialResultStream.formatRow_(metadata, row);
-    });
-  }
-
-  var formattedRow = row.values.map(function(value, index) {
-    var field = fields[index];
-    return {
-      name: field.name,
-      value: codec.decode(value, field),
```
stephenplusplus: Removing this step led to a system test meltdown. I've re-worked the way we format/decode values in the latest commit. One test is still failing.

crwilcox: the actual code fails also. there are odd values comiing through.
```
code: 'ERR_ASSERTION',
  actual: 'undefinedunt_created_on_     18215',
  expected: 'account_created_on_     18215',
  operator: '==' }
```

stephenplusplus: Yikes. Can you send over the test file?

crwilcox: https://gist.github.com/crwilcox/fa8bb2fb164afdc29747c489ee34ad8a

stephenplusplus:
Status update:

I took over the PR and did some refactors. Wrote unit tests, all was well. One quick check that the system tests work, then 💥 -- I don't remember the exact number, but let's say ~20 tests failed. I reverted my refactors to see if it was something I did, but unfortunately, same errors.

I realized something we didn't carry over when @crwilcox made his changes was a call to `codec.decode`. Since things were re-arranged from the old way, it wasn't exactly clear where it should go. But, I put it back, and **all but one** test fails. And I can't figure that one out.

Basically, I'm hoping @crwilcox can dive back in for a second to see if he has any ideas. I'll leave some comments on this PR next to the behavioral changes I've made since you last touched the code.

src/row-builder.js
```diff
@@ -72,7 +71,7 @@ RowBuilder.formatValue = function(field, value) {
   }
 
   if (field.code !== 'STRUCT') {
-    return value;
+    return codec.decode(value, field);
```
stephenplusplus: Here's where we let codec.decode convert the raw values to our custom values, e.g. `codec.Float`.

src/codec.js
```diff
@@ -118,7 +118,7 @@ codec.generateToJSONFromRow = generateToJSONFromRow;
  * @private
  */
 function decode(value, field) {
-  function decodeValue_(decoded, type) {
+  function decodeValue_(decoded, type, options) {
```
stephenplusplus: TODO: remove `options` (stale code when I was experimenting)

src/row-builder.js
```diff
@@ -46,7 +45,7 @@ RowBuilder.getValue = function(obj) {
   var value = obj;
 
   if (obj && obj.kind) {
-    value = obj[obj.kind];
+    value = commonGrpc.Service.decodeValue_(obj);
```
stephenplusplus: This is `decodeValue_`:

```js
GrpcService.decodeValue_ = function(value) {
  switch (value.kind) {
    case 'structValue': {
      return GrpcService.structToObj_(value.structValue);
    }

    case 'nullValue': {
      return null;
    }

    case 'listValue': {
      return value.listValue.values.map(GrpcService.decodeValue_);
    }

    default: {
      return value[value.kind];
    }
  }
};
```

src/row-builder.js
```diff
@@ -150,17 +156,43 @@ RowBuilder.prototype.build = function() {
...
 /**
- * Transforms values into JSON format.
+ * Flush already complete rows.
  */
-RowBuilder.prototype.toJSON = function() {
-  this.build();
+RowBuilder.prototype.flush = function() {
+  var rowsToReturn = this.rows;
 
+  if (!is.empty(this.rows[0]) && this.currentRow.length !== this.fields.length) {
```
stephenplusplus: @crwilcox While the previous changes I pointed out were trying to preserve old behavior, this is 1/2 changes to the logic that you wrote:

```diff
-if (!is.empty(this.rows)
+if (!is.empty(this.rows[0])
```

This is because `this.rows` should never be empty, as I understand it. When we reset `this.rows`, we set it to: `[[]]`. So `this.rows` will always have an object, and the is-empty check will always be false.

stephenplusplus:
@crwilcox I think we're finally good again. Units, systems, and your revision of the "issue-180.js" file I've checked are working. It would be great if you could confirm!

crwilcox:
Some system tests are failing locally after the edits:

```
 insert & query
        3) should query in callback mode
        4) should query in promise mode
        5) should query in stream mode
```

Coverage numbers:
image

stephenplusplus:
100% line coverage is what we go for. Can you show more details about the system test errors?

crwilcox:
Sure, but the system tests seem to also be an issue on builds before your commits. 

```
3) Spanner
       Tables
         insert & query
           should query in callback mode:

      Uncaught AssertionError [ERR_ASSERTION]: { SingerId: 'gcloud-tests-id-1278e710',
  Name: 'gcloud-tests-name-1278e711',
  Float: null,
  Int: null,
  Info: null,
  Create deepEqual { SingerId: 'gcloud-tests-id-e12752a0',
  Name: 'gcloud-tests-name-e12752a1',
  Float: 8.2,
  Int: 2,
  Info: <Buffer 67 63 6c 6
      + expected - actual

       {
      -  "Accents": [null]
      -  "Created": [null]
      -  "DOB": [null]
      -  "Float": [null]
      -  "HasGear": [null]
      -  "Info": [null]
      -  "Int": [null]
      -  "Name": "gcloud-tests-name-1278e711"
      -  "PhoneNumbers": [null]
      -  "SingerId": "gcloud-tests-id-1278e710"
      +  "Accents": [
      +    "jamaican"
      +  ]
      +  "Created": [Date: 2018-06-04T16:46:51.082Z]
      +  "DOB": [Date: 1969-08-20T00:00:00.000Z]
      +  "Float": 8.2
      +  "HasGear": true
      +  "Info": [Buffer: [
      +    103
      +    99
      +    108
      +    111
      +    117
      +    100
```

stephenplusplus:
@alexander-fenster ran the system tests and had a different failure. I don't believe these are related to our PR here. @callmehiphop can you run these?

I suppose we could always a merge to let Circle have a go at them, then revert it if necessary.

crwilcox:
Reran the tests and they pass on master and on this branch. I think this is safe to merge.

PR #219, opened 2018-06-12
Adding Spanner STRUCT param samples

jabubake:
- [x] Tests and linter pass

### DIFF
--- a/samples/struct.js
+++ b/samples/struct.js
@@ -36,12 +36,12 @@ function writeDataForStructQueries(instanceId, databaseId, projectId) {
   const instance = spanner.instance(instanceId);
   const database = instance.database(databaseId);
 
-  // Instantiate Spanner table objects
+  // Instantiates Spanner table objects
   const singersTable = database.table('Singers');
 
   // Inserts rows into the Singers table
-  // Note: Cloud Spanner interprets Node.js numbers as FLOAT64s, so
-  // they must be converted to strings before being inserted as INT64s
+  // Note: Cloud Spanner interprets Javascript numbers as FLOAT64s.
+  // Use strings as shown in this example if you need INT64s.
   singersTable
     .insert([
       {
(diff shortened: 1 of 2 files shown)


jabubake:
@jsimonweb FYI

samples/struct.js
```diff
@@ -0,0 +1,431 @@
...
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  // Gets a reference to a Cloud Spanner instance and database
+  const instance = spanner.instance(instanceId);
+  const database = instance.database(databaseId);
+
+  // Instantiate Spanner table objects
```
fhinkel: Nit: `Instantiates`? I personally prefer it without the `s`, but it should be consistent in one file.

samples/struct.js
```diff
@@ -0,0 +1,431 @@
...
+  const singersTable = database.table('Singers');
+
+  // Inserts rows into the Singers table
+  // Note: Cloud Spanner interprets Node.js numbers as FLOAT64s, so
```
fhinkel: Suggestion: 

`Note: Cloud Spanner interprets JavaScript numbers as FLOAT64s. Use strings as in this examples if you need INT64s.`

Definitely write JavaScript numbers, not Node numbers.

samples/struct.js
```diff
@@ -0,0 +1,431 @@
...
+        FirstName: 'Benjamin',
+        LastName: 'Martinez',
+      },
+      {
+        SingerId: '9',
+        FirstName: 'Hannah',
+        LastName: 'Harris',
+      },
+    ])
+    .then(() => {
```
fhinkel: You could shorten those to 
`.then( () => console.log('Inserted data.'))

samples/struct.js
```diff
@@ -0,0 +1,431 @@
...
+      console.log('Inserted data.');
+    })
+    .catch(err => {
+      console.error('ERROR:', err);
+    })
+    .then(() => {
```
fhinkel: You could shorten to 

`.then( () => database.close())`

If the function body is a single statement, return is implicit.

jabubake: Done

samples/struct.js
jabubake: Done

samples/struct.js
jabubake: Done

samples/struct.js
jabubake: Done

jabubake:
@fhinkel looks like the appveyor build is still pending for the last hour.
Otherwise, the PR is ready to merge.

PR #232, opened 2018-06-15
refactor: drop dependency on safe-buffer

JustinBeckwith:
I'm like 80% sure we don't need this :) . @stephenplusplus now that we're node.js 6 and up, we don't need this right?


stephenplusplus (approved):

stephenplusplus:
Yeah, I think we're good now.

PR #243, opened 2018-06-25
feature(database): make session pool hot swappable

### LINKED ISSUE #236: feature: allow users to provide custom pool class
I think no one is a stranger to the fact that we've been seeing a fair amount of chatter around issues with our pooling implementation. Its come to my attention that some teams might want to opt out of certain features of our Pooling class or in some cases just have a finer level of control. I would like to propose accepting a Pool constructor that could be used in place of our Pooling class.

I've reviewed our code and with a light refactor I think the bare minimum could be the following interface

```ts
interface PoolInterface extends EventEmitter {
  new (database: Database);
  open();
  close(callback: (err: Error) => void);
  getReadSession(callback: (err: Error, session: Session) => void);
  getwriteSession(callback: (err: Error, session: Session, transaction: Transaction) => void);
  release(session: Session);
}
```

Integrating would be as simple as

```js
class CustomPool extends EventEmitter {
  // ...
}

const database = instance.database('my-db', CustomPool);
```

Thoughts?

callmehiphop:
Fixes #236 (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass


stephenplusplus (approved):

callmehiphop:
@walshie4 I pushed a new commit here that includes a sizeable refactor to the session pool. If you have the bandwidth, could you test it out to see if you still experience any of the leaks/etc. reported in #192?

test/session-pool.js
```diff
@@ -34,7 +33,7 @@ function FakePQueue(options) {
 
 var fakeStackTrace = extend({}, stackTrace);
 
-describe('SessionPool', function() {
+describe.skip('SessionPool', function() {
```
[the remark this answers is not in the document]
callmehiphop: Yes! I need to refactor the unit tests, but I was hoping to get confirmation from the honey team before doing so.

src/session-pool.js
```diff
@@ -75,874 +128,643 @@ var READWRITE = 'readwrite';
...
-  this.isOpen = false;
+  /**
+   * Total number of read sessions.
+   * @type {number}
+   */
+  get reads() {
+    const {readonly, borrowed} = this._inventory;
+    const available = readonly.length;
+    const used = Array.from(borrowed).filter(
+      session => session[SESSION_TYPE] === READWRITE
```
walshie4: should this be `=== READONLY`?

walshie4:
@callmehiphop I'll see if I can find a task to try this out on in the next few days or this weekend.  👍

src/session-pool.js
callmehiphop: oof, yes! copypasta fail

callmehiphop:
@crwilcox @stephenplusplus still working on some unit tests, but this should be ready for a look through!

crwilcox (approved):

JustinBeckwith:
Looking good here @callmehiphop.  Are we good to merge?

src/session-pool.js
```diff
@@ -75,874 +129,662 @@ const READWRITE = 'readwrite';
...
 
-/**
- * Fills the pool with the minimum number of sessions.
- *
- * @return {Promise}
- */
-SessionPool.prototype.fill = function() {
-  const requests = [];
-  let i;
+    if (this.options.writes > 1) {
```
walshie4: should probably check this is `>= 0` too

walshie4: Should the other options have sanity checks as well?  Recently ran into a really weird error from passing a mis-named option, this could help prevent that.

src/session-pool.js
```diff
@@ -75,874 +129,662 @@ const READWRITE = 'readwrite';
...
+      throw new WritePercentError();
+    }
 
-  const minReadWrite = Math.floor(this.options.min * this.options.writes);
-  const neededReadWrite = Math.max(minReadWrite - this.writes_.length, 0);
+    this._inventory = {
+      [READONLY]: [],
+      [READWRITE]: [],
+      borrowed: new Set(),
```
walshie4: curious why a `Set` here but array for the other inventories?

src/session-pool.js
callmehiphop: What kind of error? And which option?

src/session-pool.js
callmehiphop: We want to maintain a specific order for the available options, making sure to give back the recently used sessions first. Borrowed sessions don't matter, so I went with Set to make life easier.

src/session-pool.js
callmehiphop: As far as options go, `keepAlive` is the only other one that might benefit from a sanity check, because if the value is greater than 60 minutes, the sessions will always die.

callmehiphop:
@JustinBeckwith I think we should be good unless @stephenplusplus or @walshie4 have more feedback!

src/session-pool.js
walshie4: Was getting errors relating to invalid arguments being passed to what looks like the grpc request to create a spanner session.  I had been passing an options object with unexpected keys.  I'll submit an issue with this if I can reproduce it.

PR #244, opened 2018-06-25
add synth.py and generate

### LINKED ISSUE #183: Error 14 UNAVILABLE: EOF
#### Environment details

  - OS: CoreOS
  - Node.js version: v8.9.6
  - npm version: v5.6.0
  - `@google-cloud/spanner` version: v1.4.1

#### Steps to reproduce

  Will attempt to create some if needed.

Hello 👋 

I have a set of APIs using this spanner library and as a transfer over a substantial amount of load (a few thousand requests per second) to these APIs I start to get errors from spanner with the following message: `14 UNAVAILABLE: EOF`.

Looking around it seems I should simply have to retry this request to have a good chance at succeeding.  Curious why this is not a case the library handles for me (I believe I saw some mentions of auto-reties for other transient errors somewhere in the docs).  Additionally I would like to better understand what is going on here so I can solve the real underlying issue.

Is this spanner trying to tell me that I need to scale up my cluster in order to provide more reliable throughput?  Or is this simply something I should expect to run into from time to time?

crwilcox:
Fixes #183 (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -178,9 +178,9 @@
       },
       "dependencies": {
         "globals": {
-          "version": "11.5.0",
-          "resolved": "https://registry.npmjs.org/globals/-/globals-11.5.0.tgz",
-          "integrity": "sha512-hYyf+kI8dm3nORsiiXUQigOU62hDLfJ9G01uyGMxhc6BKsircrUhC4uJPQPUSuq2GrTmiiEt7ewxlMdBewfmKQ==",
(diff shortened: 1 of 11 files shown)


JustinBeckwith:
Thanks! You're going to need to update the latest version of `google-gax` for this to work :)

crwilcox:
this passed besides linter changes. I have submitted that commit.

PR #245, opened 2018-06-25
bump version to 2.0.0

crwilcox:
# Changes
- Update to support google-gax v0.17 (#244)
- Drop support for node.js 4.x and 9.x (#226)
- Adding Spanner STRUCT param samples (#219)

# Dependencies
- Remove safe-buffer as dependency(#232)
- Remove generic-pool as dependency (#231)
- Remove lodash.flatten as dependency (#233)
- Remove array-uniq as dependency (#227)
- Remove string-obj-format as dependency (#229)
- Remove methmeth as a dependency (#228)

- [x] Tests and linter pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/spanner",
-  "version": "1.6.0",
+  "version": "2.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/spanner",
-  "version": "1.5.0",
+  "version": "1.6.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -373,7 +373,8 @@
           "dependencies": {
             "align-text": {
               "version": "0.1.4",
-              "bundled": true,
+              "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
+              "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
               "dev": true,
               "requires": {
                 "kind-of": "^3.0.2",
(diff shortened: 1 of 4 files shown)


walshie4:
Doesn't this have to be a major version bump due to dropping node 4.x.x support?

Mentioned: https://github.com/googleapis/nodejs-spanner/pull/194#issuecomment-389299592
PR dropping node 4.x.x:  https://github.com/googleapis/nodejs-spanner/pull/226

crwilcox:
holding this open for a bit while we have a chat about if other changes should go in before marking this 2.0

anupbaldawa:
We have been adding more logic to this PR for backward compatibility. Now, if we are bumping the version to 2.0.0, can I simplify my PR?
https://github.com/googleapis/nodejs-spanner/pull/194

crwilcox:
@anupbaldawa Good question. This would allow breaking surface, though we likely don't want to decrease functionality either. As I don't know the intricacies of the session pool I don't feel confident asserting much here.

@stephenplusplus and @callmehiphop do you have thoughts?

crwilcox:
@callmehiphop if these features were asked for by the spanner team I don't think we want to just drop them. As someone more familiar with the particular PR, is there a way to keep the old functionality while allowing the PR?

stephenplusplus:
> As someone more familiar with the particular PR, is there a way to keep the old functionality while allowing the PR?

I believe we went in a circle, here 😆 (or I misunderstood!). The PR was originally breaking Spanner team-requested behavior. Then, they kindly made the changes to keep the original behavior. In this PR, they asked: "hey, since we're breaking... can we kill all that legacy support?"

So, I think the answer is: No, not unless the Spanner team is into it.

@callmehiphop / @anupbaldawa Is there a list of specific deprecations that we can run by them?

anupbaldawa:
@stephenplusplus You are spot on. Sorry was on a vacation so couldn't reply sooner.

The most important thing I would like to change is the way the library is initialized, the options that are passed in. 

It is not very important as everything has been addressed in the PR. It would just result in cleaner and less complicated code.

The internal functionality has already been implemented based on the comments from the PR review.

JustinBeckwith:
@crwilcox what's the status with this one?

crwilcox:
This is parked pending various work items, notably work @callmehiphop is currently undertaking to address the issues @anupbaldawa has found and addressed in the PR #194. That code is incompatible with changes explicitly asked for by the core spanner team.

JustinBeckwith:
While not ideal, it may make sense to go ahead and ship 2.0 now.  We're starting to pile up out of date dependencies (#268).  Doing a 3.0 a few weeks later is no big deal.

PR #256, opened 2018-07-02
refactor: update auth library, common-grpc

JustinBeckwith:


system-test/spanner.js
```diff
@@ -3554,14 +3554,9 @@ describe('Spanner', function() {
...
 
-          var expectedMessage = [
-            'Row at index 0 does not contain the correct number of columns.',
-            'Missing columns: ["NumberValue"]',
-          ].join('\n\n');
-
           assert.throws(function() {
             transaction.insert(table.name, rows);
-          }, expectedMessage);
+          }, /Error: Row at index 0 does not contain the correct number of columns\.\\n\\nMissing columns: \["NumberValue"\]/);
```
crwilcox: this edit seems odd. Is this out of the generator?

crwilcox (approved):

system-test/spanner.js
JustinBeckwith: Nah.  This was failing on node.js 10 because it was trying to pass a string instead of a regex.  It just means that this test never worked the way people expected it to :)

PR #274, opened 2018-07-30
chore: use assert.deepStrictEqual instead of assert.deepEqual

jkwlui:


JustinBeckwith:
@kinwa91 I'd suggest doing this with @google/repo please :) We need to make this change in all of our repos.

PR #280, opened 2018-08-02
build: customize synth.py

jkwlui:
The nodejs-spanner package has 3 API bundled in, so we need to run artman three times to generate all three APIs. That also mean we have 3 exports that need to be combined into `src/v1/index.js`, so we use `s.replace(..`

See https://github.com/googleapis/nodejs-spanner/pull/277#discussion_r207379079 for more context.


JustinBeckwith (approved):
LGTM.  Want to maybe run this in a follow up PR and check in the results?

PR #283, opened 2018-08-05
chore: use let and const

JustinBeckwith:


stephenplusplus:
I wonder if CLA bot could be disabled on origin branches?

stephenplusplus:
@callmehiphop if we merge this, will it be a headache for any work you're already doing?

crwilcox (approved):

fhinkel:
Yeah, wondering if we should land this before I open another PR? Or land open PRs first?

PR #290, opened 2018-08-08
chore: convert to es6 classes

JustinBeckwith:
Yeah this one is gonna hurt.  The VSCode "convert to es6 class" tool is great, but does not generate great diffs.  

System tests all pass.


JustinBeckwith:
I'll come back around to this one after the next round of PRseses land

PR #297, opened 2018-08-18
chore: use es classes in a few places

JustinBeckwith:
Most of these changes are generated from the VSCode "Convert to ES class" function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ system-test/*key.json
 google-cloud-logging-winston-*.tgz
 google-cloud-logging-bunyan-*.tgz
 package-lock.json
+.vscode
+yarn.lock
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "lodash.merge": "^4.6.1",
     "lodash.snakecase": "^4.1.1",
     "merge-stream": "^1.0.1",
-    "modelo": "^4.2.3",
     "p-queue": "^2.4.2",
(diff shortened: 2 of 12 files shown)


JustinBeckwith:
Woah.  This looks like it's not broken in node 10.8, but it is broken on 10.9 😱 @ofrobots @fhinkel @mylesborins who do we tell about that kind of thing?

.gitignore
```diff
@@ -12,3 +12,5 @@ system-test/*key.json
 google-cloud-logging-winston-*.tgz
 google-cloud-logging-bunyan-*.tgz
 package-lock.json
+.vscode
+yarn.lock
```
fhinkel: Aren't we deleting lock files?

fhinkel:
We shouldn't have breaking changing in a minor release that would cause that kind of errors. I wonder if there's a dependency that changes with a different Node version and causes the difference?

fhinkel:
I can reproduce. `row-builder` tests pass on 10.8, fail on 10.9. There's a lot of V8 changes that went into 10.9. If I have time I can start a bisect of Node to see where it fails.

fhinkel:
No luck bisecting it, my g++ returns errors on most revisions between 10.8 and 10.9 😞

.gitignore
JustinBeckwith: yup!  This is just adding them to the .gitignore.

PR #302, opened 2018-08-23
feat: use es style imports

JustinBeckwith:
BREAKING CHANGE: The import syntax for this library has changed to be [es module](https://nodejs.org/api/esm.html) compliant. 

### Old code
```js
const spanner = require('@google-cloud/spanner')();
// or 
const Spanner = require('@google-cloud/spanner');
const spanner = new Spanner();
```

### New code
```js
const spanner = new Spanner();
```


JustinBeckwith:
👋 @kinwa91 whattup with all the kokoro failures?

Issue #216, opened 2018-06-10
Error 4: Transaction outcome unknown


walshie4:
#### Environment details

  - OS: CentOS
  - Node.js version: v8.9.4
  - npm version: v5.10.0
  - `@google-cloud/spanner` version: v1.4.3

#### Steps to reproduce

  1. ???
  2. ???

Attempting an insert in the same way as described here: https://github.com/googleapis/nodejs-spanner/issues/202#issuecomment-391197402

Code causing the issue for us:
```javascript
// spanner is a Spanner database object ready to be used for operations
// tableName is a String of the desired table's name
// rowData is an object of data to write
export default (spanner, tableName, rowData) => {
  return new Promise((resolve, reject) => {
    // eslint-disable-next-line consistent-return
    spanner.runTransaction((errTrx, dbTrx) => {
      if (errTrx) {
        honeyLogger.error('V3 encountered error inserting', errTrx);
        return reject(errTrx);
      }

      const addedInfo = { storeShardId: getSpannerShardId(rowData.storeId) };
      const insertColumns = Object.assign({}, addedInfo, rowData);

      dbTrx.insert(tableName, insertColumns);
      dbTrx.commit((err) => {
        if (err) {
          dbTrx.end();
          return reject(err);
        }
        return resolve();
      });
    });
  })
  .then(() => rowData);
};
```

I've added better logging to see if I can get a stack but so far this is all I have on the error.

```
code: 4    
   details: "Transaction outcome unknown."    
   message: "Deadline for Transaction exceeded."    
   metadata: {
    _internal_repr: {
    }
   }
   note: "Exception occurred in retry method that was not classified as transient"    
```

crwilcox:
How large is this transaction? Is this a single row update?

walshie4:
Yes it is a single row.

crwilcox:
Alright. I can't imagine a single row being so big that it would exceed deadlines. Just to check assumptions I may have made, the intent of your code is the following:

- Take a row `{ storeId: 'storeidval', field1: 'field1val', ..., fieldn: 'fieldnval' }`
- commit with additional column: `{ storeId: 'storeidval', field1: 'field1val', ..., fieldn: 'fieldnval', storeShardId: 'storeidval'}`

Could you measure how long `getSpannerShardId` takes? That appears to be the only things in this method I don't fully understand the runtime of. If that takes a while it could be causing this update to take an unusually long time.

walshie4:
[...]

PR #309, opened 2018-08-24
Minor: wrap the inner error on retried transactions and return when deadline exceeded

crwilcox:
Addresses some of the concerns from #216. When a transaction is retired beyond the allowed timeout the inner error is lost. This change holds the inner error.


package.json
```diff
@@ -64,6 +64,7 @@
     "ycsb": "node ./benchmark/ycsb.js  run -P ./benchmark/workloada -p table=usertable -p cloudspanner.instance=ycsb-instance -p operationcount=100 -p cloudspanner.database=ycsb"
   },
   "dependencies": {
+    "@google-cloud/common": "^0.23.0",
```
JustinBeckwith: fun fact - you don't need to do that!  This is available on `common-grpc` in the `util` import as well :)

test/transaction.js
```diff
@@ -196,16 +196,14 @@ describe('Transaction', function() {
...
-        code: 4,
-        message: 'Deadline for Transaction exceeded.',
-        a: 'a',
-        b: 'b',
-      };
-
       const formattedError = Transaction.createDeadlineError_(originalError);
 
-      assert.deepStrictEqual(expectedError, formattedError);
+      assert.strictEqual(formattedError.code, 4);
```
stephenplusplus: Can we check that we are returning a "common.util.ApiError"?

crwilcox: Sure. that seems like a reasonable additional check.

src/transaction.js
```diff
@@ -720,10 +726,13 @@ class Transaction extends TransactionRequest {
    * @return {object}
    */
   static createDeadlineError_(err) {
-    return extend({}, err, {
-      code: 4,
-      message: 'Deadline for Transaction exceeded.',
-    });
+    let apiError = new common.util.ApiError(
```
stephenplusplus: Looking at [the code](https://github.com/googleapis/nodejs-common/blob/4a3451f9a51b55b0a655f5f4990dc2564616cbe2/src/util.ts#L172), I think the intended way to call this is:

```js
let apiError = new common.util.ApiError({
  message: 'Deadline for Transaction exceeded.',
  code: DEADLINE_EXCEEDED,
  errors: [err],
});
```

This would return (from REPL):

```js
{ Error: Deadline for Transaction exceeded. - Original error message.
  code: 4,
  errors:
   [ Error: Original error message. ],
  response: undefined,
  message: 'Deadline for Transaction exceeded. - Blah.' }
```

PR #352, opened 2018-10-02
docs: various jsdoc fixes

### LINKED ISSUE #345: Incorrect JSDoc for a variety of methods
I found some discrepancies between implementation and documentation when I've tried to generate a `.d.ts` file based on JSDoc with [@otris/jsdoc-tsd](https://www.npmjs.com/package/@otris/jsdoc-tsd) npm module.

Here are some of the issues:
- [Table#read](https://github.com/googleapis/nodejs-spanner/blob/master/src/table.js#L518) method is not documented (or implemented?) correctly. When the example is consistent with the current behavior there is some discrepancy with the documentation. The documentation says that the returned promise resolves to an array containing an instance of a Table and an array of rows (objects) [TableReadResponse](https://github.com/googleapis/nodejs-spanner/blob/master/src/table.js#L389), but in fact that array only contains rows.
- [BatchTransaction#execute](https://github.com/googleapis/nodejs-spanner/blob/master/src/batch-transaction.js#L228) method's [callback argument](https://github.com/googleapis/nodejs-spanner/blob/master/src/batch-transaction.js#L221) should be optional
- [Instance#setMetadata](https://github.com/googleapis/nodejs-spanner/blob/master/src/instance.js#L625) `object` should start with capital O for [metadata argument](https://github.com/googleapis/nodejs-spanner/blob/master/src/instance.js#L591)
- [Database#run](https://github.com/googleapis/nodejs-spanner/blob/master/src/database.js#L1141) method takes `string` or `object` as the first argument, so probably there should be a separate `@typedef` for that `object`, but I'm not sure.
- For some methods that return promises the type of the resolved value is not specified
- Some types are misspelled. Should be `Error` instead of `error` and `Date` instead of `date`, unless I've missed typedefs for those types.

#### Environment details

  - OS: N/A
  - Node.js version: N/A
  - npm version: N/A

#### Steps to reproduce

```bash
git checkout v2.0.0
npm install
npm install --no-save @otris/jsdoc-tsd typescript
# Convert the `.jsdoc.js` file into JSON
node -e "console.log(JSON.stringify(require('./.jsdoc'),null,2))" > .jsdoc-dts.json
# Generate a `.d.ts` file 
$(npm bin)/jsdoc -c .jsdoc-dts.json -t ./node_modules/@otris/jsdoc-tsd -d spanner.d.ts
$(npm bin)/tsc --noEmit spanner.d.ts
```

Following these steps will guarantee the quickest resolution possible.

Map the errors returned by TypeScript compiler to the actual code and fix the documentation.

There is a way to configure TypeScript compiler to check JavaScript files (it will also pick up JSDoc comments as type annotations), but that could result in too much noise if you only want to check that the documentation is valid.
I'm pretty sure there are some other means to validate JSDoc as well.

Thanks!

PS.: I'd be glad to submit a PR(s) to fix some of this issues.

callmehiphop:
Fixes #345

- [x] Tests and linter pass

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -129,7 +129,6 @@ class Table extends TransactionRequest {
    * @see [StreamingRead API Documentation](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.StreamingRead)
    * @see [ReadRequest API Documentation](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ReadRequest)
    *
-   * @param {string} table The table to read from.
    * @param {ReadStreamRequestOptions} query Configuration object. See
    *     [`ReadRequest`](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ReadRequest).
    * @param {TransactionOptions} [options] [Transaction options](https://cloud.google.com/spanner/docs/timestamp-bounds).
@@ -515,13 +514,13 @@ class Table extends TransactionRequest {
    * region_tag:spanner_read_data_with_storing_index
    * Reading data using a storing index:
    */
-  read(keyVals, options, callback) {
+  read(query, options, callback) {
     const rows = [];
     if (is.fn(options)) {
       callback = options;
       options = null;
     }
-    this.createReadStream(keyVals, options)
+    this.createReadStream(query, options)
       .on('error', callback)
       .on('data', function(row) {
         rows.push(row);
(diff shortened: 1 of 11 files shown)


JustinBeckwith (approved):

stephenplusplus (approved):

src/table.js
```diff
@@ -387,7 +386,6 @@ class Table extends TransactionRequest {
    */
   /**
    * @typedef {array} TableReadResponse
-   * @property {Table} 0 The new {@link Table}.
    * @property {array[]} 1 Rows are returned as an array of object arrays. Each
```
ipotuzhnov: As I understand this typedef describes a tuple (in this case with just one element), so shouldn't the index for the rows be 0?

ipotuzhnov (commented):
I'm not sure about the types like `error` and `date`, but I guess types in jsdoc could be whatever you want.

PR #360, opened 2018-10-04
refactor(transaction): error handling

### LINKED ISSUE #325: spanner SDK is mixing the callback(err) and throw error will cause unhandled promise rejection
#### Environment details

  - OS: COS
  - Node.js version: 8.11.1
  - yarn version: 1.9.4

#### Steps to reproduce

  1. `transaction.end` called
  2. `SessionPool.release` throw `ReleaseError`
https://github.com/googleapis/nodejs-spanner/blob/c5500c7fb2d063bff80fe38589f7adcaad1b3645/src/session-pool.js#L339-L342
  3. `transaction.end` didn't wrap with try catch and didn't passed to callback, causing unhandled promise rejection.
https://github.com/googleapis/nodejs-spanner/blob/c5500c7fb2d063bff80fe38589f7adcaad1b3645/src/database.js#L496-L498

We toke a look into the source code and found that spanner SDK is mixing the callback(err) and throw error. error handling in the underneath spanner sdk is broken and encapsulated. as API user, we could not handle those errors on the public API provided by spanner sdk.

https://github.com/googleapis/nodejs-spanner/blob/c5500c7fb2d063bff80fe38589f7adcaad1b3645/src/transaction.js#L454-L458

this line, it should use callback(new Error) which could be handled by the promisified try catch block, but they use throw new Error which could not be properly handled by promisify utils and yields this unhandled promise rejection error

callmehiphop:
Fixes #325 

- [x] Tests and linter pass

### DIFF
--- a/test/transaction.ts
+++ b/test/transaction.ts
@@ -513,7 +513,7 @@ describe('Transaction', () => {
       });
 
       it('should destroy the transaction', done => {
-        transaction.end = done;
+        transaction.end = () => done();
         transaction.commit(assert.ifError);
       });
 
@@ -951,8 +951,9 @@ describe('Transaction', () => {
       };
 
       let destroyed = false;
-      transaction.end = () => {
+      transaction.end = callback => {
         destroyed = true;
+        callback(null);
       };
 
       transaction.rollback((err, apiResponse_) => {
--- a/test/transaction.ts
+++ b/test/transaction.ts
@@ -440,12 +440,15 @@ describe('Transaction', () => {
       transaction.queuedMutations_ = QUEUED_MUTATIONS;
     });
 
-    it('should throw an error if the transaction was ended', () => {
+    it('should return an error if the transaction was ended', done => {
+      const expectedError = new Error('Transaction has already been ended.');
+
       transaction.ended_ = true;
 
-      assert.throws(() => {
-        transaction.commit(assert.ifError);
-      }, /Transaction has already been ended\./);
+      transaction.commit(err => {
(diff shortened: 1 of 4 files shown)


src/transaction.js
```diff
@@ -270,30 +268,35 @@ class Transaction extends TransactionRequest {
...
+
     this.request(
       {
         client: 'SpannerClient',
         method: 'commit',
         reqOpts: reqOpts,
       },
-      function(err, resp) {
+      (err, resp) => {
         if (!err) {
```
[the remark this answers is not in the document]
callmehiphop: When `end()` is called thats indicates that this transaction is over and that it is safe to release the connection back into the pool. It is possible that a concurrent transaction will trigger an abort error, which in case we need to retry the transaction in its entirety. There's also a possibility that a non-retryable error will come back, at which point you might want to perform a rollback.

JustinBeckwith:
@callmehiphop @crwilcox just checking in on where this one is :)

test/transaction.ts
```diff
@@ -903,12 +898,16 @@ describe('Transaction', () => {
       transaction.id = ID;
     });
 
-    it('should throw if a transaction ID is not set', () => {
+    it('should throw if a transaction ID is not set', done => {
```
stephenplusplus: `should return error`, I believe!

callmehiphop: Doh! Good catch!

callmehiphop:
Seems like CI is choking since merging #400 into master. Waiting on #441 to resolve said issues.

PR #361, opened 2018-10-04
Only run mutations inside of a transaction.

### LINKED ISSUE #336: deleteRows transaction not atomic
#### Environment details

  - OS: mac 10.13.6
  - Node.js version: 8.11
  - npm version: 5.6.0

#### Steps to reproduce

  1. ???
  2. ???

https://github.com/googleapis/nodejs-spanner/blob/2b8cddad7c2ace98431286be7320fc21c2534f7a/src/transaction-request.js#L271-L301

Performed that command up above. I tried to delete around 8000 rows, each having multiple indices/keys, so it went over the 20k mutations per transaction limit. I did not realize this, and ran it. The entire transaction should fail, but it actually deleted as much as it could under the 20k limit, then threw the over 20k mutations/transaction error. Is this expected?

Thanks!

### LINKED ISSUE #347: Outside of Transactions, always use single-use transactions
RE: https://github.com/googleapis/nodejs-spanner/pull/16#issuecomment-399130367

And default `strong: true` to enable strong reads.

@snehashah16 -- correct?

stephenplusplus:
Fixes #336
Fixes #347 

#### To Dos

- [x] Use transactions for write operations
- [x] Unit tests


src/transaction-request.js
```diff
@@ -274,6 +274,17 @@ class TransactionRequest {
    * ];
    */
   deleteRows(table, keys, callback) {
+    if (!this.transaction) {
```
stephenplusplus: @callmehiphop what do you think about this approach?

src/transaction-request.js
```diff
@@ -708,6 +702,17 @@ class TransactionRequest {
    * @returns {Promise}
    */
   mutate_(method, table, keyVals, cb) {
+    if (!this.transaction) {
+      this.database.runTransaction((err, transaction) => {
+        if (err) {
+          callback(err);
```
crwilcox: cb?

src/transaction-request.js
callmehiphop: Seems fine to me! I'm not sure if its within the scope of this PR but we could probably do away with the Table extending the TransactionRequest.

stephenplusplus:
Another case of the "works for me, but not for Kokoro!" :( @callmehiphop could you try out the system tests on this when you have a chance?

PR #366, opened 2018-10-09
Add Cloud Spanner DML/PDML samples.

jsimonweb:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/samples/dml.js
+++ b/samples/dml.js
@@ -237,7 +237,9 @@ function writeAndReadUsingDml(instanceId, databaseId, projectId) {
           VALUES (11, 'Timothy', 'Campbell')`,
       })
       // Queries rows from the Singers table
-      .then(() => transaction.run({sql: `SELECT FirstName, LastName FROM Singers`}))
+      .then(() =>
+        transaction.run({sql: `SELECT FirstName, LastName FROM Singers`})
+      )
       .then(results => {
         const rows = results[0];
         rows.forEach(row => {
--- a/samples/dml.js
+++ b/samples/dml.js
@@ -237,7 +237,7 @@ function writeAndReadUsingDml(instanceId, databaseId, projectId) {
           VALUES (11, 'Timothy', 'Campbell')`,
       })
       // Queries rows from the Singers table
-      .run(`SELECT FirstName, LastName FROM Singers`)
+      .run({sql: `SELECT FirstName, LastName FROM Singers`})
       .then(results => {
         const rows = results[0];
         rows.forEach(row => {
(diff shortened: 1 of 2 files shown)


samples/dml.js
```diff
@@ -0,0 +1,666 @@
...
+    transaction
+      .runUpdate({
+        sql: `UPDATE Singers SET LastName = 'Grant'
+        WHERE STRUCT<FirstName STRING, LastName STRING>(FirstName, LastName) = @name`,
+        params: {
+          name: nameStruct,
+        },
+      })
+      .then(rowCount => {
+        console.log(`Successfully updated ${rowCount} record.`);
```
crwilcox: should this be `records.`? Since record is correct only for 1==rowCount?

samples/dml.js
```diff
@@ -0,0 +1,666 @@
...
+      console.error(err);
+      return;
+    }
+    transaction
+      .runUpdate({
+        sql: `INSERT Singers (SingerId, FirstName, LastName)
+          VALUES (11, 'Timothy', 'Campbell')`,
+      })
+      // Queries rows from the Singers table
+      .run({sql: `SELECT FirstName, LastName FROM Singers`})
```
callmehiphop: This needs to be 

```js
.then(() => transaction.run('SELECT FirstName, LastName FROM Singers'))`
```

samples/dml.js
```diff
@@ -0,0 +1,666 @@
...
+        secondBudget += transferAmount;
+        firstBudget -= transferAmount;
+
+        console.log(firstBudget, secondBudget);
+
+        // Update the database
+        // Note: Cloud Spanner interprets Node.js numbers as FLOAT64s, so they
+        // must be converted (back) to strings before being inserted as INT64s.
+
+        transaction.runUpdate({
```
callmehiphop: Do we want a `return` or `Promise.all()` for these two calls?

jsimonweb: Leaving as is for now as this follows the pattern in transactions.js

samples/dml.js
jsimonweb: This is intended as a single record update.

samples/dml.js
callmehiphop: I might be reading it wrong, but it appears to run two concurrent updates and the promise chain does not wait for either to finish before committing the transaction. The closest I could find in transaction.js to this was the [usage of `update`](https://github.com/googleapis/nodejs-spanner/blob/master/samples/transaction.js#L176) which is a synchronous method whereas `runUpdate` is async.

PR #367, opened 2018-10-09
fix(tests): use unique label for tests

### LINKED ISSUE #365: System tests deletes all running instances of another test run
Currently the system tests are deleting all instances matching `filter: 'labels.gcloud-tests:true'`. This is problematic when we are running parallel tests on the now- Kokoro system-test enabled PRs and continuous integration builds.

We need to label instances with a uuid prefix to distinguish them between builds, and let the `after` hook of Mocha clean up the instances based on the prefix.

kinwa91:
Fixes #365


stephenplusplus:
> We will setup a cron job on Kokoro to periodically clean up instances that are stale.

Should we block this on that? And will the code that does that be here, or something internal?

kinwa91:
@stephenplusplus @callmehiphop cleanup script ready for your review, thanks! :)

.kokoro/cleanup.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Setup service account credentials.
+export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/spanner-key.json
+export GCLOUD_PROJECT=long-door-651
+
+cd $(dirname $0)/..
+
+npm install
+
+npm run cleanup
```
stephenplusplus: How often / when is cleanup.sh run? It would be nice if there was a way to run it from the system tests, if we get an error about too many instances existing.

kinwa91: It's run every hour

PR #373, opened 2018-10-16
feat(session): add label support

callmehiphop:
- [x] Tests and linter pass

### DIFF
--- a/src/database.js
+++ b/src/database.js
@@ -344,14 +344,15 @@ class Database extends ServiceObject {
       callback = options;
       options = {};
     }
-    options = options || {};
+
+    const gaxOpts = extend({}, options);
     const reqOpts = {
       database: this.formattedName_,
     };
 
-    if (options.labels) {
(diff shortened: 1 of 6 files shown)


src/database.js
```diff
@@ -348,6 +348,12 @@ class Database extends ServiceObject {
     const reqOpts = {
       database: this.formattedName_,
     };
+
+    if (options.labels) {
+      reqOpts.session = {labels: options.labels};
+      delete options.labels;
```
stephenplusplus: This modifies the user object, so maybe we should clone `options`?

stephenplusplus (changes requested):

src/database.js
callmehiphop: hey good call! just pushed a new commit

JustinBeckwith:
👋 @stephenplusplus would you mind taking another look?

stephenplusplus (approved):

PR #385, opened 2018-10-22
docs(samples): updated samples code to use async await

vijay-qlogic:
Fixes https://github.com/googleapis/google-cloud-node/issues/2869 (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/samples/batch.js
+++ b/samples/batch.js
@@ -45,13 +45,8 @@ async function createQueryPartitions(
 
   const query = 'SELECT * FROM Singers';
 
-  try {
-    const data = await transaction.createQueryPartitions(query);
-    const partitions = data[0];
-    console.log(`Successfully created ${partitions.length} query partitions.`);
-  } catch (err) {
-    console.error('ERROR:', err);
-  }
+  const [partitions] = await transaction.createQueryPartitions(query);
+  console.log(`Successfully created ${partitions.length} query partitions.`);
   // [END spanner_batch_client]
 }
 
@@ -85,15 +80,8 @@ async function executePartition(
   const database = instance.database(databaseId);
   const transaction = database.batchTransaction(identifier);
 
-  try {
-    const data = transaction.execute(partition);
-    const rows = data[0];
-    console.log(
-      `Successfully received ${rows.length} from executed partition.`
-    );
-  } catch (err) {
-    console.error('ERROR:', err);
-  }
+  const [rows] = await transaction.execute(partition);
+  console.log(`Successfully received ${rows.length} from executed partition.`);
   // [END spanner_batch_execute_partitions]
 }
 
--- a/samples/batch.js
+++ b/samples/batch.js
@@ -15,7 +15,12 @@
 
 'use strict';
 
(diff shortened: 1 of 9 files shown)


samples/batch.js
```diff
@@ -40,21 +45,17 @@ function createQueryPartitions(instanceId, databaseId, identifier, projectId) {
...
-      const partitions = data[0];
-      console.log(
-        `Successfully created ${partitions.length} query partitions.`
-      );
-    })
-    .catch(err => {
-      console.error('ERROR:', err);
-    });
+  try {
+    const data = await transaction.createQueryPartitions(query);
```
JustinBeckwith: Here (and throughout) - it's a lot easier to just destructure the array during assignment:

```js
const [partitions] = await transaction.createQueryPartitions(query);
```

Then you can drop the next line all together :)

samples/batch.js
```diff
@@ -40,21 +45,17 @@ function createQueryPartitions(instanceId, databaseId, identifier, projectId) {
...
+    const partitions = data[0];
+    console.log(`Successfully created ${partitions.length} query partitions.`);
+  } catch (err) {
```
JustinBeckwith: Let's just drop the `try/catch` stuff unless there's a finally we really need to execute.

samples/system-test/spanner.test.js
```diff
@@ -172,493 +172,493 @@ test.serial(`should add a column to a table`, async t => {
...
-  async t => {
-    const results = await tools.runAsyncWithIO(
-      `${dmlCmd} deleteUsingPartitionedDml ${INSTANCE_ID} ${DATABASE_ID} ${PROJECT_ID}`,
-      cwd
-    );
-    const output = results.stdout + results.stderr;
-    t.regex(output, /Successfully deleted 5 records/);
-  }
-);
+// // update_data
```
JustinBeckwith: Heh, assuming an update is coming here :)

JustinBeckwith:
Looks like the system tests are still failin

vijay-qlogic:
Yes, working to fix those failure

JustinBeckwith:
@vijay-qlogic I sent you an invite to the googleapis org.  If you accept that, sample and system tests will run automatically on these PRs :)

vijay-qlogic:
@JustinBeckwith  Yes, I accepted that invitation Thanks!, also enabled 2-factor authentication as well
Few tests are still failing for me, working on it to fix the issues, will update the PR as soon I resolve those

samples/transaction.js
```diff
@@ -122,7 +119,7 @@ function readWriteTransaction(instanceId, databaseId, projectId) {
   const transferAmount = 200000;
   const minimumAmountToTransfer = 300000;
 
-  database.runTransaction((err, transaction) => {
+  database.runTransaction(async (err, transaction) => {
```
JustinBeckwith: Having an async callback is very weird.  Should we just make the top level method this is under async instead?

PR #402, opened 2018-10-26
refactor(typescript): rename src and system-test files to *.ts

JustinBeckwith:
Not perfect by any stretch, and doesn't include tests/, but it's a start.

### DIFF
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -18,8 +18,9 @@
 
 const arrify = require('arrify');
 const common = require('@google-cloud/common-grpc');
-const {paginator} = require('@google-cloud/paginator');
-const {promisifyAll} = require('@google-cloud/promisify');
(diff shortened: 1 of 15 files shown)


src/instance.ts
```diff
@@ -18,8 +18,9 @@
 
 const arrify = require('arrify');
 const common = require('@google-cloud/common-grpc');
-const {paginator} = require('@google-cloud/paginator');
-const {promisifyAll} = require('@google-cloud/promisify');
+import { ServiceObjectConfig } from '@google-cloud/common';
```
crwilcox: nit for spaces

crwilcox (approved):

src/instance.ts
JustinBeckwith: Will fix when we turn the formatter back on :)

PR #417, opened 2018-11-01
chore: improve typescript config and types

JustinBeckwith:
FYI @googleapis/yoshi-nodejs this is currently busted on master, and this *should* fix things.

### DIFF
--- a/src/codec.ts
+++ b/src/codec.ts
@@ -21,7 +21,7 @@ import {Service} from '@google-cloud/common-grpc';
 import * as extend from 'extend';
 import * as is from 'is';
 
-class SpannerDate {
+export class SpannerDate {
   value;
   constructor(value) {
     if (arguments.length > 1) {
@@ -39,7 +39,7 @@ class SpannerDate {
   }
 }
 
-class Float {
+export class Float {
   value;
   constructor(value) {
     this.value = value;
@@ -49,7 +49,7 @@ class Float {
   }
 }
 
-class Int {
+export class Int {
   value;
   constructor(value) {
     this.value = value.toString();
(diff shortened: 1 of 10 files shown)


test/codec.ts
```diff
@@ -479,14 +477,12 @@ describe('codec', () => {
     it('should encode structs', () => {
       const value = codec.Struct.fromJSON({a: 'b', c: 'd'});
       const encoded = codec.encode(value);
-      assert.deepStrictEqual(encoded, ['b', 'd']);
+      assert.deepStrictEqual([].concat(encoded.slice()), ['b', 'd']);
```
JustinBeckwith: @ofrobots I am not wild about this test change, but I can't see how my change above would require it.  Can you spot something with my es6-ification of the struct?

ofrobots: Which change above are you referring to? How was the compiler complaining with the original form? Sorry, I am not familiar with the code here.

src/codec.ts
```diff
@@ -84,73 +84,67 @@ const TYPE = Symbol();
...
-  return struct;
+  /**
+   * Checks to see if the provided object is a Struct.
+   *
+   * @private
+   *
+   * @param {*} thing The object to check.
+   * @returns {boolean}
+   */
+  static isStruct(thing: {}) {
```
ofrobots: Good opportunity for:
```suggestion
  static isStruct(thing: {}) : thing is Struct {
```

src/database.ts
```diff
@@ -156,7 +167,7 @@ class Database extends ServiceObject {
       createMethod: (_, options, callback) => {
         return instance.createDatabase(formattedName_, options, callback);
       },
-    });
+    } as {} as ServiceObjectConfig);
```
kinwa91: Curious why this is needed

kinwa91: Ah, we don't have `requestModule` here. OK for now - we'll have to add types to `common-grpc`

jkwlui (approved):

PR #432, opened 2018-11-12
fix: generateToJSONFromRow should serialize Datetimes

### LINKED ISSUE #431: .toJSON() does not convert Datetimes/Timestamp
When running toJSON on the response of a query in a cloud function, it does not serialises Datetimes (Timestamp in GC Spanner).

This is a problem, when I run my query in a Google Cloud Function (that I call in a webapp with firebase), the Datetime is not serialised and thus not returned to the client.

Cloud Function:
```
        return database.run(query).then(results => {
            const rows = results[0];
            const json = [];
            rows.forEach(row => {
                json.push(row.toJSON());                
            });
            console.log(json); // [ ...,  { name: 'received_at', value: 2018-10-23T22:21:22.000Z }, ...]
            return json;
        });
```

On the client:
```
            const cloudFunction = firebase.functions().httpsCallable('cloudFunction');
            const response = await cloudFunction();
            console.log(response); // {   data: [ ...,  { name: 'received_at', value: {}, ...] }
```

#### Environment details

  - OS: Google Cloud Function
  - Node.js version: 8
  - npm version: 6.4.1
  - `@google-cloud/spanner` version: 2.1

#### Steps to reproduce

  1. Run query on database with Timestamps and call .toJSON() on each row in cloud function
  2. return rows to client
  3. rows with Timestamp have an emtpy object ({}) as value

#### Possible Solution:
https://github.com/googleapis/nodejs-spanner/pull/432

BudickDa:
Fixes #431 (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass


BudickDa:
I signed it!

BudickDa:
I signed it!

src/codec.ts
```diff
@@ -176,6 +176,10 @@ function generateToJSONFromRow(row) {
         }
       }
 
+      if (value instanceof Date) {
+        value = value.toString();
```
callmehiphop: I think we'll probably want to call `toJSON()` instead of `toString()` here. @stephenplusplus have any opinions here?

BudickDa: (new Date()).toJSON() => '2018-11-20T16:08:06.026Z'
(new Date()).toString() => 'Tue Nov 20 2018 17:08:12 GMT+0100 (Central European Standard Time)'

crwilcox:
This PR would result in a breaking change as users that currently expect a `Date` object (and use, for instance, `Date. getTime`) would be interefered with. I think the same thing can be accomplished by stringifying the json via `JSON.stringify`

JustinBeckwith:
Just circling back.  @crwilcox are you saying we should close this one out?

PR #436, opened 2018-11-14
Enable grpc channel management support for spanner client.

WeiranFang:
Enabling grpc connection management for Spanner client using [grpc-gcp](https://github.com/GoogleCloudPlatform/grpc-gcp-node/tree/master/grpc-gcp) library. By default it will be using a channel pool of size 10, and each with 100 max concurrent streams.


package.json
```diff
@@ -53,7 +53,8 @@
     "events-intercept": "^2.0.0",
     "extend": "^3.0.1",
     "google-auth-library": "^2.0.0",
-    "google-gax": "^0.22.0",
+    "google-gax": "^0.22.1",
+    "grpc-gcp": "^0.1.0",
```
JustinBeckwith: There are a few issues here:
- The code in `src/v1` is all auto-generated.  The configs you're bringing in - are these auto-generated via gapic?
- `grpc-gcp` seems to take a dependency on `lodash`.  That package is very large, and impacts load perf.  Can you strip that out?
- Both the `google-gax` and `grpc-gcp` libraries take a dependency on `grpc`, which can get us into diamond dependency trouble.  Is there a way we could just make `grpc-gcp` part of gax? 

Pulling in @alexander-fenster to check here too.

WeiranFang: Thank you Justin for the comments!
1. The config file was not auto-generated. Do you mind share some information on where this file should be put in gapic?
2. Yes I believe I can work around `lodash` by replacing it with some standard methods.
3. Currently only Spanner client is being tested using grpc-gcp extension. Our plan is to gradually support more Cloud APIs in the future, but we don't want to enable grpc-gcp for all the cloud APIs immediately. Thus I don't think we want to make grpc-gcp as part of gax. It should be API specific. WDYT?

JustinBeckwith: Thanks!  On integrating this with gapic/gax, I would chat with @alexander-fenster.  Phasing it in gradually is fine, but really we should find a way to do that through flags or something in `gax`.

WeiranFang: Hi @alexander-fenster , do you mind point me where I should put my json config file for spanner, if it should be generated?

alexander-fenster: Hi @WeiranFang, it's not that easy to add an arbitrary file for the generated code. The code is generated by [gapic-generator](https://github.com/googleapis/gapic-generator) based on [proto and yaml configs](https://github.com/googleapis/googleapis/tree/master/google/spanner) and the generator is the same across all GAPIC libraries.

Having that said, the changes are small enough to make it possible to change gapic-generator (and the future TypeScript generator) to support that. For gapic-generator, [this](https://github.com/googleapis/gapic-generator/blob/master/src/main/resources/com/google/api/codegen/nodejs/main.snip) is the place you want to change, and [these](https://github.com/googleapis/gapic-generator/tree/master/src/test/java/com/google/api/codegen/gapic/testdata/nodejs) are the baseline tests that will likely be affected.

My only concern is the JSON config file. The main question is: is there any possibility to build that config file based on the service configuration (protos and yaml)? If not, it's OK to add it manually, but it must be outside of `src/v1` folder. Maybe `src/` is a good place for it since we're OK with manually editing files under `src/` (but not under `src/v*`).

Do you have ideas if similar configuration files exist in other languages (e.g. in Python)?

alexander-fenster: One way I see it is, if the generated code will check if the configuration file with the predefined name exists, and if it does, load it and set these options. That way it can be brought in for all libraries with no incompatible changes.

WeiranFang: Thanks @alexander-fenster for these information! 

In python, the config file is put [here](https://github.com/googleapis/google-cloud-python/tree/master/spanner/google/cloud/spanner_v1/gapic/transports), And they have this [synth.py](https://github.com/googleapis/google-cloud-python/blob/master/spanner/synth.py) file to partially modify the generated code, to make the grpc-gcp library only work in spanner. 

In php, the config file is put [here](https://github.com/googleapis/google-cloud-php/tree/master/Spanner/src/V1/resources). And the gax has this [line](https://github.com/googleapis/gax-php/blob/master/src/ApiCore/GapicClientTrait.php#L182) to check whether the config file exists, if it exists, enable the channel management.

I personally prefer the php approach, which is also similar to your suggestion. As for where to put our config file, it's really up to you :)

WDYT?

alexander-fenster: Thank you for this investigation! I like the PHP approach but one more question before we make a final decision. 

Do users ever create `SpannerClient` directly? Spanner library has manually written wrappers around autogen'd classes from `v1/` folder so I guess users never do it and all calls to `SpannerClient` constructor is under our control. 

I guess it's constructed either somewhere in manual code or somewhere in `common-grpc`, I don't really know the exact place. @callmehiphop Dave, can you point to the place where `v1.SpannerClient` is actually constructed?

I guess that place would be a good place to read any kind of extra configuration file and populate `opts`. Then the generated (and unchanged!) `SpannerClient` class will just push those `opts`, unchanged, all the way to gax and gax will take care of them.

Does that make sense for  you?

WeiranFang: Thank you! That makes sense, but this won't solve the third bullet point from Justin's initial comment, correct?

alexander-fenster: You're right, I missed the comment. I agree that if we make `grpc-gcp` a part of gax and just export it from gax as is, it will be much easier to handle in future. It should be a quick fix in gax, can you make it or do you want me to do it? :)

WeiranFang: I think I can make the change as long as we have a final decision :) So our idea is, we move all our logic to gax, including the logic to check the existence of the config file? If that's the case, where should we put the config file? And also we will want to change the config file to a more generic name like `grpc_gcp_config.json`, right?

JustinBeckwith:
It looks like the call here is to move this logic over to google-gax :) I am going to close this out for now!

package.json
WeiranFang: Hi @callmehiphop , just a follow-up on this thread. What would be a good place to put this configuration json file? Is `gax-nodejs/src/` a good place to put it?

callmehiphop: Sorry, missed this one somehow. We allow users to specify options via constructor that [we then cache](https://github.com/googleapis/nodejs-spanner/blob/master/src/index.ts#L214) and pass into every gax client we create.

I would probably set a default value for this option in the constructor

```js
this.options = Object.assign({
  [grpc_gcpApiConfigPath]: 'my/value.json'
}, options);
```

crwilcox: If going into google-gax, putting it at `build/src/grpc_gcp_config.json` would match other configs, such as `build/src/operations_client_config.json`

WeiranFang: The config json file is service specific, so spanner service should have a config file like spanner_grpc_config.json, and bigtable will have bigtable_grpc_config.json. So I'm not sure build/src/ is the right place to put those json files. Or should we put it somewhere in the spanner client, then parse and store the json object in one of the spanner options, then gax can check the option for the json object. WDYT?

WeiranFang: Hi @crwilcox @callmehiphop , any idea on our previous discussion? Thanks!

PR #446, opened 2018-11-21
fix(transaction): accept json options in run/runStream

### LINKED ISSUE #374: Cannot use json: true in Transaction.run query options
It appears `transaction.run` ignores the `json` boolean that can be passed into query options. Is this intended? Seeing as transaction.read and database.run works with query.json, I expected `transaction.run` to also allow this flag.

#### Environment details

  - OS: macOS 10.14
  - Node.js version: 8.9.0
  - npm version: 6.4.1

#### Steps to reproduce

  1. Call `transaction.run`
```
transaction.run({
      sql: `SELECT * FROM Singers`,
      json: true,
    });
```
  2. `PartialResultStream` doesn't receive query options in transaction.js
https://github.com/googleapis/nodejs-spanner/blob/01602ab808490c60539ade22cbf13500c595d63b/src/transaction.js#L699-L726
  3. Rows return in the usual array of arrays format instead of JSON. 

It seems like PartialResultStream just needs the additional arg as written in TransactionRequest. `transaction.run` works as expected when the query options are passed in.
https://github.com/googleapis/nodejs-spanner/blob/01602ab808490c60539ade22cbf13500c595d63b/src/transaction-request.js#L225-L228

callmehiphop:
Fixes #374

- [x] Tests and linter pass

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -714,6 +714,10 @@ class Transaction extends TransactionRequest {
           transaction: {},
         },
         codec.encodeQuery(query));
+
+    delete reqOpts.json;
+    delete reqOpts.jsonOptions;
+
     if (this.id) {
       reqOpts.transaction.id = this.id;
     } else {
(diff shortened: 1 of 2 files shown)


callmehiphop:
@JustinBeckwith no idea, I couldn't view them.

PR #449, opened 2018-11-27
refactor: removed async from dependency list

vijay-qlogic:
Fixes [#2883](https://github.com/googleapis/google-cloud-node/issues/2883) (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "@types/lodash.snakecase": "^4.1.4",
     "@types/merge-stream": "^1.1.2",
     "@types/mocha": "^5.2.5",
+    "@types/p-limit": "^2.0.0",
     "@types/p-queue": "^3.0.0",
     "@types/p-retry": "^2.0.0",
     "@types/proxyquire": "^1.3.28",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "lodash.merge": "^4.6.1",
     "lodash.snakecase": "^4.1.1",
     "merge-stream": "^1.0.1",
+    "p-limit": "^2.0.0",
     "p-queue": "^3.0.0",
     "p-retry": "^3.0.0",
     "protobufjs": "^6.8.6",
--- a/package.json
+++ b/package.json
@@ -86,7 +86,6 @@
     "@types/through2": "^2.0.34",
     "@types/time-span": "^2.0.0",
     "@types/uuid": "^3.4.4",
-    "async": "^2.6.1",
     "binary-search-bounds": "^2.0.4",
     "codecov": "^3.0.2",
     "concat-stream": "^1.6.2",
--- a/scripts/cleanup.js
+++ b/scripts/cleanup.js
@@ -15,8 +15,7 @@
  */
 
 const {Spanner} = require('../');
-const async = require('async');
-
+const pLimit = require('p-limit');
 const spanner = new Spanner({projectId: process.env.GCLOUD_PROJECT});
 
 // Delete instances that are 1 hour old.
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -84,7 +84,6 @@
     "@types/through2": "^2.0.34",
     "@types/time-span": "^2.0.0",
     "@types/uuid": "^3.4.4",
-    "async": "^2.6.1",
```
crwilcox: system-test.ts relies on this along with cleanup.js. We can move it dev dependencies though.

JustinBeckwith:
aye, the work here is to strip the use of async out of those files :)

vijay-qlogic:
@JustinBeckwith yes 'll update this PR, working on it, looking for how can I replace async.eachLimit (npm module) with native async/await

JustinBeckwith:
I typically use p-limit or p-queue for that :)

vijay-qlogic:
@JustinBeckwith Thanks! I used p-limit for async.eachLimit

package.json
```diff
@@ -60,6 +60,7 @@
     "lodash.merge": "^4.6.1",
     "lodash.snakecase": "^4.1.1",
     "merge-stream": "^1.0.1",
+    "p-limit": "^2.0.0",
```
JustinBeckwith: You probably want to add `@types/p-limit` as well?

scripts/cleanup.js
```diff
@@ -31,21 +30,15 @@ async function deleteStaleInstances(labelFilter) {
...
-          instance.delete(callback);
-        }, 500); // Delay allows the instance and its databases to fully clear.
-      },
-      (err, res) => {
-        if (err) return reject(err);
-        resolve(res);
-      }
-    );
+  const limit = pLimit(5);
+  const input = [];
```
JustinBeckwith: I think this whole thing can be reduced to: 

```js
await Promise.all(
  filtered.map(instance => limit(instance.delete()))
);
```

system-test/spanner.ts
```diff
@@ -45,15 +45,10 @@ describe('Spanner', () => {
...
-          deleteTestResources,
-          (next) => {
-            instance.create(INSTANCE_CONFIG, execAfterOperationComplete(next));
-          },
-        ],
-        done);
+  before(async () => {
+    await deleteTestResources;
+    await instance.create(INSTANCE_CONFIG);
+    await execAfterOperationComplete;
```
JustinBeckwith: That doesn't look right :) What is `execAfterOperationComplete`?  Is that a promise?

system-test/spanner.ts
```diff
@@ -873,25 +868,13 @@ describe('Spanner', () => {
...
-              session.create(next);
-            },
-          ],
-          done);
+    before(async () => {
+      await database.create({
+        schema: `CREATE TABLE Singers (SingerId STRING(1024) NOT NULL,
+          Name STRING(1024),) PRIMARY KEY(SingerId)`,
+      });
+      await execAfterOperationComplete;
```
JustinBeckwith: Same as above, that don't look right

system-test/spanner.ts
```diff
@@ -4246,29 +4221,23 @@ function execAfterOperationComplete(callback) {
...
-          setTimeout(() => {
-            // tslint:disable-next-line no-any
-            (instance as any).delete(callback);
-          }, 500);  // Delay allows the instance and its databases to fully
-                    // clear.
-        }, done);
-      });
+  const limit = pLimit(5);
+  const input: Array<{}> = [];
+  instances.forEach(instance => {
```
JustinBeckwith: `await Promise.all(instances.map(limit(instance.delete())));`

system-test/spanner.ts
```diff
@@ -4246,29 +4221,23 @@ function execAfterOperationComplete(callback) {
...
+    input.push(limit(async () => await instance.delete()));
+  });
+  return (async () => {
+    return await Promise.all(input);
+  })();
 }
 
-function deleteTestResources(callback) {
-  async.series([deleteTestInstances], callback);
+async function deleteTestResources() {
```
JustinBeckwith: I don't see a reason for this method to exist if that's all it does :)

package.json
vijay-qlogic: image
I face this error on adding "@type/p-limit" 
Not sure if I should add this...

scripts/cleanup.js
vijay-qlogic: Done!

system-test/spanner.ts
vijay-qlogic: It's not a Promise, should not call it this way. 
I'll check if I can promisify this and call then

package.json
JustinBeckwith: Just means you need to import it using `import pLimit = require('p-limit');` which is admittedly confusing.

PR #487, opened 2018-12-20
chore(typescript): add types for spanner gapic

callmehiphop:
While trying to chip away at improving types, I found myself defining a lot of request/response types and seeing a need for them across multiple files. I thought it might be useful to have definition files for our gapic clients (at least until we start generating them in TypeScript).


JustinBeckwith:
Wow @callmehiphop, thanks for doing this!

Issue #223, opened 2018-06-13
Convert to TypeScript


JustinBeckwith:
Currently blocked on support TypeScript generation in GAPIC.

PR #490, opened 2018-12-28
refactor(codec): improve typescript defs

callmehiphop:
Relates to #223

- [x] Tests and linter pass

---

### This PR includes breaking changes!

Because Spanner rows can contain duplicate keys, the default format is an array of objects containing key value pairs.

```js
const row = [
  {
    name: 'myfield',
    value: 'myvalue'
  }
];
```

Row arrays come equipped with a `toJSON()` method that transforms said arrays into a plain old JavaScript object while removing any custom wrappers (e.g. `Spanner.int` becomes `number`). However this was not applied recursively. The breaking changes do the following:

1. Unwrap nested values.
2. Transform Structs into plain old objects.

To elaborate on the second point, Structs behave exactly like rows and they can contain duplicate keys, so we've traditionally returned Arrays. However now they will return pojo's when `toJSON()` is called.

```js
// before
const row = {
  mystruct: [
    {
      name: 'key',
      value: Spanner.int(10)
    }
  ]
};

// after
const row = {
  mystruct: {
    key: 10
  }
};
```

Similarly to wrapped numbers, we should allow the raw struct to be returned by passing a `wrapStructs` option.

```js
row.toJSON({wrapStructs: true});
```


src/batch-transaction.ts
```diff
@@ -119,9 +119,7 @@ class BatchTransaction extends Transaction {
     }
     const reqOpts = codec.encodeQuery(query);
     const gaxOpts = query.gaxOptions;
-    if (gaxOpts) {
```
JustinBeckwith: This seems to be unrelated to type changes.  I dunno if this would break something or not 🤷‍♂️

JustinBeckwith (approved):

JustinBeckwith:
I really don't have a good handle on what this code is doing, so it would be great if @stephenplusplus could review this as well :)

src/batch-transaction.ts
callmehiphop: I ended up just shuffling this bit around to a custom type used for making sql & read requests.
https://github.com/googleapis/nodejs-spanner/pull/490/files/f917648aa5ff45cebf41885b879cebe93f906e1d#diff-5507e931de507ce1dd409414c1c57e1eR456

I think my end goal is to remove it from the codec file entirely but I don't think it can realistically happen until we have better typings across the board

stephenplusplus (approved):

PR #491, opened 2018-12-28
build: check broken links in generated docs

kinwa91:


kinwa91:
> I think the hand written link failures will be resolved per #488 (maybe), but do we have a strategy for all the broken gapic links? I'm kind of assuming we'll need a fix for that across all repos.

The gapic team (@alexander-fenster) said they won't be fixing it so I've just been adding ad-hoc synthtool replacements for broken links. That said, are we ready to merge #488?

PR #503, opened 2019-01-29
Add spanner_grpc_config.json and enable grpc-gcp support for spanner

WeiranFang:
Hi @JustinBeckwith @alexander-fenster,

This is to enable channel pool support for spanner (from previous discussions in googleapis/nodejs-spanner#436).

The channel pooling won't be enabled until https://github.com/googleapis/gax-nodejs/pull/396 is merged and released.

Question: Once gax changes are merged and released, do I pin the new version in this PR or create another PR just for the gax version bump?

Thanks


src/spanner_grpc_config.json
```diff
@@ -0,0 +1,92 @@
+{
```
JustinBeckwith: So is this all essentially fine tuning for each API endpoint we access?  Who is supposed to keep this up to date?

test/index.ts
```diff
@@ -27,6 +27,8 @@ import * as pfy from '@google-cloud/promisify';
 import * as sinon from 'sinon';
 import * as spnr from '../src';
 
+const apiConfig = require('../src/spanner_grpc_config.json');
```
JustinBeckwith: Can you `import` this instead?

src/spanner_grpc_config.json
WeiranFang: Currently it's configured as this to address the max 100 concurrent streams issue, but I think as for who's responsible for future changes in this file, it depends on which parts to change. For example if the affinity key changes, then the service owner needs to change it, and if the grpc channel max throughput changes, then we should make the changes accordingly. Does that sound reasonable?

test/index.ts
WeiranFang: It seems in TS it shows error "[ts] Cannot find module './spanner_grpc_config.json'."

src/spanner_grpc_config.json
crwilcox: My one thought here is if there is a way to combine this with our other configuration at some point. We already specify retry conditions for rpc calls.

WeiranFang: @crwilcox We already maintain a general ApiConfig proto definition for all languages, so I believe it would be good to have a separate config file specifically for this proto. Also FWIK in future more grpc features (like retry logic and regional support) will also be supported in this grpc-gcp library. In that case those grpc related configs can all be put into this config file.. WDYT?

test/index.ts
JustinBeckwith: then ignore me :)

WeiranFang:
Hi @JustinBeckwith , I noticed that there are some tests failed for grpc-js system test. grpc-js is the new pure js grpc package, which currently doesn't support our grpc-gcp extension. Can we ignore the grpc-js test for now for spanner?

alexander-fenster:
@WeiranFang These tests now fail for many packages, not only for Spanner and not only because of the unsupported options. I'm going to eventually make gax work with grpc-js without failures since we'll eventually move from grpc to grpc-js. By that time, you'll need to get the support for your features ported to grpc-js. We expect that to happen around Q2.

Until that time, it's OK to ignore those grpc-js failures. That's why that check is not marked Required in GitHub :)

WeiranFang:
Thanks @alexander-fenster for letting me know :) And is there anything I should do before you merge this PR?

PR #508, opened 2019-01-31
refactor(types): enable noImplicitAny in session-pool.ts

joecervino:
Fixes #<issue_number_goes_here>

- [ ] Tests and linter pass

Closed then opened again to make revisions simpler.


joecervino:
I signed it!

joecervino:
Most of the build fails have this error: ` Config validation failed!`. Any idea what this means?

callmehiphop:
@joecervino I think Kokoro just isn't feeling well today, haha.

src/session-pool.ts
```diff
@@ -193,6 +193,7 @@ const enum errors {
 }
 
 interface SessionInventory {
+  [key: string]: Any;
```
callmehiphop: Why do we need this?

src/session-pool.ts
```diff
@@ -157,9 +158,8 @@ const DEFAULTS: SessionPoolOptions = {
  * Error to be thrown when attempting to release unknown resources.
  */
 export class ReleaseError extends Error {
-  // tslint:disable-next-line no-any
-  resource: any;
-  constructor(resource) {
+  resource: Any;
```
callmehiphop: We could probably just use `object` in place of `any`.

src/session-pool.ts
```diff
@@ -157,9 +158,8 @@ const DEFAULTS: SessionPoolOptions = {
...
+  constructor(resource: Session) {
```
callmehiphop: This is incorrect, this error will only get thrown if whatever the user is attempting to release is not a session. So the value should mirror the `resource` type

src/session-pool.ts
```diff
@@ -712,7 +713,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       throw new Error('No resources available.');
     }
 
-    let removeListener;
+    let removeListener: Function|undefined;
```
callmehiphop: I think if we're calling this function with `!` we could probably just set the type to `Function`.

src/session-pool.ts
callmehiphop: Actually, can we use `unknown` here? Seems fitting, haha.

src/session-pool.ts
joecervino: Because `borrow_`, `evictIdleSessions_`, and `_release` use `this._inventory[type].-` which produce this error: `Element implicitly has an 'any' type because type 'SessionInventory' has no index signature.`, so I added a general index signature 👍

src/session-pool.ts
joecervino: Respect - fixed!

src/session-pool.ts
callmehiphop: Friendly ping! Any thoughts about this @joecervino?

callmehiphop:
@JustinBeckwith could you fight the cla robot for me?

JustinBeckwith:
I fought the bot, and the bot won

PR #517, opened 2019-02-08
feat(timestamp): create new date/timestamp classes

### LINKED ISSUE #494: timestamps returned by the lib lose microsecond precision
Hi,

When querying values from Spanner with the nodejs lib, all timestamps lose microsecond precision (they are truncated to milliseconds).
This makes pagination really painful on fields with commit timestamps, since truncating can create lots of cases where entries are missed/duplicate depending on the order.
Other languages (I tried python) properly keep microsecond precision.
Is this a know issue ?

  - OS: all
  - Node.js version: 10
  - `@google-cloud/spanner` version: 2.0

### LINKED ISSUE #526: `npm test` fails on Timezones that are greater than UTC.
#### Environment details

  - OS: Linux (WSL)
  - Node.js version: 11.9.0
  - npm version: 6.8.0

#### Steps to reproduce

  2. Run `npm test`.

Thanks!

callmehiphop:
Fixes #494
Fixes #526 
Closes #529

- [x] Tests and linter pass

---

### Timestamps

This PR brings a new class for dealing with Cloud Spanner Timestamps. Previously we would return Date objects, but these were only good for millisecond precision. Since timestamps can come from the API in the form of either an ISO string or a protobuf timestamp, the class can handle either allowing users to easily use commit and read timestamps in queries, as well as result timestamps in snapshot bounds.

With ISO string.

```js
const timestamp = Spanner.timestamp('2019-02-08T10:34:29.481145231Z');
```

With protobuf timestamp.

```js
const [seconds, nanos] = process.hrtime();
const timestamp = Spanner.timestamp({seconds, nanos});
```

The Timestamp class extends the native Date class, so users can perform basic operations to the Timestamp if they choose.

```js
timestamp.setNanoseconds(10);

console.log(timestamp.toISOString());
// => "2019-02-08T10:34:29.481145010Z"
```

It will also accept any single value that the Date object would accept in scenarios where micro-precision isn't important.

```js
const timestamp = Spanner.timestamp(Date.now());

console.log(timestamp.toISOString());
// => "1985-06-03T22:37:14.312Z"
```

### SpannerDate

To keep dealing with date times consistent, I've also refactored the SpannerDate class to extend the native Date class. It accepts any single value that the Date class would also accept.

```js
const date = Spanner.date('6-3-1985');

console.log(date.toISOString());
// => "1985-06-03T00:00:00.000Z"

console.log(date.toJSON());
// => "1985-06-03"
```

### DIFF
--- a/src/codec.ts
+++ b/src/codec.ts
@@ -39,8 +39,18 @@ export interface JSONOptions {
   wrapStructs?: boolean;
 }
 
+// https://github.com/Microsoft/TypeScript/issues/27920
+type DateFields = [number, number, number];
+
 /**
- * Date-like object used to represent Cloud Spanner Dates.
+ * Date-like object used to represent Cloud Spanner Dates. DATE types represent
+ * a logical calendar date, independent of time zone. DATE values do not
+ * represent a specific 24-hour period. Rather, a given DATE value represents a
+ * different 24-hour period when interpreted in a different time zone. Because
+ * of this, all values passed to {@link Spanner.date} will be interpreted as
+ * local time.
+ *
+ * To represent an absolute point in time, use {@link Spanner.timestamp}.
  *
  * @see Spanner.date
  * @see https://cloud.google.com/spanner/docs/data-types#date-type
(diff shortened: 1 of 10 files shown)


callmehiphop:
@JustinBeckwith I'm not sure what your feelings are on my Timestamp class, but assuming it is something we would like to move forward with, I think it would be better suited for `@google-cloud/common` since we might be able to utilize it elsewhere (https://github.com/googleapis/nodejs-bigquery/issues/6)

JustinBeckwith:
I'm +1 to exposing nanosecond precision in a reusable way across our libs.  Is common really the right place for it though?  Thinking out loud:
1.  I have a general aversion to the notion of a `common` module, as it usually turns into a dumping ground.  Though I get for many of our modules it makes sense.
2. It sounds like this could be of general use to the rest of the node community.  Maybe it makes sense to have our own individual npm module just for nanosecond precision dates?

Adding @ofrobots and @googleapis/node-team as they will know if there's already something in the ecosystem we should be using, or if what you got here is of enough value to the community to spin out :)

JustinBeckwith:
Oh also @callmehiphop silly question - is this a breaking change?

callmehiphop:
Yeah, technically it is, I hadn't released our latest major breaking change yet so my plan was to ship this with it.

src/codec.ts
```diff
@@ -39,22 +39,306 @@ export interface JSONOptions {
...
+ * @see Spanner.timestamp
+ * @see https://cloud.google.com/spanner/docs/data-types#timestamp-type
+ *
+ * @class
+ * @extends Date
+ *
+ * @example
+ * const timestamp = Spanner.timestamp('2019-02-08T10:34:29.481145231Z');
+ *
+ * @example <caption>With a `google.protobuf.Timestamp` object</caption>
```
stephenplusplus: Should we just call this "an object" instead of the google.protobuf thing? I'm not sure (m)any users are touching that. Or, if I'm wrong, we could link it to the definition of that type, so users can see what options they have. But it kind of just feels like a regular old object that we should notate in the params.

src/codec.ts
```diff
@@ -39,22 +39,306 @@ export interface JSONOptions {
...
+   * @param {number|string} n number to pad.
+   * @param {number} [size=3] Desired string size.
+   * @returns {string}
+   */
+  static padLeft(n: number|string, size?: number): string {
+    const padding = Timestamp.getPadding(n.toString(), size);
+    return `${padding}${n}`;
+  }
+  /**
+   * Pads a number with 0s on the right hand size.
```
stephenplusplus: side*

src/codec.ts
callmehiphop: Thats a good point, I was sort of wondering if we should instead accept a tuple to mirror whats going on with [`process.hrtime()`](https://nodejs.org/api/process.html#process_process_hrtime_time)

ofrobots:
How does this compare to something like [nano-date](https://www.npmjs.com/package/nano-date)?

JustinBeckwith:
😨 
image

ofrobots:
My preference would be a new public module for this. This seems like a feature anybody could take advantage of. IMHO there is not reason to sequester it into the `common` module. Two ways I see this can happen are:

1. See if we can contribute to the existing nano-date module.
1. Create a new module (date-nano?)

callmehiphop:
@stephenplusplus I made some changes to the `SpannerDate` class per #526 if you wouldn't mind giving it another go through D:

PR #527, opened 2019-02-15
Fixes #526: npm test fails on Timezones that are greater than UTC

### LINKED ISSUE #526: `npm test` fails on Timezones that are greater than UTC.

sulaysumaria:
Fixes #526 (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass

### DIFF
--- a/src/codec.ts
+++ b/src/codec.ts
@@ -54,7 +54,14 @@ export class SpannerDate {
     if (is.undefined(value)) {
       value = new Date();
     }
-    this.value = new Date(value!).toJSON().replace(/T.+/, '');
+
+    value = new Date(value!);
+    const year = value.getFullYear();
+    // Since Months start at 0, we need to add 1 when retrieving it.
+    const month = (value.getMonth() + 1).toString().padStart(2, '0');
+    const date = value.getDate().toString().padStart(2, '0');
+
+    this.value = `${year}-${month}-${date}`
   }
 }
 


src/codec.ts
```diff
@@ -54,7 +54,14 @@ export class SpannerDate {
     if (is.undefined(value)) {
       value = new Date();
     }
-    this.value = new Date(value!).toJSON().replace(/T.+/, '');
+
+    value = new Date(value!);
+    const year = value.getFullYear();
+    // Since Months start at 0, we need to add 1 when retrieving it.
+    const month = (value.getMonth() + 1).toString().padStart(2, '0');
```
callmehiphop: Unfortunately we can't use `padStart` since it's not available in Node 6.x. That should be ok though, because it looks like [its optional](https://cloud.google.com/spanner/docs/data-types#date-type) anyways.

sulaysumaria: Even though its optional, we can achieve same functionality by some other logic also. But the test cases are written to return two digit numbers.

If modifying the test cases, then we can remove the padStart.

What do you suggest to do? Modify the test cases or modify the logic?

callmehiphop: I just ran a quick test and it looks like Spanner chokes if we don't supply the padding, so lets modify the logic. It also looks like some of our e2e broke as a result (they compared timestamps), but I'll fix those

sulaysumaria: Okay.

Thanks for the heads up about padStart function. I didn't knew about padStart was not present in Node 6.x. When I started using Node, it was already Node 8.x.

I will create another PR. You can close this PR I think.

callmehiphop: Your choice! I would just add another commit here, we'll end up squashing them anyways

PR #529, opened 2019-02-15
Change the way SpannerDate returns value.

### LINKED ISSUE #526: `npm test` fails on Timezones that are greater than UTC.

sulaysumaria:
Fixes #526 (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass


callmehiphop:
@sulaysumaria so interestingly enough this change will cause an off-by-one error since the Date constructor interprets `Y-m-d` timestamps UTC-0 and those date methods all return local time values. I think the underlying issue here is that the Date object might not be the wrong way to go since it infers a timezone based on the format of the date string. I'm going to hold off on this PR for now and try to open a dialogue on how to move forward with this bug.

PR #544, opened 2019-03-01
docs: Update grammar

romlogic:
Corrected errors like "Execute a DML statements" to "Execute a DML statement", "If the callback is omitted, we'll return a Promise" to "If the callback is omitted, the function returns a Promise". Added punctuation, and made general edits to make sentences simpler, and in the Active voice.


src/transaction.ts
```diff
@@ -141,7 +141,7 @@ export interface RunUpdateCallback {
  *     // Error handling omitted.
  *   }
  *
- *   // should be called when finished with the snapshot
+ *   // It is called when the snapshot finishes.
```
callmehiphop: This should be `It should be called when the snapshot finishes`. Since snapshots don't need to be committed or rolled back, it is up to the user to tell us they're done.

src/transaction.ts
```diff
@@ -114,8 +114,8 @@ export interface RunUpdateCallback {
  * but does not allow writes. Snapshot read-only transactions can be configured
  * to read at timestamps in the past.
  *
- * When finished with the Snapshot, you should call {@link Snapshot#end} to
- * release the underlying {@link Session}. Failure to do can result in a
+ * When finished with the Snapshot, call {@link Snapshot#end} to
+ * release the underlying {@link Session}. Failure to do so, can result in a
```
JustinBeckwith: Not sure the comma here makes sense

src/transaction.ts
```diff
@@ -393,7 +393,7 @@ export class Snapshot extends EventEmitter {
    *   });
    *
    * @example <caption>Alternatively, set `query.json` to `true`, and this step
-   * will be performed automatically.</caption>
```
JustinBeckwith: I actually found the previous wording a little more clear.

src/transaction.ts
```diff
@@ -605,7 +605,7 @@ export class Snapshot extends EventEmitter {
    * });
    *
    * @example <caption>Alternatively, set `query.json` to `true`, and this step
-   * will be performed automatically.</caption>
```
JustinBeckwith: "performs" sounds weird here.  I would prefer it stay the way it was

src/transaction.ts
```diff
@@ -681,7 +680,7 @@ export class Snapshot extends EventEmitter {
    *   // ]
    * });
    *
-   * @example <caption>The SQL query string can contain parameter placeholders.
```
JustinBeckwith: Is it a foregone conclusion that it *will* contain placeholders? The meaning of this text was changed. @callmehiphop

src/transaction.ts
```diff
@@ -651,8 +651,7 @@ export class Snapshot extends EventEmitter {
    * **Performance Considerations:**
    *
    * This method wraps the streaming method,
-   * {@link Snapshot#run} for your convenience. All rows will
-   * be stored in memory before being released to your callback. If you intend
+   * {@link Snapshot#run} for your convenience. All rows are stored in memory before being released to your callback. If you intend
```
callmehiphop: We try to limit docs and code to 80 chars per line. This looks like it went over

src/transaction.ts
callmehiphop: They are definitely optional, this should probably be reverted.

src/transaction.ts
```diff
@@ -780,7 +779,7 @@ export class Snapshot extends EventEmitter {
    *     // All results retrieved.
    *   });
    *
-   * @example <caption>The SQL query string can contain parameter placeholders.
+   * @example <caption>The SQL query string contains parameter placeholders.
```
callmehiphop: Parameters are optional, please revert this.

src/transaction.ts
```diff
@@ -163,22 +163,22 @@ export class Snapshot extends EventEmitter {
    * @type {?(string|Buffer)}
    */
   /**
-   * Whether or not the transaction has been ended. If true, no further requests
-   * should be made and the transaction should be discarded.
+   * Whether or not the transaction has ended. If true, no further requests
+   * are made and the transaction is discarded.
```
callmehiphop: I think this should be reverted, we're basically telling the user if they see this condition that they should discard the transaction object. Transactions have to be manually ended, which is why we say `should`.

src/transaction.ts
```diff
@@ -651,9 +651,8 @@ export class Snapshot extends EventEmitter {
    * **Performance Considerations:**
    *
    * This method wraps the streaming method,
-   * {@link Snapshot#run} for your convenience. All rows will
-   * be stored in memory before being released to your callback. If you intend
-   * on receiving a lot of results from your query, consider using the streaming
+   * {@link Snapshot#run} for your convenience. All rows are stored in memory before being released to your callback.
```
callmehiphop: Definitely running over that 80 char limit here. Does `npm run fix` correct this?

src/transaction.ts
romlogic: Can we say, "If true, make no further requests, and discard the transaction." ? This makes it in the active voice.

romlogic:
Thanks a lot. I left one suggestion in the comment. Please, let me know if
you are ok with it, else, I will revert the changes.

Regards,
Romola.

src/transaction.ts
callmehiphop: SGTM!

PR #547, opened 2019-03-06
chore(deps): use bundled p-queue types

### LINKED ISSUE #546: fix p-queue references
Looks like `PQueue` starting shipping its own types, but they are not playing nicely with our implementation.

callmehiphop:
Fixes #546

- [ ] Tests and linter pass


callmehiphop:
Looks like the latest sinon types are borked.

PR #548, opened 2019-03-06
refactor(ts): enable noImplicitAny on table.ts

### LINKED ISSUE #223: Convert to TypeScript

joecervino:
Fixes #223

- [x] Tests and linter pass


src/transaction.ts
```diff
@@ -93,6 +96,30 @@ export interface RunUpdateCallback {
...
+export type ProtoITransactionOptions = spanner_client.spanner.v1.ITransactionOptions;
+export type ProtoITransaction = spanner_client.spanner.v1.ITransaction;
+export type ProtoTransaction = spanner_client.spanner.v1.Transaction;
+export type ProtoITimestamp = spanner_client.protobuf.ITimestamp;
+export type ProtoBeginTransactionCallback = spanner_client.spanner.v1.Spanner.BeginTransactionCallback;
+export type ProtoIBeginTransactionRequest = spanner_client.spanner.v1.IBeginTransactionRequest;
+export type ProtoIReadOnly = spanner_client.spanner.v1.TransactionOptions.IReadOnly;
+export type ProtoCommitCallback = spanner_client.spanner.v1.Spanner.CommitCallback;
+export type ProtoCommitResponse = spanner_client.spanner.v1.CommitResponse;
+export type ProtoIMutation = spanner_client.spanner.v1.IMutation;
```
joecervino: Was wondering if the above should be put in `src/common.ts` to simplify type sourcing throughout the library. Optimistically speaking, we'd import the proto files once then distribute renamed types as needed from there

joecervino:
You guys are probably already aware of this, but just in case the `TransactionRequest` section of the tests are erroring out with `TypeError: Cannot stub non-existent own property encodeRead`. Might explain the check failures

src/codec.ts
```diff
@@ -16,9 +16,9 @@
 import {Service} from '@google-cloud/common-grpc';
 import {DateStruct, PreciseDate} from '@google-cloud/precise-date';
 import * as arrify from 'arrify';
-import {CallOptions} from 'google-gax';
 import * as is from 'is';
 import {common as p} from 'protobufjs';
+import {google as spanner_client} from '../proto/spanner';
```
callmehiphop: Why the `spanner_client` alias? Up until now we've used `google` everywhere, plus we should always prefer camel casing.

src/common.ts
```diff
@@ -15,3 +15,9 @@ export interface CreateSessionOptions {
 export type GetTimestamp = {
   nanos: number; seconds: number;
 };
+
+export type Schema = string | SchemaObject;
+export type SchemaObject = {statements: string[], operationId?: string};
```
callmehiphop: Lets use `google.spanner.admin.database.v1.IUpdateDatabaseDdlRequest` instead.

src/table.ts
```diff
@@ -16,38 +16,37 @@
...
+export interface CreateTableCallback {
   (err: ServiceError, table?: null, operation?: null, apiResponse?: null): void;
-  (err: null, table: Table, operation: d.Operation,
-   apiResponse: d.GrpcOperation): void;
+  (err: null, table: Table, operation: GaxOperation,
+    apiResponse: DeleteResponse): void;
 }
-
-interface DropTableCallback {
+export type CreateTableResponse = [Table, GaxOperation, database_admin_client.longrunning.Operation];
```
callmehiphop: So just a general note, I'm seeing several places in the code were we reference the grpc message class object (like `longrunning.Operation`) but we actually don't return those anywhere. We should never use these types and instead use the interface counterparts `longrunning.IOperation`

src/table.ts
```diff
@@ -16,38 +16,37 @@
...
+export type CommitResponse = [spanner_client.spanner.v1.CommitResponse];
+export type CommitCallback = spanner_client.spanner.v1.Spanner.CommitCallback;
```
callmehiphop: We can't use the generated callbacks, ever, because they return the grpc message class and we want the interface counterpart.

src/table.ts
```diff
@@ -210,24 +210,27 @@ class Table {
       PartialResultStream {
     const proxyStream = through.obj();
 
-    this.database.getSnapshot(options, (err, snapshot) => {
+    this.database.getSnapshot(options, (err: Error|null, snapshot?: Snapshot|null) => {
```
callmehiphop: Why do we need to do this? Are these types inferred?

src/table.ts
```diff
@@ -764,15 +772,15 @@ class Table {
    * @param {function} callback The callback function.
    */
   private _mutate(
-      method: string, rows: object|object[], callback: s.CommitCallback): void {
-    this.database.runTransaction((err, transaction) => {
+      method: string, rows: object|object[], callback: CommitCallback): void {
+    this.database.runTransaction((err: Error|null, transaction?: Transaction|null) => {
```
callmehiphop: Why can't these types be inferred?

src/transaction.ts
```diff
@@ -93,6 +96,30 @@ export interface RunUpdateCallback {
   (err: null|ServiceError, rowCount: number): void;
 }
 
+export interface CommitRequest {
```
callmehiphop: Can't we just use `ICommitRequest`?

src/transaction.ts
```diff
@@ -93,6 +96,30 @@ export interface RunUpdateCallback {
...
+  session: string;
+  transactionId?: Uint8Array|string|null;
+  singleUseTransaction?: ProtoITransactionOptions;
+  mutations: ProtoIMutation[];
+}
+
+export interface TransactionSelector {
```
callmehiphop: Can't we just use `ITransactionSelector`?

src/transaction.ts
callmehiphop: I think that's ok? Although truthfully we shouldn't need most of these since the callbacks and grpc message classes aren't available through the client.

src/transaction.ts
```diff
@@ -721,7 +748,7 @@ export class Snapshot extends EventEmitter {
   run(query: string|ExecuteSqlRequest,
       callback?: RunCallback): void|RunPromise {
     const rows: Rows = [];
-    let stats;
+    let stats: Any;
```
callmehiphop: I believe this should be `IResultSetStats`

src/codec.ts
joecervino: Hey! Just because there are two other proto files with the same exported `google` namespace, so  the `spanner_client` alias is to differentiate the `spanner.d.ts` file from `spanner_database_admin.d.ts` and `spanner_instance_admin.d.ts`

src/table.ts
joecervino: Yeah I was wondering about that - so would that instead be `GaxOperation` or would you rather it be `longrunning.IOperation`?

src/table.ts
joecervino: That's going to be fun with the rest of these PR's 🤦‍♂️ gotcha, purely manual then

src/table.ts
joecervino: Nope, they aren't inferred

src/table.ts
joecervino: Not exactly sure - VS Code is signaling implicit `any`

src/table.ts
callmehiphop: Yeah, unfortunately :( if we ever get typescript gapics then we should be able to use those, but in the mean time I would just make a generic to try and simplify the types.

```typescript
interface RequestCallback<T> {
  (err: ServiceError|null, response?: T|null): void;
}

export type CommitCallback = RequestCallback<google.spanner.v1.ICommitResponse>;
```

src/transaction.ts
joecervino: Long story short: I wish, but not with how things are currently set up
Explanation: Generated files annotate the `transactionId` property with `?: Uint8Array|null`, but Transaction has an `id` property annotated as `string|UInt8Array` (we assign the `x.transactionId` value to `this.id` in transaction). Creating an interface that mirrors `ICommitRequest` but alters `transactionId` to include `string` seemed like the best way forward

src/transaction.ts
joecervino: Same as `ICommitRequest` ⬆️

src/common.ts
joecervino: 🔥

src/transaction.ts
joecervino: I'll start clearing house then 👍

src/table.ts
joecervino: Good thinking - okay, I'll do that

src/table.ts
callmehiphop: Well I think the full response would be something like

```typescript
export type CreateTableResponse = [Table, GaxOperation, dbAdminClient.longrunning.IOperation];
```

src/table.ts
```diff
@@ -764,15 +769,15 @@ class Table {
...
       if (err) {
         callback(err);
         return;
       }
 
-      transaction[method](this.name, rows);
-      transaction.commit(callback);
+      transaction![method](this.name, rows);
```
joecervino: @callmehiphop Wanted to mention that, when enabling noImplicitAny, an error pops up here saying `Element implicitly has an 'any' type because type 'Transaction' has no index signature.`

Doing something like this works but isn't a good fix:
```
  export class Transaction extends Dml {
+  [k: string]: any;
   commitTimestamp?: PreciseDate;
```

Don't know how you guys want to go about solving this one. This same error pops up in multiple different places in other files including `src/v1/spanner_client.d.ts`

joecervino:
@callmehiphop hey, not too sure what to do about this system-test error:
```
 pack and install
       should be able to use the d.ts:
     Error: Command failed: npm install --unsafe-perm
      at Promise.all.then.arr (node_modules/execa/index.js:231:11)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

callmehiphop:
Fails locally for me as well, here's some additional output from my console

```
node_modules/@google-cloud/spanner/build/src/batch-transaction.d.ts:232:27 - error TS2307: Cannot find module '../proto/spanner'.

232         timestamp: import("../proto/spanner").google.protobuf.ITimestamp | undefined;
                              ~~~~~~~~~~~~~~~~~~

node_modules/@google-cloud/spanner/build/src/common.d.ts:1:49 - error TS2307: Cannot find module '../proto/spanner_database_admin'.

1 import { google as database_admin_client } from '../proto/spanner_database_admin';
                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@google-cloud/spanner/build/src/common.d.ts:2:42 - error TS2307: Cannot find module '../proto/spanner'.

2 import { google as spanner_client } from '../proto/spanner';
                                           ~~~~~~~~~~~~~~~~~~

node_modules/@google-cloud/spanner/build/src/transaction.d.ts:23:42 - error TS2307: Cannot find module '../proto/spanner'.

23 import { google as spanner_client } from '../proto/spanner';
```

joecervino:
Well that was annoying :)

package.json
```diff
@@ -46,9 +47,9 @@
     "pretest": "npm run compile",
     "presystem-test": "npm run compile",
     "proto": "npm run proto:spanner && npm run proto:spanner_instance_admin && npm run proto:spanner_database_admin",
-    "proto:spanner": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/v1/spanner.proto | pbts -o proto/spanner.d.ts -",
-    "proto:spanner_instance_admin": "pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/admin/instance/v1/spanner_instance_admin.proto | pbts -o proto/spanner_instance_admin.d.ts -",
-    "proto:spanner_database_admin": "pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/admin/database/v1/spanner_database_admin.proto | pbts -o proto/spanner_database_admin.d.ts -",
+    "proto:spanner": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/v1/spanner.proto | pbts -i Long -o proto/spanner.d.ts -",
```
joecervino: Justin originally wrote it as `-i long` and doing it either way didn't make much of a difference within this specific use case. Let me know if you want me to change it back, it's no biggie

joecervino:
Yeah, this is a bit weird - I''m passing the `should be able to use the d.ts` test on my local system-test but it's failing on the Kokoro system-test. Any ideas?

callmehiphop:
By chance did you manually copy the proto folder into your build folder? Looks like the `compile` script doesn't copy it there (and it should) and we're not shipping the proto directory per the <kbd>package.json</kbd> `files` list.

joecervino:
Nope, didn't add those manually, but I **was** playing with the compile script before pushing my changes

Sorry, I'm not sure if I'm understanding you correctly - do you mean we shouldn't have `build/proto` in the `files` section of `package.json`? Because whenever I remove it I get the same error you copied from your console earlier 😅

package.json
```diff
@@ -41,14 +42,11 @@
...
     "prepare": "npm run compile",
     "pretest": "npm run compile",
     "presystem-test": "npm run compile",
-    "proto": "npm run proto:spanner && npm run proto:spanner_instance_admin && npm run proto:spanner_database_admin",
-    "proto:spanner": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/v1/spanner.proto | pbts -o proto/spanner.d.ts -",
-    "proto:spanner_instance_admin": "pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/admin/instance/v1/spanner_instance_admin.proto | pbts -o proto/spanner_instance_admin.d.ts -",
-    "proto:spanner_database_admin": "pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/admin/database/v1/spanner_database_admin.proto | pbts -o proto/spanner_database_admin.d.ts -",
+    "proto": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/spanner/v1/spanner.proto google/spanner/admin/instance/v1/spanner_instance_admin.proto google/spanner/admin/database/v1/spanner_database_admin.proto | pbts -o proto/spanner.d.ts -",
```
callmehiphop: Any reason why we couldn't use the local proto files instead of `google-proto-files`?

src/common.ts
```diff
@@ -15,3 +25,23 @@ export interface CreateSessionOptions {
...
+  (err: ServiceError|null, response?: T|null): void;
+}
+
+//tslint:disable-next-line no-any
+export type Any = any;
+
+export type IUpdateDatabaseDdlRequest = dbAdminClient.spanner.admin.database.v1.IUpdateDatabaseDdlRequest;
+export type ITransactionOptions = spClient.spanner.v1.ITransactionOptions;
+export type ITransaction = spClient.spanner.v1.ITransaction;
+export type GeneratedTransaction = spClient.spanner.v1.Transaction;
```
callmehiphop: Should remove this, the non `I` types aren't used anywhere.

src/common.ts
```diff
@@ -15,3 +25,23 @@ export interface CreateSessionOptions {
...
+export type ITimestamp = spClient.protobuf.ITimestamp;
+export type IReadOnly = spClient.spanner.v1.TransactionOptions.IReadOnly;
+export type ICommitResponse = spClient.spanner.v1.ICommitResponse;
+export type IMutation = spClient.spanner.v1.IMutation;
+export type LongrunningIOperation = dbAdminClient.longrunning.IOperation;
```
callmehiphop: Any reason we don't just call this `IOperation`?

src/table.ts
```diff
@@ -16,38 +16,34 @@
...
-type DropTablePromise = Promise<[d.Operation, d.GrpcOperation]>;
-type ReadPromise = Promise<[Array<Row|Json>]>;
-
-interface CreateTableCallback {
+export interface CreateTableCallback {
   (err: ServiceError, table?: null, operation?: null, apiResponse?: null): void;
-  (err: null, table: Table, operation: d.Operation,
-   apiResponse: d.GrpcOperation): void;
+  (err: null, table: Table, operation: GaxOperation,
+    apiResponse: DeleteResponse): void;
```
callmehiphop: This should be `LongrunningIOperation`. It's a little confusing (to me at least) but any types with the `Response` suffix are typically only applicable to Promises.

src/table.ts
```diff
@@ -16,38 +16,34 @@
...
-
-type CommitPromise = Promise<[s.CommitResponse]>;
-type CreateTablePromise = Promise<[Table, d.Operation, d.GrpcOperation]>;
-type DropTablePromise = Promise<[d.Operation, d.GrpcOperation]>;
-type ReadPromise = Promise<[Array<Row|Json>]>;
-
-interface CreateTableCallback {
+export interface CreateTableCallback {
   (err: ServiceError, table?: null, operation?: null, apiResponse?: null): void;
-  (err: null, table: Table, operation: d.Operation,
```
callmehiphop: This isn't anything you did, but I recently learned that interfaces with overloads (like these) don't infer types properly so I think we actually want to avoid them. Might be worthwhile to create a generic for callbacks in `common.ts`.

src/table.ts
```diff
@@ -434,12 +437,14 @@ class Table {
...
    */
-  insert(rows: object|object[]): CommitPromise;
-  insert(rows: object|object[], callback: s.CommitCallback): void;
-  insert(rows: object|object[], callback?: s.CommitCallback): CommitPromise|
-      void {
+  insert(rows: object | object[], callback?: CommitCallback): Promise<CommitResponse>|void {
     this._mutate('insert', rows, callback!);
   }
+
+  read(request: ReadRequest, options?: TimestampBounds): Promise<[Array<Row | Json>]>;
```
callmehiphop: Can we plop this guy in a type called `ReadResponse` or something?

src/table.ts
```diff
@@ -592,13 +597,9 @@ class Table {
...
    * Reading data using a storing index:
    */
-  read(request: ReadRequest, options?: TimestampBounds): ReadPromise;
-  read(request: ReadRequest, callback: ReadCallback): void;
-  read(request: ReadRequest, options: TimestampBounds, callback: ReadCallback):
-      void;
   read(
       request: ReadRequest, options?: TimestampBounds|ReadCallback,
-      callback?: ReadCallback): ReadPromise|void {
+      callback?: ReadCallback): Promise<[Array<Row|Json>]>|void {
```
callmehiphop: Re: `ReadResponse`

src/transaction-runner.ts
```diff
@@ -81,7 +82,7 @@ interface ErrorCallback {
  *
  * @param {Error} [err] The last known retryable Error.
  */
-export class DeadlineError extends Error implements ServiceError {
+export class DeadlineError extends Error implements RowCountsServiceError {
```
callmehiphop: I don't think we should be extending `RowCountsServiceError` here

src/transaction.ts
```diff
@@ -59,14 +59,15 @@ export interface ExecuteSqlRequest extends Statement, RequestOptions {
...
-
 export interface KeyRange {
   startClosed?: Value[];
   startOpen?: Value[];
   endClosed?: Value[];
   endOpen?: Value[];
 }
-
+export interface ReadCallback {
+  (err: null|ServiceError, rows: Rows): void;
```
callmehiphop: I assume rows can be undefined or null, so we might want to use the `RequestCallback` generic here.

src/transaction.ts
```diff
@@ -29,24 +29,24 @@ import {PartialResultStream, partialResultStream, ResumeToken, Row} from './part
...
+export type BeginResponse = [s.Transaction];
+export type CommitResponse = [ICommitResponse];
+export type ReadResponse = [Rows];
+export type RunResponse = [Rows, s.ResultSetStats];
+export type RunUpdateResponse = [number];
+
+export type BeginTransactionCallback = RequestCallback<GeneratedTransaction>;
+export type CommitCallback = RequestCallback<ICommitResponse>;
+
+export interface CommitRequest {
```
callmehiphop: Sorry if I asked this before, but why do we need this? Isn't there a `ICommitRequest` that covers it?

test/transaction-runner.ts
```diff
@@ -124,7 +124,7 @@ describe('TransactionRunner', () => {
     describe('getNextDelay', () => {
       const FAKE_RETRY_INFO = Buffer.from('fake-retry-info');
 
-      const FAKE_ERROR: ServiceError = new Error('err');
+      const FAKE_ERROR = new Error('err') as ServiceError;
```
callmehiphop: Just curious, why do we prefer this form of casting? Just want to know the reasoning so I can avoid doing it wrong in the future. 😎

package.json
joecervino: Yeah, I get bugs like: `Error: no such type: google.iam.v1.SetIamPolicyRequest` which basically means that the `/iam/` folder can't be found within `protos/`. The `protos/google/` folder, if I'm correctly understanding the error, doesn't have every type or doesn't have them fully mapped out 😅

src/common.ts
joecervino: Not a great reason: wanted to communicate it was Longrunning, for what it's worth 😅

package.json
callmehiphop: Well then!

@jkwlui any reason why synth tool doesn't give us IAM protos?

src/common.ts
callmehiphop: Oh! It's kind of implied, operations are only used when the API can't immediately return a response.

src/table.ts
joecervino: Great to know, will do!

src/transaction-runner.ts
joecervino: I agree that it doesn't seem correct, and maybe it's just my IDE throwing false errors, but `ServiceError` doesn't seem to get extended properly because of its `rowCounts` method. I'll play around with it 👍

src/table.ts
joecervino: Gotcha - I'm going to give all of the responses a refactoring

src/transaction.ts
joecervino: `ICommitRequest` seems to, by default, annotate properties as `undefined` and `null`, so this was a solution that made it a bit stricter where needed. I'll play around with this one too

test/transaction-runner.ts
callmehiphop: Oh, my bad, for some reason I didn't think `code` was a required property, that makes sense then.

package.json
jkwlui: @alexander-fenster

src/transaction-runner.ts
joecervino: @callmehiphop just wanted to show you [this](https://imgur.com/a/Vc8tmf0) real quick

callmehiphop: How come `BatchUpdateError` doesn't just extend `ServiceError`? It extends what appears to be a duplicate interface and this error is probably because it has the same properties with different types.

joecervino: I think I found the problem - something was up with `node_modules/grpc/index.d.ts` so I had to re-`npm i`, you aren't going crazy 😄

callmehiphop: Lol, well lets not jump to any conclusions about that yet

joecervino:
@callmehiphop any idea what to do about this error in System test grpc-js? I can't reproduce the error locally via `npm run system-test`: 
```
Spanner
    1) "before all" hook
(node:193) UnhandledPromiseRejectionWarning: TypeError: Channel's second argument must be a ChannelCredentials
```

callmehiphop:
@joecervino I think you can probably ignore that, we have an open issue about `grpc-js` not working with Spanner (https://github.com/googleapis/nodejs-spanner/issues/555).

joecervino:
@callmehiphop Oh! Okay, cool, I thought 555 could be why - let me know if I need to do anything else to get this one merged 👍

src/common.ts
```diff
@@ -1,11 +1,37 @@
...
-  exactStaleness?: number;
-  readTimestamp?: Date;
-  returnTimestamp?: boolean;
-  strong?: boolean;
+import {google as spTypes} from '../proto/spanner';
+import {Operation as GaxOperation} from 'google-gax/build/src/longrunning';
+import {ServiceError} from 'grpc';
+
+//tslint:disable-next-line no-any
+export type Any = any;
```
callmehiphop: Any chance we could get rid of this? I just noticed something similar in BigQuery and it turned out that an any alias was being used all over the place in places it shouldn't. While it is annoying to add tslint comments, I actually kind of prefer the explicitness of doing so.

src/common.ts
```diff
@@ -1,11 +1,37 @@
...
+export type BasicResponse = [IOperation];
```
callmehiphop: I think we could do a better job name this and `BasicCallback` here. `IOperation` isn't really that basic, and it looks like we already have a similar (more correct?) response below called `LongRunningOperationResponse`.

src/common.ts
```diff
@@ -1,11 +1,37 @@
...
+export interface BasicCallback {
+  (err: ServiceError|null, response?: IOperation|null): void;
+}
+export interface GenericCallback<T> {
+  (err: ServiceError|null, response?: T|null): void;
+}
+export interface ErrorCallback {
+  (error: null|ServiceError|Error): void;
+}
+export interface LongrunningOperationCallback {
```
callmehiphop: Can we normalize this and `LongRunningOperationResponse`? The response version is `LongRunning` instead of `Longrunning`. Truthfully we could drop that part entirely and just do `OperationCallback` if you wanted.

src/table.ts
```diff
@@ -327,11 +319,12 @@ class Table {
    *     const apiResponse = data[0];
    *   });
    */
-  deleteRows(keys: Key[]): CommitPromise;
-  deleteRows(keys: Key[], callback: s.CommitCallback): void;
-  deleteRows(keys: Key[], callback?: s.CommitCallback): CommitPromise|void {
+  deleteRows(keys: Key[], callback?: BasicCallback): Promise<BasicResponse>|void {
```
callmehiphop: Pretty sure this does not return an `IOperation`, unless `BasicResponse` has a typo?

callmehiphop: Follow up to this, I'm noticing `BasicCallback` and `BasicResponse` being used on most crud methods, rather than calling them out can we just make a point to fix them across the board?

src/common.ts
```diff
@@ -14,4 +40,12 @@ export interface CreateSessionOptions {
 }
 export type GetTimestamp = {
   nanos: number; seconds: number;
-};
+}
+export interface TransactionOptions {
```
callmehiphop: Pretty sure this object was superseded by `TimestampBounds` and we shouldn't need it, or at the very least we should remove the `timeout` option.

src/table.ts
```diff
@@ -592,25 +588,21 @@ class Table {
...
   read(
-      request: ReadRequest, options?: TimestampBounds|ReadCallback,
-      callback?: ReadCallback): ReadPromise|void {
-    const rows: Array<Row|Json> = [];
-
-    if (is.fn(options)) {
-      callback = options as ReadCallback;
-      options = {} as TimestampBounds;
-    }
+      request: ReadRequest, optionsOrCallback?: TransactionOptions|ReadCallback,
```
callmehiphop: Pretty sure we want to use `TimestampBounds`, but if that is not the case then we need to update the casts below to use `TransactionOptions` instead.

src/table.ts
```diff
@@ -764,15 +757,15 @@ class Table {
...
+    this.database.runTransaction((err: ServiceError|null, transaction?: Transaction|null) => {
       if (err) {
         callback(err);
         return;
       }
 
-      transaction[method](this.name, rows);
-      transaction.commit(callback);
+      transaction![method](this.name, rows);
+      transaction!.commit(callback as CommitCallback);
```
callmehiphop: Why are we casting this? Why not just have it passed in as `CommitCallback`?

src/transaction.ts
```diff
@@ -29,44 +29,45 @@ import {PartialResultStream, partialResultStream, ResumeToken, Row} from './part
...
-  strong?: boolean;
-  minReadTimestamp?: PreciseDate|p.ITimestamp;
-  maxStaleness?: number|p.IDuration;
-  readTimestamp?: PreciseDate|p.ITimestamp;
-  exactStaleness?: number|p.IDuration;
-  returnReadTimestamp?: boolean;
-}
+export type BatchUpdatePromise = Promise<[number[], s.ExecuteBatchDmlResponse]>;
+export type BeginPromise = Promise<[s.Transaction]>;
+export type CommitPromise = Promise<[s.CommitResponse]>;
```
callmehiphop: I see some mixing between the interfaces declare in the common file and the gapic types. Can we use one or the other?

src/transaction.ts
```diff
@@ -221,8 +221,8 @@ export class Snapshot extends EventEmitter {
     this._options = {readOnly};
   }
 
-  begin(): BeginPromise;
-  begin(callback: s.BeginTransactionCallback): void;
+  begin(): Promise<BeginResponse>;
```
callmehiphop: Can we mirror the callback version and call this `BeginTransactionResponse`?

joecervino: You got it 👍

src/common.ts
joecervino: Yeah, I've been starting to prefer the same. Will do 👌

src/common.ts
joecervino: I'll clean it all up asap

src/table.ts
joecervino: I'll fix them across the board 👍

PR #549, opened 2019-03-06
chore(deps): fix broken dep types

callmehiphop:
A new patch release of sinon types came out yesterday and it looks like maybe the Stub type is borked. Pinning to previous patch release as a temporary solution.


callmehiphop:
Looks like `p-limit` types have also been changed. In addition to that I noticed it was listed as a dependency, but looks like it was meant to be a dev dep?

PR #554, opened 2019-03-07
test: add install tests and fix issues

JustinBeckwith:


src/index.ts
```diff
@@ -44,6 +42,10 @@ const gapic = Object.freeze({
 
 const gcpApiConfig = require('./spanner_grpc_config.json');
 
+export interface SpannerOptions extends GoogleAuthOptions {
```
callmehiphop: I think maybe we want `gax.GrpcClientOptions` instead of `GoogleAuthOptions`. Also let's add `port` and `sslCreds`.

PR #571, opened 2019-03-25
docs(samples): Add Cloud Spanner Batch DML sample

jsimonweb:


samples/dml.js
```diff
@@ -529,6 +529,61 @@ async function deleteUsingPartitionedDml(instanceId, databaseId, projectId) {
...
+
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const projectId = 'my-project-id';
+  // const instanceId = 'my-instance';
+  // const databaseId = 'my-database';
+
+  // Creates a client
+  const spanner = new Spanner({
```
JustinBeckwith: do we need to explicitly define the projectId here?

samples/dml.js
```diff
@@ -529,6 +529,61 @@ async function deleteUsingPartitionedDml(instanceId, databaseId, projectId) {
...
+  };
+
+  const update = {
+    sql: `UPDATE Albums SET MarketingBudget = MarketingBudget * 2
+      WHERE SingerId = 1 and AlbumId = 3`,
+  };
+
+  const dmlStatements = [insert, update];
+
+  database.runTransaction(async (err, transaction) => {
```
JustinBeckwith: can we use `async/await` here?

samples/dml.js
```diff
@@ -529,6 +529,61 @@ async function deleteUsingPartitionedDml(instanceId, databaseId, projectId) {
...
+    }
+    try {
+      const rowCounts = await transaction.batchUpdate(dmlStatements);
+
+      console.log(
+        `Successfully executed ${rowCounts.length} SQL statements using Batch DML.`
+      );
+      await transaction.commit();
+    } catch (err) {
+      console.error('ERROR:', err);
```
JustinBeckwith: Can we re-throw here?

samples/dml.js
callmehiphop: Here's an example of that - https://github.com/googleapis/nodejs-spanner/blob/master/system-test/spanner.ts#L3836

samples/dml.js
```diff
@@ -529,6 +529,61 @@ async function deleteUsingPartitionedDml(instanceId, databaseId, projectId) {
...
+
+  const dmlStatements = [insert, update];
+
+  database.runTransaction(async (err, transaction) => {
+    if (err) {
+      console.error(err);
+      return;
+    }
+    try {
+      const rowCounts = await transaction.batchUpdate(dmlStatements);
```
callmehiphop: `batchUpdate` returns an array so this should be 

```js
const [rowCounts] = await transaction.batchUpdate(dmlStatements);
```

samples/dml.js
```diff
@@ -529,6 +529,57 @@ async function deleteUsingPartitionedDml(instanceId, databaseId, projectId) {
...
+  };
+
+  const dmlStatements = [insert, update];
+
+  try {
+    await database.runTransactionAsync(async transaction => {
+      const [rowCounts] = await transaction.batchUpdate(dmlStatements);
+      await transaction.commit();
+      console.log(
+        `Successfully executed ${rowCounts.length} SQL statements using Batch DML.`
```
callmehiphop: Oh whoops, I missed this, but `rowCounts` should be a number

callmehiphop: Made a mistake, ignore me!

samples/dml.js
jsimonweb: This matches working client instantiation already existing in sample. I'd prefer to address this in a separate PR if possible.

PR #618, opened 2019-05-21
docs(samples): fix code snippets that transfer money between albums

hegemonic:
The `writeWithTransactionUsingDml` function moved $200,000 from Album1 to Album2. However, Album1 is initialized with a budget of $100,000. As a result, you can't work through the [getting started doc](https://cloud.google.com/spanner/docs/getting-started/nodejs/) from start to finish. This function also defined and checked a `minimumAmountToTransfer`, which isn't consistent with the equivalent snippets for other languages.

In addition, the `readWriteTransaction` snippet moved $200,000, but in the opposite direction, which was somewhat confusing.

Finally, neither snippet confirmed that the source album's budget was at least as large as the transfer amount.

I made both snippets transfer $100,000 from Album1 to Album2. I also fixed the logic errors, updated the tests as needed, and updated a few comments for consistency.

- [x] Tests and linter pass


hegemonic:
@jsimonweb Any idea why the CI builds aren't running?

hegemonic:
I filed #621 for the docs build issue, which is unrelated to this PR.

bcoe:
```bash
/tmpfs/src/github/nodejs-spanner/node_modules/intl-messageformat/lib/index.js:1
(function (exports, require, module, __filename, __dirname) { module.paths = ["/tmpfs/src/github/nodejs-spanner/node_modules/jsdoc/lib"].concat(module.paths).concat(["/tmpfs/src/github/nodejs-spanner/node_modules/jsdoc/node_modules"]); import IntlMessageFormat from './core';
                                                                                                                                                                                                                                                   ^^^^^^^^^^^^^^^^^

SyntaxError: Unexpected identifier
    at new Script (vm.js:80:7)
    at createScript (vm.js:274:10)
    at Object.runInThisContext (vm.js:326:10)
```

I can confirm that this is the same bug I was bumping into; [fix here](https://github.com/hegemonic/jsdoc-baseline/pull/195).

hegemonic:
I'm making some changes to this PR at the request of @jsimonweb. Please hold off on reviewing/merging for now.

hegemonic:
Okay, this is now ready to review! @jsimonweb: PTAL.

hegemonic:
Looks like someone with the right privileges might need to add the `kokoro:force-run` label again...

package.json
```diff
@@ -92,6 +92,7 @@
     "@types/through2": "^2.0.34",
     "@types/uuid": "^3.4.4",
     "binary-search-bounds": "^2.0.4",
+    "chai": "^4.2.0",
```
JustinBeckwith: This probably belongs in `samples/package.json`

hegemonic: Ah, I didn't notice that the `samples/` directory was also its own package. I took this out. It's already in `samples/package.json`, so no need to change that file.

hegemonic:
The system test failure appears to be unrelated to this PR.

PR #619, opened 2019-05-22
refactor(types): enable noImplicitAny in instance.ts

### LINKED ISSUE #223: Convert to TypeScript

laljikanjareeya:
towards #223 

- [x] Tests and linter pass


src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+import {google as dbInstanceAdminClient} from '../proto/spanner_instance_admin';
+import {google as dbDatabaseAdminClient} from '../proto/spanner_database_admin';
+import {Operation as GaxOperation} from 'google-gax/build/src/longRunningCalls/longrunning';
+import {SessionPoolOptions, SessionPoolInterface} from './session-pool';
+import {SessionPool, Spanner} from '.';
+import {ServiceError} from 'grpc';
+
+export type IDatabase = dbDatabaseAdminClient.spanner.admin.database.v1.IDatabase;
+export type IInstance = dbInstanceAdminClient.spanner.admin.instance.v1.IInstance;
+export type LongrunningIOperation = dbInstanceAdminClient.longrunning.IOperation;
```
callmehiphop: Can this either be `IOperation` or `ILongrunningOperation`? We should be consistent with how we prefix interfaces.

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+];
+export type CreateDatabaseResponse = [
+  Database,
+  GaxOperation,
+  LongrunningIOperation
+];
+export type DeleteInstanceResponse = [dbInstanceAdminClient.protobuf.IEmpty];
+export type ExistsInstanceResponse = [boolean];
+export type GetInstanceResponse = [Instance, IInstance];
+export type GetDatabasesResponse = [Database[], LongrunningIOperation];
```
callmehiphop: This looks wrong, I think it should be `IListDatabasesResponse`

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+export type LongrunningOperationResponse = [
+  GaxOperation,
+  LongrunningIOperation
+];
+
+export interface CreateDatabaseOptions {
+  [k: string]: string | string[] | SessionPoolOptions | SessionPool | undefined;
```
callmehiphop: Why do we need this?

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+export type DeleteInstanceResponse = [dbInstanceAdminClient.protobuf.IEmpty];
+export type ExistsInstanceResponse = [boolean];
+export type GetInstanceResponse = [Instance, IInstance];
+export type GetDatabasesResponse = [Database[], LongrunningIOperation];
+export type LongrunningOperationResponse = [
+  GaxOperation,
+  LongrunningIOperation
+];
+
+export interface CreateDatabaseOptions {
```
callmehiphop: I think we want to extend `ICreateDatabaseRequest` here

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+  [k: string]: string | string[] | SessionPoolOptions | SessionPool | undefined;
+  poolOptions?: SessionPoolOptions;
+  poolCtor?: SessionPool;
+  schema?: string;
+  extraStatements?: string[];
+}
+export interface CreateInstanceRequest {
```
callmehiphop: I think we want to extend `IInstance` here and only include the convenience options.

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+  config: string;
+  nodes: number;
+}
+export interface GetDatabasesRequest {
```
callmehiphop: I think we'll probably want to extend `IListDatabasesRequest` here

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+  autoPaginate?: boolean;
+  maxApiCalls?: number;
+  maxResults?: number;
+  pageSize?: number;
+  pageToken?: string;
+}
+export interface QueryObject {
```
callmehiphop: Why do we need this?

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+  [k: string]: string;
+}
+
+export interface CreateInstanceCallback {
+  (
+    err: Error | null,
```
callmehiphop: Definitely going to want to use `ServiceError` from gRPC here. I'm noticing some inconsistencies between similar interfaces, where somethings are listed as optional and others are not. We may want to consider using a generic for callbacks/promises/etc. You can find some [examples](https://github.com/googleapis/nodejs-pubsub/blob/master/src/pubsub.ts#L132) in the PubSub repo.

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+  (err: Error | null, apiResponse: dbInstanceAdminClient.protobuf.IEmpty): void;
+}
+export interface ExistsInstanceCallback {
+  (err: Error | null, exists: boolean | null): void;
+}
+export interface GetDatabasesCallback {
+  (
+    err: Error | null,
+    databases?: Database[],
+    apiResponse?: LongrunningIOperation
```
callmehiphop: Like the `GetDatabasesResponse` interface, I believe this is incorrect.

src/instance.ts
```diff
@@ -16,13 +16,110 @@
...
+  ): void;
+}
+export interface GetInstanceCallback {
+  (err: Error | null, instance?: Instance, apiResponse?: IInstance): void;
+}
+export interface GetInstanceMetadataCallback {
+  (
+    err: ServiceError | null,
+    metadata?: IInstance,
+    apiResponse?: LongrunningIOperation
```
callmehiphop: I don't think this type is correct here

src/instance.ts
```diff
@@ -319,15 +447,18 @@ class Instance extends common.ServiceObject {
             method: 'deleteInstance',
             reqOpts,
           },
-          (err, resp) => {
+          (err: Error, resp: LongrunningIOperation) => {
```
callmehiphop: The `apiResponse` type has `IEmpty` listed within the `DeleteInstanceCallback` interface, this appears to conflict with that.

src/instance.ts
```diff
@@ -415,24 +552,39 @@ class Instance extends common.ServiceObject {
...
       if (err) {
         if (err.code === 5 && options.autoCreate) {
-          this.create(options, (err, database, operation) => {
-            if (err) {
-              callback(err);
-              return;
+          this.create(
+            options,
+            (
+              err: ServiceError | null,
```
callmehiphop: I think we should be able to infer these types?

src/instance.ts
```diff
@@ -526,18 +693,20 @@ class Instance extends common.ServiceObject {
...
         if (databases) {
           arguments[1] = databases.map(database => {
-            const databaseInstance = self.database(database.name);
+            const databaseInstance = self.database(database.name!);
             databaseInstance.metadata = database;
             return databaseInstance;
           });
         }
-        callback.apply(null, arguments);
+        // tslint:disable-next-line no-any
```
callmehiphop: Shouldn't need this, please refer to the PubSub [example](https://github.com/googleapis/nodejs-pubsub/blob/master/src/pubsub.ts#L631) on how we've handled paginated methods like this.

src/instance.ts
```diff
@@ -576,7 +745,7 @@ class Instance extends common.ServiceObject {
    *   const apiResponse = data[1];
    * });
    */
-  getMetadata(callback) {
+  getMetadata(callback: GetInstanceMetadataCallback): Metadata {
```
callmehiphop: This looks wrong

- I think we want to use `IInstance` over `Metadata` (AFAIK `Metadata` is an alias to `any`). 
- Current signature implies that this is a synchronous method, but it is not
- Missing want a Promise overload

src/instance.ts
```diff
@@ -589,6 +758,9 @@ class Instance extends common.ServiceObject {
       callback
     );
   }
+
+  setMetadata(metadata: Metadata): Promise<LongrunningOperationResponse>;
```
callmehiphop: I think we want to use `IInstance` over `Metadata` here.

system-test/spanner.ts
```diff
@@ -51,7 +51,8 @@ describe('Spanner', () => {
   after(deleteTestInstances);
 
   describe('types', () => {
-    const database = instance.database(generateName('database'));
+    // tslint:disable-next-line: no-any
+    const database = instance.database(generateName('database')) as any;
```
callmehiphop: Why do we need this?

system-test/spanner.ts
```diff
@@ -3889,7 +3894,7 @@ describe('Spanner', () => {
           let err;
 
           try {
-            await txn.batchUpdate(null);
+            await txn.batchUpdate([]);
```
callmehiphop: Why are we changing this test? If it is a type issue then you should cast it to `unknown` or `{}`.

test/instance.ts
```diff
@@ -450,9 +462,7 @@ describe('Instance', () => {
     });
 
     it('should return true if error is absent', done => {
-      instance.getMetadata = callback => {
-        callback(null);
-      };
+      sandbox.stub(instance, 'getMetadata').yields(null);
```
callmehiphop: Should this be `resolves()`?

system-test/spanner.ts
laljikanjareeya: Resolved!!

src/instance.ts
laljikanjareeya: Done!!

src/instance.ts
laljikanjareeya: Done!!

src/instance.ts
laljikanjareeya: Done!!

src/instance.ts
laljikanjareeya: Done!

src/instance.ts
laljikanjareeya: Removed

src/instance.ts
laljikanjareeya: used `GetInstanceCallback` instead

test/instance.ts
laljikanjareeya: here `resolves()` fails this test case..

src/instance.ts
laljikanjareeya: Removed!!

src/instance.ts
laljikanjareeya: here no definition for `this.create`, so need to expose datatype..

test/instance.ts
```diff
@@ -438,9 +455,9 @@ describe('Instance', () => {
     it('should return any non-404 like errors', done => {
       const error = {code: 3};
 
-      instance.getMetadata = callback => {
-        callback(error);
-      };
+      sandbox
```
AVaksman: It might be a good idea to add
```
afterEach(() => sandbox.restore());
```
In this `describe` block

laljikanjareeya: Done!!

laljikanjareeya:
@callmehiphop resolved all the given changes can you review it again please.

src/instance.ts
```diff
@@ -16,13 +16,101 @@
...
+  SessionPool,
+} from './session-pool';
+import {Spanner} from '.';
+import {ServiceError} from 'grpc';
+import {RequestCallback, PagedResponse} from './common';
+import {Stream} from 'stream';
+
+export type IDatabase = databaseAdmin.spanner.admin.database.v1.IDatabase;
+export type IInstance = instanceAdmin.spanner.admin.instance.v1.IInstance;
+export type ILongrunningOperation = instanceAdmin.longrunning.IOperation;
```
callmehiphop: Any reason we don't just call this `IOperation`?

src/instance.ts
```diff
@@ -16,13 +16,101 @@
...
+export interface CreateInstanceRequest extends IInstance {
+  nodes: number;
+}
+export interface GetDatabasesRequest
+  extends databaseAdmin.spanner.admin.database.v1.IListDatabasesRequest {
+  autoPaginate?: boolean;
+  maxApiCalls?: number;
+  maxResults?: number;
+}
+export interface CreateInstanceCallback {
```
callmehiphop: You might consider creating a generic for calls like this, `LongRunningCallback<Resource>`?

src/instance.ts
```diff
@@ -16,13 +16,101 @@
...
+  (
+    err: ServiceError | null,
+    instance: Instance | null,
```
callmehiphop: I think these should all be optional.

src/instance.ts
```diff
@@ -16,13 +16,101 @@
...
+  databaseAdmin.spanner.admin.database.v1.IListDatabasesResponse
+>;
+export interface GetInstanceCallback {
+  (
+    err: ServiceError | null,
+    instance?: Instance,
+    apiResponse?: IInstance
+  ): void;
+}
+export interface GetMetadataCallback {
```
callmehiphop: These names are probably too generic, especially if other classes have similar methods. I think we want to call these `GetInstanceMetadataCalback`, etc.

src/instance.ts
```diff
@@ -41,7 +129,13 @@ import {Database} from './database';
  * const instance = spanner.instance('my-instance');
  */
 class Instance extends common.ServiceObject {
-  constructor(spanner, name) {
+  formattedName_: string;
+  // tslint:disable-next-line: no-any
+  request: (config: any, callback: Function) => void;
+  requestStream: (config?: {}) => Stream;
```
callmehiphop: `requestStream` returns a `Duplex` stream, is the `Stream` type compatible?

src/instance.ts
```diff
@@ -41,7 +129,13 @@ import {Database} from './database';
  * const instance = spanner.instance('my-instance');
  */
 class Instance extends common.ServiceObject {
-  constructor(spanner, name) {
+  formattedName_: string;
+  // tslint:disable-next-line: no-any
+  request: (config: any, callback: Function) => void;
```
callmehiphop: Maybe we should add a types for `config` and `callback` inside of the index file.

src/instance.ts
```diff
@@ -108,6 +206,18 @@ class Instance extends common.ServiceObject {
     this.requestStream = spanner.requestStream.bind(spanner);
     this.databases_ = new Map();
   }
+
+  createDatabase(name: string): Promise<CreateDatabaseResponse>;
```
callmehiphop: Lets delete this overload and mark `options` as optional on the other Promise returning one.

src/instance.ts
```diff
@@ -415,32 +549,55 @@ class Instance extends common.ServiceObject {
...
+              instance?: Instance,
+              operation?: GaxOperation | null
+            ) => {
+              if (err) {
+                callback(err);
+                return;
+              }
+              operation!
+                .on('error', callback)
+                .on('complete', (metadata: Instance) => {
```
callmehiphop: This should be `IInstance`.

src/instance.ts
```diff
@@ -415,32 +549,55 @@ class Instance extends common.ServiceObject {
...
+                  this.metadata = metadata;
+                  callback(null, this, metadata as IInstance);
```
callmehiphop: Shouldn't need to cast here.

src/instance.ts
```diff
@@ -415,32 +549,55 @@ class Instance extends common.ServiceObject {
...
-            });
-          });
+          );
           return;
         }
         callback(err);
         return;
       }
-      callback(null, this, metadata);
+      callback(null, this, metadata as IInstance);
```
callmehiphop: Shouldn't need to cast here

src/instance.ts
```diff
@@ -415,32 +549,55 @@ class Instance extends common.ServiceObject {
...
     });
   }
+
+  getDatabases(): Promise<GetDatabasesResponse>;
+  getDatabases(query: GetDatabasesRequest): Promise<GetDatabasesResponse>;
```
callmehiphop: Make `query` optional and then we can delete the first Promise overload.

src/transaction.ts
```diff
@@ -1159,7 +1159,7 @@ export class Transaction extends Dml {
     this._options = {readWrite: options};
   }
 
-  batchUpdate(queries: Array<string | Statement>): BatchUpdatePromise;
+  batchUpdate(queries: Array<string | Statement> | null): BatchUpdatePromise;
```
callmehiphop: Pretty sure this should never be null?

src/instance.ts
laljikanjareeya: change `ILongrunningOperation` to `IOperation`.

src/instance.ts
laljikanjareeya: Done!

src/transaction.ts
laljikanjareeya: removed!!

src/instance.ts
laljikanjareeya: Change `Stream`  to `Duplex` not able to find any document for same.

src/common.ts
```diff
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2017 Google Inc. All Rights Reserved.
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
callmehiphop: Should we be changing the year here?

src/common.ts
```diff
@@ -13,20 +13,37 @@
...
  */
-import {google as databaseAdmin} from '../proto/spanner_database_admin';
+
 import {ServiceError, CallOptions} from 'grpc';
+import {Operation as GaxOperation} from 'google-gax';
+import {google as instanceAdmin} from '../proto/spanner_instance_admin';
+import {google as databaseAdmin} from '../proto/spanner_database_admin';
+
+export type IOperation = instanceAdmin.longrunning.IOperation;
+export interface TransactionOptions {
```
callmehiphop: Do we need this? I think `readOnly` should be gone and we have [`TimestampBounds`](https://github.com/googleapis/nodejs-spanner/blob/master/src/transaction.ts#L42)

src/instance.ts
```diff
@@ -388,19 +463,25 @@ class Instance extends common.ServiceObject {
...
         return;
       }
 
       const exists = !err || err.code !== NOT_FOUND;
-      callback(null, exists);
+      callback!(null, exists);
     });
   }
+
+  get(options?: GetConfig): Promise<GetInstanceResponse>;
```
callmehiphop: I might advocate that we should extend `GetConfig` and create an interface called `GetInstanceConfig`. In the event that Spanner decides to add options later, this would make it a little more future proof and prevent us from having to make a breaking change. WDYT?

system-test/spanner.ts
```diff
@@ -1253,7 +1262,8 @@ describe('Spanner', () => {
         err => {
           assert.ifError(err);
 
-          database.run('SELECT * FROM Singers', (err, rows) => {
+          // tslint:disable-next-line: no-any
+          database.run('SELECT * FROM Singers', (err, rows: any) => {
```
callmehiphop: There's actually a [`Row` interface](https://github.com/googleapis/nodejs-spanner/blob/master/src/partial-result-stream.ts#L62) you can use here.

system-test/spanner.ts
```diff
@@ -1411,7 +1421,11 @@ describe('Spanner', () => {
           )
           .then(data => {
             const rows = data[0];
-            assert.deepStrictEqual(rows.shift().toJSON(), EXPECTED_ROW);
+            assert.deepStrictEqual(
+              // tslint:disable-next-line: no-any
+              (rows as any).shift().toJSON(),
```
callmehiphop: And here!

system-test/spanner.ts
```diff
@@ -1558,17 +1572,17 @@ describe('Spanner', () => {
                   child: 'bool',
                 },
               },
-            };
+            } as {}) as ExecuteSqlRequest;
```
callmehiphop: Any reason we have to cast it like this? I would think `const query: ExecuteSqlRequest = {};` would work

callmehiphop: Follow up - assuming that does work, we should change all the fixtures that use this form of casting.

system-test/spanner.ts
```diff
@@ -3217,9 +3231,9 @@ describe('Spanner', () => {
             assert.ifError(err);
 
             // We just want the two most recent ones.
-            rows.splice(0, rows.length - 2);
+            rows!.splice(0, rows!.length - 2);
 
-            rows = rows.map(x => x.toJSON());
+            rows = rows!.map(x => x.toJSON()) as Row[];
```
callmehiphop: I think this should be [`as Json[]`](https://github.com/googleapis/nodejs-spanner/blob/master/src/codec.ts#L32)

test/instance.ts
```diff
@@ -347,15 +353,24 @@ describe('Instance', () => {
       const database = instance.database(NAME, poolOptions);
 
       assert(database instanceof FakeDatabase);
-      assert.strictEqual(database.calledWith_[0], instance);
-      assert.strictEqual(database.calledWith_[1], NAME);
-      assert.strictEqual(database.calledWith_[2], poolOptions);
+      assert.strictEqual(
```
callmehiphop: might be easier if we just cast it once when we declare the variable?

system-test/spanner.ts
laljikanjareeya: I need to change type of this as `any` b/c not able to find the property like `SingerId` and `Name` is used below.

system-test/spanner.ts
laljikanjareeya: same thing here row is single object and we are using `shift()` will give compilation error.

system-test/spanner.ts
laljikanjareeya: `rows` is type of `Row[]` while using `Json[]` it gives below error:
`Type 'Json[]' is not assignable to type 'Row[]'.`

system-test/spanner.ts
laljikanjareeya: tried with that already but gives below Error:
`Type '{ v: { type: string; child: string; }; }' is not assignable to type '{ [param: string]: string; }'.
 Property 'v' is incompatible with index signature.
 Type '{ type: string; child: string; }' is not assignable to type 'string'`

also tried with `as ExecuteSqlRequest` only but that is also giving error.

system-test/spanner.ts
callmehiphop: I see. I'm guessing it is because we're attempting to re-use the variable `rows` that is set to type `Row[]`. We may just want to assign it to a new variable altogether of type `Json[]`.

system-test/spanner.ts
callmehiphop: I think this is because we're casting this incorrectly. Further down we map the rows into JSON objects but cast it as `Row[]` which a `Row` is an array, not an object. I believe creating a new variable with the correct type should circumvent the need to cast to any.

system-test/spanner.ts
callmehiphop: I see, it looks like the [typing](https://github.com/googleapis/nodejs-spanner/blob/master/src/transaction.ts#L60) is incorrect. We should probably fix it to allow for it to be a [`Type` object](https://github.com/googleapis/nodejs-spanner/blob/master/src/codec.ts#L397)

src/common.ts
laljikanjareeya: my bad, Removed!!

src/instance.ts
laljikanjareeya: sounds good...

src/instance.ts
```diff
@@ -501,7 +502,7 @@ class Instance extends common.ServiceObject {
    * normally required for the `create` method must be contained within this
    * object as well.
    *
-   * @param {options} [options] Configuration object.
+   * @param {GetInstanceConfig} [options] Configuration object.
```
callmehiphop: I think this needs to stay `options`

system-test/spanner.ts
callmehiphop: Why do we need to cast this to `any`?

system-test/spanner.ts
laljikanjareeya: used 
```ts
const rows = data[0] as {} as Row[];
```

PR #627, opened 2019-05-28
refactor(types): enable noImplicitAny in database.ts

laljikanjareeya:
towards #223 

- [ ] Tests and linter pass


src/database.ts
```diff
@@ -55,24 +60,102 @@ import {
...
+import {google as spanner_client} from '../proto/spanner';
+import {Any, Schema} from './common';
+import {ServiceError, CallOptions} from 'grpc';
+import {Readable, Transform} from 'stream';
+import {Stats} from 'fs';
+import {StreamProxy} from 'google-gax/build/src/streamingCalls/streaming';
+
+interface CreateBatchTransactionCallback {
+  (
+    err?: Error | null,
```
callmehiphop: We shouldn't be using `Error`, gRPC Errors also have `status`, `metadata` and `details` fields, please use the appropriate gRPC error type.

src/database.ts
```diff
@@ -55,24 +60,102 @@ import {
...
 
-export interface GetTransactionCallback {
-  (err: Error, transaction?: null): void;
-  (err: null, transaction: Transaction): void;
+interface GetTransactionCallback {
+  (err?: Error | null, transaction?: Transaction | null): void;
+}
+
+interface SessionPoolCtor {
+  (arg0: Database, arg1: SessionPoolOptions | null): void;
```
callmehiphop: Let's make these param names more descriptive.

src/database.ts
```diff
@@ -55,24 +60,102 @@ import {
...
+    [key: string]: Any;
+    session?: string;
+  };
+  client?: string;
+  method?: string;
+  gaxOptions: CallOptions;
+}
+
+interface MakePooledRequestCallback {
+  (err: Error | null, args: IArguments | null): void; //IArguments to represent arguments object used
```
callmehiphop: This is incorrect, `apply()` works the same way as the spread argument, so we won't be passing the `arguments` object back as a single parameter, but rather each argument will be its own parameter.

src/codec.ts
```diff
@@ -14,12 +14,11 @@
...
  */
 import {Service} from '@google-cloud/common-grpc';
-import {DateStruct, PreciseDate} from '@google-cloud/precise-date';
+import {PreciseDate} from '@google-cloud/precise-date';
 import arrify = require('arrify');
-import {CallOptions} from 'google-gax';
 import * as is from 'is';
 import {common as p} from 'protobufjs';
-
+import {google as spanner_client} from '../proto/spanner';
```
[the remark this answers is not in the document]
laljikanjareeya: Done!!

src/database.ts
```diff
@@ -328,15 +442,28 @@ class Database extends ServiceObject {
    *   const apiResponse = data[1];
    * });
    */
-  createSession(options, callback?) {
-    if (is.function(options)) {
-      callback = options;
-      options = {};
-    }
+  createSession(
```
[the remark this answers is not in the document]
laljikanjareeya: Done!!!

src/database.ts
```diff
@@ -599,11 +733,11 @@ class Database extends ServiceObject {
...
    * });
    */
-  get(options?: GetDatabaseOptions): Promise<DatabaseResponse>;
-  get(options: GetDatabaseOptions, callback: DatabaseCallback): void;
+  get(options?: GetConfig): Promise<DatabaseResponse>;
   get(callback: DatabaseCallback): void;
+  get(options: GetConfig, callback: DatabaseCallback): void;
   get(
-    optionsOrCallback?: GetDatabaseOptions | DatabaseCallback,
+    optionsOrCallback?: GetConfig | DatabaseCallback,
```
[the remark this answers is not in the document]
laljikanjareeya: added new type

src/database.ts
```diff
@@ -613,22 +747,27 @@ class Database extends ServiceObject {
...
+            (err, database: Database, operation: GaxOperation) => {
+              if (err) {
+                callback!(err);
+                return;
+              }
+              operation
+                .on('error', callback!)
+                .on('complete', (metadata: Metadata) => {
+                  this.metadata = metadata;
+                  callback!(null, this, metadata as r.Response);
```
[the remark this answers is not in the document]
laljikanjareeya: in this case base class`(ServiceObject)` have same name method and it has `r.Response` as `apiResponse`, you want me to use `ServiceObject` as delegate here?

src/database.ts
```diff
@@ -613,22 +747,27 @@ class Database extends ServiceObject {
...
-            });
-          });
+          );
           return;
         }
         callback!(err);
         return;
       }
-      callback!(null, this, metadata);
+      callback!(null, this, metadata as r.Response);
```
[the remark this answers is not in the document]
laljikanjareeya: same thing as above

src/database.ts
```diff
@@ -731,8 +872,12 @@ class Database extends ServiceObject {
    *   const apiResponse = data[1];
    * });
    */
-  getSchema(callback) {
-    const reqOpts = {
+  getSchema(): Promise<[string[], r.Response]>;
+  getSchema(callback: GetSchemaCallback): void;
+  getSchema(
+    callback?: GetSchemaCallback
+  ): void | Promise<[string[], r.Response]> {
```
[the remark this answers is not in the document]
laljikanjareeya: Done!

src/database.ts
```diff
@@ -1013,21 +1171,30 @@ class Database extends ServiceObject {
    * @param {object} config Request config
    * @param {function} callback Callback function
    */
-  makePooledRequest_(config, callback) {
+  makePooledRequest_(config: MakePooledConfig): Promise<Session>;
```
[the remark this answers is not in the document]
laljikanjareeya: Done!

src/database.ts
```diff
@@ -142,21 +229,22 @@ class Database extends ServiceObject {
...
+        options: CreateDatabaseOptions,
+        callback: CreateDatabaseCallback
+      ) => {
         return instance.createDatabase(formattedName_, options, callback);
       },
     } as {}) as ServiceObjectConfig);
 
+    this.pool_ =
+      typeof poolOptions === 'function'
+        ? new (poolOptions as Any)(this, null)
```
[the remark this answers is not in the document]
laljikanjareeya: it gives below error:
`Cannot use 'new' with an expression whose type lacks a call or construct signature.ts(2351)`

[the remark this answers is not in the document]
callmehiphop: Ah, right, we probably need to either add a signature for `new` or else create a separate type for the constructor.

src/database.ts
```diff
@@ -55,24 +65,110 @@ import {
...
+type GetMetadataResponse = [databaseAdmin.spanner.admin.database.v1.IDatabase];
+interface GetMetadataCallback {
+  (
+    err: Error | null,
+    metadata?: databaseAdmin.spanner.admin.database.v1.IDatabase
+  ): void;
+}
+
+interface GetSchemaCallback {
+  (err: Error | null, ...args: Array<{} | string>): void;
```
callmehiphop: This doesn't really provide any type safety, I think we should mirror what we did in [PubSub](https://github.com/googleapis/nodejs-pubsub/blob/master/src/pubsub.ts#L148)

src/database.ts
```diff
@@ -55,24 +65,110 @@ import {
...
+    err: Error | null,
+    sessions?: Session[],
+    nextQuery?: spannerClient.spanner.v1.IListSessionsRequest,
+    apiResponse?: spannerClient.spanner.v1.IListSessionsResponse
+  ): void;
+}
+
+type GetSessionsResponse = [
+  Session[],
+  spannerClient.spanner.v1.IListSessionsRequest,
```
callmehiphop: I think we need a union type for list response types. By default an `autoPaginate` option will make the response look like this

```ts
[Session[]]
```

But if `autoPaginate` is set to false, it'll look more like this

```ts
[
  Session[],
  IListSessionsRequest | null,
  IListSessionsResponse
]
```

In addition to that, the `nextQuery` item (in this case `IListSessionsRequest`) can sometimes contain convenience properties. In other clients we've used a [generic](https://github.com/googleapis/nodejs-bigquery/blob/master/src/index.ts#L70) here as well

src/database.ts
```diff
@@ -182,13 +284,19 @@ class Database extends ServiceObject {
...
-    if (is.string(session)) {
-      session = this.session(session);
-    }
-    const transaction = new BatchTransaction(session, options);
+    const transaction: BatchTransaction = new BatchTransaction(
+      session,
+      options
+    );
     transaction.id = id;
     transaction.readTimestamp = identifier.readTimestamp;
```
callmehiphop: Looks like you uncovered a bug, I think it should be `identifier.timestamp` maybe?

src/database.ts
```diff
@@ -228,12 +336,20 @@ class Database extends ServiceObject {
    *   }
    * });
    */
-  close(callback) {
+  close(callback: SessionPoolCloseCallback): void {
```
callmehiphop: Looks like we're missing overloads for Promises here

src/database.ts
```diff
@@ -460,6 +602,8 @@ class Database extends ServiceObject {
       }
     });
   }
+  delete(): Promise<BasicResponse>;
```
callmehiphop: I believe there is a type called `IEmpty` that we should be using here instead

src/database.ts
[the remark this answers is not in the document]
callmehiphop: Well that's not good, sounds like a TypeScript bug in the `common-grpc` lib. 

@stephenplusplus @JustinBeckwith @bcoe just an FYI

src/database.ts
```diff
@@ -1011,21 +1187,25 @@ class Database extends ServiceObject {
...
     pool.getReadSession((err, session) => {
       if (err) {
-        callback(err, null);
+        callback!(err, null);
         return;
       }
       config.reqOpts.session = session!.formattedName_;
       // tslint:disable-next-line only-arrow-functions
-      this.request(config, function() {
+      this.request(config, function(err: ServiceError, ...args: Array<{}>) {
```
callmehiphop: Could probably use an arrow function here and remove the tslint comment.

src/database.ts
```diff
@@ -1254,23 +1455,26 @@ class Database extends ServiceObject {
    * @param {RunUpdateCallback} [callback] Callback function.
    * @returns {Promise<RunUpdateResponse>}
    */
-  runPartitionedUpdate(query, callback) {
+  runPartitionedUpdate(
+    query: string | ExecuteSqlRequest,
+    callback: RunUpdateCallback
+  ): void | Promise<[]> {
```
callmehiphop: The Promise response doesn't look right, also looks like we're missing overloads for this method.

src/database.ts
```diff
@@ -1742,13 +1945,16 @@ class Database extends ServiceObject {
    * region_tag:spanner_create_storing_index
    * Creating a storing index:
    */
-  updateSchema(statements, callback) {
-    if (!is.object(statements)) {
+  updateSchema(
+    statements: Schema,
+    callback: UpdateSchemaCallback
+  ): Promise<[GaxOperation, databaseAdmin.longrunning.IOperation]> {
```
callmehiphop: We should export this as `UpdateSchemaResponse`

src/session.ts
```diff
@@ -38,12 +37,25 @@ import {
...
 export interface CreateSessionCallback {
-  (err: Error | null, session?: Session | null, apiResponse?: r.Response): void;
+  (
+    err: Error | null,
+    session: Session | null,
+    apiResponse: spannerClient.spanner.v1.ISession
+  ): void;
+}
+export type CreateSessionResponse = [
+  spannerClient.spanner.v1.Session,
```
callmehiphop: Shouldn't be using this type, it refers to a protobuf message object, which we don't use.

src/table.ts
```diff
@@ -20,40 +20,43 @@ import {promisifyAll} from '@google-cloud/promisify';
 import * as is from 'is';
 import {ServiceError} from 'grpc';
 import * as through from 'through2';
-
+import {Operation as GaxOperation} from 'google-gax/build/src/longRunningCalls/longrunning';
```
callmehiphop: Why use the full path here?

src/database.ts
```diff
@@ -55,24 +65,110 @@ import {
   TransactionRunner,
 } from './transaction-runner';
 
-export interface GetDatabaseOptions {
+import {google as spannerClient} from '../proto/spanner';
+import {Schema, BasicResponse} from './common';
+import {ServiceError, CallOptions} from 'grpc';
+import {Readable, Transform} from 'stream';
+import {Stats} from 'fs';
+import {StreamProxy} from 'google-gax/build/src/streamingCalls/streaming';
```
callmehiphop: I don't think we should import this way. Maybe we should just import `Duplex` and make a union type to include `cancel()`.

callmehiphop:
Sort of a general comment, I'm still seeing `Error` used in places where `ServiceError` should be used instead.

src/database.ts
laljikanjareeya: same method in base class it give below Error:
`Property 'delete' in type 'Database' is not assignable to the same property in base type 'GrpcServiceObject'.
  Type '{ (): Promise<Empty>; (callback: DropDatabaseCallback): void; }' is not assignable to type '{ (): Promise<[Response]>; (callback: RequestCallback): void; }'.`

src/database.ts
```diff
@@ -55,24 +65,110 @@ import {
   TransactionRunner,
 } from './transaction-runner';
 
-export interface GetDatabaseOptions {
+import {google as spannerClient} from '../proto/spanner';
```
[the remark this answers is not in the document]
laljikanjareeya: will follow below, plz share your suggestion:
`import {google} from '../proto/spanner';`
`import {google as databaseAdmin} from '../proto/spanner_database_admin';`
`import {google as instanceAdmin} from '../proto/spanner_instance_admin';`

src/database.ts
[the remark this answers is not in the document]
laljikanjareeya: Done!

src/common.ts
```diff
@@ -1,18 +1,8 @@
...
-  seconds: number;
-}
+import {google as databaseAdmin} from '../proto/spanner_database_admin';
+import * as r from 'request';
+
+export type Schema =
+  | string
+  | string[]
+  | databaseAdmin.spanner.admin.database.v1.IUpdateDatabaseDdlRequest;
+export type BasicResponse = [r.Response];
```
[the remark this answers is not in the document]
laljikanjareeya: Removed!!

src/database.ts
laljikanjareeya: Done!!

src/database.ts
laljikanjareeya: Done!!

src/database.ts
laljikanjareeya: it's working correctly, as transaction is type of `BatchTransaction` and identifier is type of `TransactionIdentifier`.

JustinBeckwith:
@laljikanjareeya @AVaksman @sduskis 👋 what's the status with these?  This PR is a month old, and hasn't seen action for a while.  

Can we please drive this to resolution?

AVaksman:
I believe @laljikanjareeya addressed all requested changes, @callmehiphop Could you please take another look.

src/batch-transaction.ts
```diff
@@ -18,8 +18,15 @@ import {PreciseDate} from '@google-cloud/precise-date';
...
 import * as is from 'is';
-import {codec} from './codec';
 import {Snapshot} from './transaction';
+import {google} from '../proto/spanner';
+import {Session} from '.';
+
+export interface TransactionIdentifier {
+  session: string | Session;
+  transaction?: string;
+  readTimestamp?: string | google.protobuf.ITimestamp;
```
callmehiphop: This does not match what is returned here https://github.com/googleapis/nodejs-spanner/blob/master/src/batch-transaction.ts#L322

src/instance.ts
```diff
@@ -23,6 +23,26 @@ import * as extend from 'extend';
 import * as is from 'is';
 import snakeCase = require('lodash.snakecase');
 import {Database} from './database';
+import {SessionPoolOptions, SessionPool} from './session-pool';
+import {Operation as GaxOperation} from 'google-gax';
+import {google as databaseAdmin} from '../proto/spanner_database_admin';
+
+export interface CreateDatabaseOptions {
```
callmehiphop: This should probably extend [`ICreateDatabaseRequest`](https://github.com/googleapis/nodejs-spanner/blob/master/proto/spanner_database_admin.d.ts#L535)

src/instance.ts
```diff
@@ -41,6 +61,15 @@ import {Database} from './database';
...
 class Instance extends common.ServiceObject {
+  // tslint:disable-next-line: no-any
+  metadata: any;
+  formattedName_: string;
+  // tslint:disable-next-line: no-any
+  request: any;
+  // tslint:disable-next-line: no-any
+  requestStream: any;
+  // tslint:disable-next-line: no-any
+  databases_: Map<any, any>;
```
callmehiphop: Should be `<string, Database>`

src/instance.ts
```diff
@@ -41,6 +61,15 @@ import {Database} from './database';
  * const instance = spanner.instance('my-instance');
  */
 class Instance extends common.ServiceObject {
+  // tslint:disable-next-line: no-any
+  metadata: any;
```
callmehiphop: I think you can pull in `Metadata` from the common module.

src/instance.ts
```diff
@@ -41,6 +61,15 @@ import {Database} from './database';
...
+  formattedName_: string;
+  // tslint:disable-next-line: no-any
+  request: any;
```
callmehiphop: Can we get/make a proper type for this and `requestStream`?

src/session-pool.ts
```diff
@@ -389,7 +391,8 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
   getWriteSession(callback: GetWriteSessionCallback): void {
     this._acquire(types.ReadWrite).then(
       session => callback(null, session, session.txn!),
-      callback
+      // tslint:disable-next-line: no-any
+      callback as any
```
callmehiphop: Why do we need this?

src/session.ts
```diff
@@ -38,12 +37,25 @@ import {
...
-import {CreateSessionOptions} from './common';
+import {google as spannerClient} from '../proto/spanner';
 
 export type GetSessionResponse = [Session, r.Response];
 
 export interface CreateSessionCallback {
-  (err: Error | null, session?: Session | null, apiResponse?: r.Response): void;
+  (
+    err: Error | null,
+    session: Session | null,
```
callmehiphop: This and `apiResponse` should be optional. Maybe consider using your `ResourceCallback` generic here instead.

src/session.ts
```diff
@@ -38,12 +37,25 @@ import {
...
+    apiResponse: spannerClient.spanner.v1.ISession
+  ): void;
+}
+export type CreateSessionResponse = [
+  Session,
+  spannerClient.spanner.v1.ISession
+];
+
+export interface CreateSessionOptions {
```
callmehiphop: Shouldn't this be declared in `Database` where the `createSession` method lives?

src/session.ts
```diff
@@ -38,12 +37,25 @@ import {
...
 
 export interface CreateSessionCallback {
-  (err: Error | null, session?: Session | null, apiResponse?: r.Response): void;
+  (
+    err: Error | null,
+    session: Session | null,
+    apiResponse: spannerClient.spanner.v1.ISession
+  ): void;
+}
+export type CreateSessionResponse = [
```
callmehiphop: Same comment as `CreateSessionOptions`.

src/table.ts
```diff
@@ -20,40 +20,43 @@ import {promisifyAll} from '@google-cloud/promisify';
...
+    table: Table | null,
+    operation: GaxOperation | null,
+    apiResponse: databaseAdmin.longrunning.IOperation | null
   ): void;
 }
 
 interface DropTableCallback {
-  (err: ServiceError, operation?: null, apiResponse?: null): void;
-  (err: null, operation: d.Operation, apiResponse: d.GrpcOperation): void;
+  (
```
callmehiphop: Could probably use `ResourceCallback` generic here.

src/table.ts
```diff
@@ -20,40 +20,43 @@ import {promisifyAll} from '@google-cloud/promisify';
...
+export interface CreateTableCallback {
   (
-    err: null,
-    table: Table,
-    operation: d.Operation,
-    apiResponse: d.GrpcOperation
+    err: ServiceError | null,
+    table: Table | null,
+    operation: GaxOperation | null,
+    apiResponse: databaseAdmin.longrunning.IOperation | null
```
callmehiphop: I think all these fields should be marked optional.

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
 }
 
 export interface ReadRequest extends RequestOptions {
-  keySet?: s.KeySet;
+  session?: string | null;
```
callmehiphop: I don't think the user should ever be providing the session, so we shouldn't include it.

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
+  transaction?: TransactionSelector | null;
```
callmehiphop: I don't think the user should ever be providing this either.

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
+  table?: string | null;
+  index?: string | null;
```
callmehiphop: Why add `null`?

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
+  columns?: string[] | null;
   keys?: string[];
-  ranges?: KeyRange[];
-  index?: string;
-  columns?: string[];
-  limit?: number;
-  resumeToken?: ResumeToken;
-  partitionToken?: Uint8Array | string;
+  ranges?: spannerClient.spanner.v1.IKeyRange[];
```
callmehiphop: I think we had a custom `KeyRange` type that change this is stomping on.

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
+  keySet?: spannerClient.spanner.v1.IKeySet | null;
+  limit?: number | Long | null;
+  resumeToken?: Uint8Array | null;
+  partitionToken?: Uint8Array | null;
+}
+
+export interface TransactionSelector {
```
callmehiphop: Why not just use `google.spanner.v1.ITransactionSelector`?

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
   rowCounts: number[];
 }
 
+export type CommitRequest = spannerClient.spanner.v1.ICommitRequest;
+
 export type BatchUpdatePromise = Promise<[number[], s.ExecuteBatchDmlResponse]>;
-export type BeginPromise = Promise<[s.Transaction]>;
-export type CommitPromise = Promise<[s.CommitResponse]>;
+export type BeginPromise = Promise<[spannerClient.spanner.v1.ITransaction]>;
+export interface BeginTransactionCallback {
```
callmehiphop: Consider using `NormalCallback` generic.

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
 export interface BatchUpdateError extends ServiceError {
   rowCounts: number[];
 }
 
+export type CommitRequest = spannerClient.spanner.v1.ICommitRequest;
+
 export type BatchUpdatePromise = Promise<[number[], s.ExecuteBatchDmlResponse]>;
-export type BeginPromise = Promise<[s.Transaction]>;
-export type CommitPromise = Promise<[s.CommitResponse]>;
+export type BeginPromise = Promise<[spannerClient.spanner.v1.ITransaction]>;
```
callmehiphop: Should be `BeginResponse` and we should change the `begin` return type to be `Promise<BeginResponse>`

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
+export interface BeginTransactionCallback {
+  (err: Error | null, transaction: spannerClient.spanner.v1.ITransaction): void;
+}
+export type CommitPromise = Promise<[spannerClient.spanner.v1.CommitResponse]>;
```
callmehiphop: Should be `CommitResponse` and we should change the return type to be `Promise<CommitResponse>`

src/transaction.ts
```diff
@@ -73,25 +74,41 @@ export interface KeyRange {
...
 export type ReadPromise = Promise<[Rows]>;
-export type RunPromise = Promise<[Rows, s.ResultSetStats]>;
+export type RunPromise = Promise<
```
callmehiphop: Should be `RunResponse` and we should change the return type to be `Promise<RunResponse>`

src/transaction.ts
```diff
@@ -1284,7 +1309,7 @@ export class Transaction extends Dml {
   }
 
   commit(): CommitPromise;
-  commit(callback: s.CommitCallback): void;
+  commit(callback: spannerClient.spanner.v1.Spanner.CommitCallback): void;
```
callmehiphop: Please do not use the generated callback types.

src/transaction.ts
```diff
@@ -1331,13 +1356,15 @@ export class Transaction extends Dml {
    *   });
    * });
    */
-  commit(callback?: s.CommitCallback): void | CommitPromise {
+  commit(
+    callback?: spannerClient.spanner.v1.Spanner.CommitCallback
```
callmehiphop: Please do not use this type.

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
 
-export interface GetDatabaseOptions {
-  autoCreate?: boolean;
+import {google} from '../proto/spanner';
+import {Schema, RequestCallback, PagedRequest} from './common';
+import {ServiceError, CallOptions} from 'grpc';
+import {Readable, Transform, Duplex} from 'stream';
+import {PreciseDate} from '@google-cloud/precise-date';
+
+interface CreateBatchTransactionCallback {
```
callmehiphop: Consider use `ResourceCallback` generic here.

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
-  (err: Error | null, database?: Database, apiResponse?: r.Response): void;
+
+type CreateBatchTransactionResponse = [
+  BatchTransaction,
+  google.spanner.v1.ITransaction | google.spanner.v1.ISession
+];
+interface Cancellable {
+  cancel(): void;
 }
+type CancellableDuplex = Duplex & Cancellable;
```
callmehiphop: I think you can pull the stream types from `grpc`.

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
+type DatabaseResponse = [Database, r.Response];
+interface DatabaseCallback {
```
callmehiphop: Consider using `ResourceCallback` generic

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
+  (
+    err: ServiceError | null,
+    database?: Database,
+    apiResponse?: r.Response
+  ): void;
+}
+type GetSnapshotCallback = RequestCallback<Snapshot>;
```
callmehiphop: I think you want `ResourceCallback` here

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
+
+type GetTransactionCallback = RequestCallback<Transaction>;
```
callmehiphop: I think you want `ResourceCallback` here

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
+  new (
+    database: Database,
+    options?: SessionPoolOptions | null
+  ): SessionPoolInterface;
 }
 
-export interface GetTransactionCallback {
-  (err: Error, transaction?: null): void;
-  (err: null, transaction: Transaction): void;
+interface UpdateSchemaCallback {
```
callmehiphop: Consider using `ResourceCallback` generic here

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
+  (
+    err: ServiceError | null,
+    operation: GaxOperation,
+    resp: databaseAdmin.longrunning.IOperation
```
callmehiphop: These should be optional

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
+    [key: string]: any;
+    session?: string;
+  };
+  client?: string;
+  method?: string;
+  gaxOptions: CallOptions;
+}
+
+interface PoolRequestCallback {
+  (err: ServiceError | null, ...args: Array<{} | null>): void;
```
callmehiphop: I could be wrong, but will there ever actually be more than one param? Seems like maybe this should be a generic that accepts the response type.

src/database.ts
```diff
@@ -55,24 +66,109 @@ import {
...
+  string[],
+  databaseAdmin.spanner.admin.database.v1.IGetDatabaseDdlResponse
+];
+
+type GetSessionsCallback = RequestCallback<
+  Session,
+  google.spanner.v1.IListSessionsResponse
+>;
+
+type GetSessionsResponse =
```
callmehiphop: Can we use `PagedResponse` generic here?

src/database.ts
```diff
@@ -182,17 +283,25 @@ class Database extends ServiceObject {
...
-      session = this.session(session);
-    }
-    const transaction = new BatchTransaction(session, options);
+    const transaction: BatchTransaction = new BatchTransaction(
+      session,
+      options
+    );
     transaction.id = id;
-    transaction.readTimestamp = identifier.readTimestamp;
+    transaction.readTimestamp = identifier.readTimestamp as PreciseDate;
```
callmehiphop: I think you've stumbled upon a bug as this does not match the return type for `BatchTransaction#identifier`. I would either update this to be `timestamp` or update `identifier()` to return `readTimestamp`.

src/database.ts
```diff
@@ -182,17 +283,25 @@ class Database extends ServiceObject {
...
+    const session =
+      typeof identifier.session === 'string'
+        ? this.session(identifier.session)
+        : identifier.session;
     const id = identifier.transaction;
-    if (is.string(session)) {
-      session = this.session(session);
-    }
-    const transaction = new BatchTransaction(session, options);
+    const transaction: BatchTransaction = new BatchTransaction(
```
callmehiphop: why do you need to add a type specification here? It should be inferred

src/database.ts
```diff
@@ -182,17 +283,25 @@ class Database extends ServiceObject {
...
+      session,
+      options
+    );
     transaction.id = id;
-    transaction.readTimestamp = identifier.readTimestamp;
+    transaction.readTimestamp = identifier.readTimestamp as PreciseDate;
     return transaction;
   }
+  close(callback: SessionPoolCloseCallback): void;
+  close(): Promise<google.protobuf.IEmpty>;
```
callmehiphop: You should make a type for this

```ts
type DatabaseCloseResponse = [google.protobuf.IEmpty];
```

src/database.ts
```diff
@@ -451,7 +596,7 @@ class Database extends ServiceObject {
    * @param {Transaction} transaction The transaction to observe.
    * @returns {Transaction}
    */
-  private _releaseOnEnd(session, transaction) {
+  private _releaseOnEnd(session: Session, transaction: Transaction | Snapshot) {
```
callmehiphop: Can this be just `Snapshot` since `Transaction` extends it?

src/database.ts
```diff
@@ -460,6 +605,8 @@ class Database extends ServiceObject {
       }
     });
   }
+  delete(): Promise<[r.Response]>;
```
callmehiphop: This should have its own type and we should be using `IEmpty` in this case.

callmehiphop: Or is this another case of conflicts with the base class?

src/instance.ts
laljikanjareeya: taken care in Instance PR#[619](https://github.com/googleapis/nodejs-spanner/pull/619)

src/instance.ts
laljikanjareeya: taken care in Instance PR#619

src/instance.ts
laljikanjareeya: taken care in Instance PR#619

src/instance.ts
laljikanjareeya: taken care in Instance PR#619

src/transaction.ts
laljikanjareeya: Create new Type

src/database.ts
laljikanjareeya: used `ClientDuplexStream`

src/database.ts
laljikanjareeya: Created Generic Request response as below:
`request: <T, R = void>(config: RequestConfig,callback: RequestCallback<T, R>) => void;`

src/database.ts
laljikanjareeya: looks breaking changes to rename in identifier(), so renamed in TransactionIdentifier Property.

src/database.ts
laljikanjareeya: created new type here..

src/database.ts
laljikanjareeya: yes conflict with base class...

src/common.ts
```diff
@@ -1,18 +1,40 @@
...
-  returnTimestamp?: boolean;
-  strong?: boolean;
+import {google as databaseAdmin} from '../proto/spanner_database_admin';
+import {ServiceError, CallOptions} from 'grpc';
+
+export type Schema =
+  | string
+  | string[]
+  | databaseAdmin.spanner.admin.database.v1.IUpdateDatabaseDdlRequest;
+export type ResourceCallback<Resource, Response = void> = Response extends void
```
callmehiphop: Why are we conditionally changing the number of arguments here?

src/database.ts
```diff
@@ -55,24 +60,104 @@ import {
...
+>;
+
+type CreateBatchTransactionResponse = [
+  BatchTransaction,
+  google.spanner.v1.ITransaction | google.spanner.v1.ISession
+];
+type DatabaseResponse = [Database, r.Response];
+type DatabaseCallback = ResourceCallback<Database, r.Response>;
+
+type GetSnapshotCallback = ResourceCallback<Snapshot>;
```
callmehiphop: Should we include the response type?

src/database.ts
```diff
@@ -55,24 +60,104 @@ import {
...
+
+type GetTransactionCallback = ResourceCallback<Transaction>;
```
callmehiphop: Should we include the response type?

src/database.ts
```diff
@@ -55,24 +60,104 @@ import {
...
+export interface CreateSessionOptions {
+  name?: string | null;
+  labels?: {[k: string]: string} | null;
 }
 
+export type CreateSessionCallback = ResourceCallback<
+  Session,
+  spannerClient.spanner.v1.ISession
+>;
+export interface DatabaseDeleteCallback {
```
callmehiphop: Maybe use `NormalCallback` here?

src/database.ts
```diff
@@ -228,12 +327,22 @@ class Database extends ServiceObject {
    *   }
    * });
    */
-  close(callback) {
+  close(
+    callback?: SessionPoolCloseCallback
+  ): void | Promise<google.protobuf.IEmpty> {
```
callmehiphop: We have a different type listed in the overload for the Promise, not sure which one is correct but lets be consistent here.

src/database.ts
```diff
@@ -731,25 +877,27 @@ class Database extends ServiceObject {
...
    * });
    */
-  getSchema(callback) {
-    const reqOpts = {
+  getSchema(callback?: GetSchemaCallback): void | Promise<GetSchemaResponse> {
+    const reqOpts: databaseAdmin.spanner.admin.database.v1.IGetDatabaseDdlRequest = {
       database: this.formattedName_,
     };
-    this.request(
+    this.request<string[] | null>(
```
callmehiphop: Is this right? I would think it would be `IGetDatabaseDdlResponse`

src/database.ts
```diff
@@ -832,36 +980,46 @@ class Database extends ServiceObject {
...
+        ? (optionsOrCallback as GetSessionsOptions)
+        : {gaxOptions: {}};
+    const gaxOpts: CallOptions = options.gaxOptions as CallOptions;
+    const reqOpts = extend({}, options, {
+      database: this.formattedName_,
+    });
+
     delete reqOpts.gaxOptions;
-    this.request(
+    this.request<Session, google.spanner.v1.IListSessionsResponse>(
```
callmehiphop: Pretty sure this is wrong, it looks like the first argument is `ISession` and that we map it to a `Session` object.

src/index.ts
```diff
@@ -57,7 +58,22 @@ export interface SpannerOptions extends GrpcClientOptions {
   port?: number;
   sslCreds?: ChannelCredentials;
 }
-
+export interface RequestConfig {
+  client: string;
+  method: string;
+  reqOpts: {
```
callmehiphop: Shouldn't this be either any or a a generic type?

src/session-pool.ts
```diff
@@ -37,8 +37,7 @@ export interface SessionPoolCloseCallback {
  * @param {Session} session The read-only session.
  */
 export interface GetReadSessionCallback {
```
callmehiphop: Consider using `NormalCallback`?

src/transaction.ts
```diff
@@ -72,27 +74,42 @@ export interface KeyRange {
   endOpen?: Value[];
 }
 
+export interface KeyRang {
```
callmehiphop: `KeyRange`*

src/transaction.ts
```diff
@@ -114,6 +131,13 @@ export interface RunUpdateCallback {
   (err: null | ServiceError, rowCount: number): void;
 }
 
+export interface CommitCallback {
```
callmehiphop: `NormalCallback`?

src/transaction.ts
callmehiphop: Also is this correct? It looks like we actually offer a custom `KeyRange` type and convert each bound to a `IListValue` per https://github.com/googleapis/nodejs-spanner/blob/102cae1970d5e357e3dd7dea6e29b39dd194fcd4/src/transaction.ts#L893

src/database.ts
laljikanjareeya: `GetSnapshotCallback` has only two argument as Error and Snapshot.

src/database.ts
laljikanjareeya: `GetTransactionCallback` has only two argument as Error and Transaction.

src/common.ts
laljikanjareeya: added `Response` required type and used `NormalCallback<Resource>`  instead `ResourceCallback<Resource>`

src/database.ts
laljikanjareeya: Done

src/index.ts
laljikanjareeya: currently added as `any` will see how this is used in other file and if we can make it generic than will do in subsequent PR.

src/transaction.ts
laljikanjareeya: removed interface `KeyRang` and used existing one.

src/database.ts
callmehiphop: Ah, my mistake.

src/database.ts
callmehiphop: Looks like this still uses 2 different types for the Promise generic type.

src/database.ts
```diff
@@ -832,36 +979,47 @@ class Database extends ServiceObject {
...
-        if (sessions) {
-          arguments[1] = sessions.map(metadata => {
-            const session = self.session(metadata.name);
+      (err, ...args) => {
+        if (args[0]) {
+          args[0] = args[0].map(metadata => {
+            const session = self.session(metadata.name!);
             session.metadata = metadata;
-            return session;
+            return session as google.spanner.v1.ISession;
```
callmehiphop: This isn't right. I would suggest doing something like

```js
(err, sessions, ...args) => {
  let sessionInstances = null;
  if (sessions) sessionInstances = sessions.map(metadata => ...);
  callback!(err, sessionInstances, ...args);
}
```

src/database.ts
```diff
@@ -1034,20 +1197,20 @@ class Database extends ServiceObject {
    * @param {object} config Request config
    * @returns {Stream}
    */
-  makePooledStreamingRequest_(config) {
+  makePooledStreamingRequest_(config: RequestConfig): Readable {
     const self = this;
     const pool = this.pool_;
-    let requestStream;
-    let session;
+    let requestStream: ClientDuplexStream<RequestConfig, Readable>;
```
callmehiphop: Pretty sure this is incorrect, the first generic type should be the grpc message being sent and the second should be the grpc message received. Making this method a generic might be more trouble than it is worth (unless you disagree?) so I might just use a plain `Duplex` here instead.

src/transaction.ts
```diff
@@ -105,10 +98,10 @@ export type BeginResponse = [spannerClient.spanner.v1.ITransaction];
 export type BeginTransactionCallback = NormalCallback<
   spannerClient.spanner.v1.ITransaction
 >;
-export type CommitResponse = [spannerClient.spanner.v1.CommitResponse];
+export type CommitResponse = [spannerClient.spanner.v1.ICommitResponse];
 
 export type ReadResponse = [Rows];
-export type RunResponse = [Rows, spannerClient.spanner.v1.ResultSetStats];
+export type RunResponse = [Rows, s.ResultSetStats];
```
callmehiphop: Any reason why we don't use the `spannerClient.spanner.v1` version?

src/database.ts
```diff
@@ -1008,15 +1012,16 @@ class Database extends ServiceObject {
         reqOpts,
         gaxOpts,
       },
-      (err, ...args) => {
-        if (args[0]) {
-          args[0] = args[0].map(metadata => {
+      (err, sessions, ...args) => {
+        let sessionInstances: google.spanner.v1.ISession[] | null = null;
```
callmehiphop: This should be `Session[] | null`

src/database.ts
callmehiphop: SGTM

laljikanjareeya:
@callmehiphop can we go ahead for merge?

Issue #535, opened 2019-02-19
System test flakiness


callmehiphop:
Looks like running multiple system test tasks concurrently causes some failed assertions when querying resources at the project level.

JustinBeckwith:
@stephenplusplus can you take a peek here?  This is causing lots of PR re-runs and backups.

JustinBeckwith:
Things are looking a little more stable today ...

PR #629, opened 2019-05-29
refactor: (system-test) switch to uuid.v4 to improve random id generation

AVaksman:
Towards #535

* switch to uuid.v4 to improve random id generation for resource names.
* fix flake with inserting and querying multiple rows tests.


callmehiphop:
@bcoe @JustinBeckwith I think this should be safe to merge. The system-test failures relate to `PQueue`, which unfortunately we still can't upgrade at this time.

bcoe:
@callmehiphop what's limiting us from upgrading `PQueue`. 

Either way, could we comment out the one test for the time being, with a `TODO` note, I'd rather that than force land failing tests (I don't actually have the permission levels to force merge honestly).

callmehiphop:
@bcoe I think we can upgrade now (famous last words), I think our `PQueue` issues *should* be resolved (see https://github.com/sindresorhus/p-queue/pull/68 and https://github.com/sindresorhus/p-queue/pull/73). I'll take a crack at it this afternoon, unless some one beats me to it first.

PR #630, opened 2019-05-29
refactor(system-test): implement logic to clean up old instances

AVaksman:
Towards #535

- [x] Tests and linter pass


system-test/spanner.ts
```diff
@@ -4405,6 +4405,31 @@ async function deleteTestInstances() {
...
 }
 
+async function deleteOldTestInstances() {
+  const [instances] = await spanner.getInstances();
+  // Leave only instances that contain PREFIX in their name
+  // and where created more that an hour ago.
+  const toDelete = instances.filter(
+    instance =>
+      instance.id.includes(PREFIX) &&
+      (Math.round(Date.now() / 1000) -
```
stephenplusplus: Could this be moved out and coupled with some logical variable naming so that it's easier to figure out what's going on? Feels a bit magical right now.

system-test/spanner.ts
```diff
@@ -4405,6 +4405,31 @@ async function deleteTestInstances() {
...
+        1
+  );
+
+  const limit = pLimit(5);
+  return Promise.all(
+    toDelete.map(instance =>
+      limit(() =>
+        setTimeout(() => {
+          instance.delete();
+        }, 500)
```
stephenplusplus: Maybe move to a variable whose name also explains why the delay is necessary.

system-test/spanner.ts
AVaksman: Refactored calculations as method with meaningful name

system-test/spanner.ts
AVaksman: Refactored and clarified as per original [comment](https://github.com/googleapis/nodejs-spanner/blob/bf778cc0936c16ee0ef7b396c4698c2a1df30ce5/system-test/spanner.js#L4079)

bcoe:
@stephenplusplus  are you feeling ready to sign off on this?

@AVaksman any idea what's going on with the system-test, would be nice to get it green before we land.

PR #634, opened 2019-05-31
feat: support apiEndpoint override

JustinBeckwith:


callmehiphop:
@JustinBeckwith probably should just hammer this unless we want to rollback/pin p-queue first.

callmehiphop:
Although some unit tests would be nice too :)

PR #644, opened 2019-06-08
fix(deps): upgrade to google-gax 1.x

### LINKED ISSUE #622: Upgrade to google-gax 1.0
I noticed that we're still on an old version of gax.  We need to update to the latest, and add the explicit dependency on grpc (not grpc-js) here.

JustinBeckwith:
Fixes #622.  

@crwilcox / @callmehiphop / @alexander-fenster  - I am a little freaked out that these tests pass 😆 This is using `google-gax@1.x`, which swaps out native `grpc` for `grpc-js`. With the whole `gcp-grpc` thing, I figured this would cause test failures.  But .. it just doesn't.

Then, I noticed that we're not using `grpc-gcp` anywhere 🤔 

TL;DR:  I don't know what's real anymore


alexander-fenster:
@JustinBeckwith The whole `grpc-gcp` thing was in gax and is removed from there in v1.0.0 (because it does not work with `@grpc/grpc-js`). In spanner, we only set its options:

https://github.com/googleapis/nodejs-spanner/blob/ace135997c5a5aa2948dfe9f9dfbf9bf55b4fb33/src/index.ts#L225

which will be ignored by gax v1.0.0 (but no failures will occur). Nothing will break, but the functionality will be lost.

To move to gax v1.0.0, someone would need to copy the `grpc-gcp` functionality that was removed from gax into the spanner code, and pass the C-core `grpc` as a parameter. Here is the removed code:

https://github.com/googleapis/gax-nodejs/pull/484/files#diff-bb99529408df06caafed25d9e0cd083fL298

```js
const apiConfigDefinition = options['grpc_gcp.apiConfig'];	
if (apiConfigDefinition) {	
  const apiConfig = grpcGcp.createGcpApiConfig(apiConfigDefinition);	   
  grpcOptions['channelFactoryOverride'] = grpcGcp.gcpChannelFactoryOverride;	
  grpcOptions['callInvocationTransformer'] =	
    grpcGcp.gcpCallInvocationTransformer;	
  grpcOptions['gcpApiConfig'] = apiConfig;	
}
```

Basically, this exact code should live in spanner, before client constructor call. Import `grpc-gcp`, do the same, but prefix all options with `grpc.` (e.g. `grpc.channelFactoryOverride`), and pass them as options to the client constructor alongside with the C-core `grpc` object in `src/index.ts`:

```js
// Enable grpc-gcp support
const apiConfig = grpcGcp.createGcpApiConfig(gcpApiConfig);
const grpcGcpOptions = {
  "grpc.channelFactoryOverride": grpcGcp.gcpChannelFactoryOverride,
  "grpc.callInvocationTransformer": grpcGcp.gcpCallInvocationTransformer,
  "grpc.gcpApiConfig": apiConfig
};
options = Object.assign({'grpc_gcp.apiConfig': gcpApiConfig}, options);
```

I did not check this code but should be pretty much something like this. Also, some tests that verify the options are passed down are needed.

alexander-fenster:
(note: I updated the code in the comment above)

JustinBeckwith:
Closing this out, and leaving it to @stephenplusplus to drive forward 👍

stephenplusplus:
Please let me know if this was the right way to go! #651

PR #648, opened 2019-06-18
build: switch to GitHub magic proxy, for release-please

bcoe:


callmehiphop:
@bcoe looks like this generated with an older version of `gts` installed, going to need to run a `fix`.

PR #651, opened 2019-06-21
fix(deps): upgrade to google-gax 1.x

stephenplusplus:
Carrying on from #644, this upgrades to google-gax@1, making the suggestions provided by @alexander-fenster.

#### To Dos

- [ ] Unit tests


stephenplusplus:
I think that would be on the gax end. We can confirm we're passing the right options, but if gax says and tests that the options we pass work, then we should trust that per separation of concerns.

alexander-fenster:
@stephenplusplus The changes look good to me. Some tests failed because the options object format has changed, this is expected I guess.

As for @JustinBeckwith's suggesting to check if `grpc-gcp` is actually used, I can suggest trying to `sinon.spy` the functions that is being passed in the config object (`gcpChannelFactoryOverride` and `gcpCallInvocationTransformer `) and see if they are actually called. The problem here is that to make this test work, you'll need to mock a huge part of gRPC so it might make sense to do it in as a part of a system test and not the unit tests, so that we could just `sinon.spy` this two functions and just use the real service without a need to mock it.

JustinBeckwith:
I guess the question here is - what value does `grpc-gcp` provide.  Presumably it does *something*.

stephenplusplus:
I will definitely test that we're passing the grpc-gcp things to the (eventual) GAX constructor. But speaking of if we're using grpc or not, I would just suggest checking that we pass it to GAX. GAX should have its own test that confirm it uses grpc when options.grpc is set.

alexander-fenster:
@stephenplusplus I oppose that. It was my fault to let the original design pass through. Gax should know nothing about how the underlying transport works, and the best it can do is to pass any `grpc.*` options down to gRPC (whatever gRPC implementation is used), without doing an extra job trying to parse and understand these options.

I like thinking of Gax as of one more layer in the stack, that uses gRPC (using the common well-defined gRPC API so that any compatible gRPC implementation can be used) and that does not dig deep into how exactly that gRPC works. Gax has the default gRPC implementation to use (`@grpc/grpc-js`) but in the perfect world, it should not have any gRPC dependency at all, just accepting the whatever 🦆-typed gRPC that is passed to it.

In this picture, `grpc-gcp` is basically a _property_ of the gRPC implementation to be used (you're kind of sending a mixture of `grpc` + `grpc-gcp` instead of just gRPC), so the only thing you can ask Gax is to pass down the corresponding options. To check that the gRPC (that client library supplied) actually calls the functions that the client library also supplied is something that client library can logically do, not the Gax.

(having that said, Gax does have a test that all `grpc.` options are actually passed down to gRPC - but it cannot check the _semantics_ of these options because it should not know anything about that)

stephenplusplus:
> and the best it can do is to pass any grpc.* options down to gRPC (whatever gRPC implementation is used), without doing an extra job trying to parse and understand these options.

I don't think it should do an extra job, either. It should just make sure that the options are given to wherever they go.

> To check that the gRPC (that client library supplied) actually calls the functions that the client library also supplied is something that client library can logically do, not the Gax.

Both this library and gax should follow the same system here-- we shouldn't have to double check that a dependency is doing what it says it does. We should trust the test suite of the dependency. nodejs-spanner trusts that the options we pass to gax are passed to grpc. And gax trusts that the options is passes to grpc are used properly by grpc.

alexander-fenster:
The test that verifies that the options are passed down to gRPC is here: https://github.com/googleapis/gax-nodejs/blob/9e4d6ed7e444c3b4d64857d24a0ad0f51919bf17/test/grpc.ts#L171

They are really passed down there, I promise :)

But we do need a test that verifies that `grpc-gcp` actually works with Spanner. I do think it should be done in some system test for Spanner. At the very least, check that callback functions passed as parameters are called; even better is to check that something meaningful is going on (if not, why do we need this `grpc-gcp` at all :) )

bcoe:
@stephenplusplus anything blocking getting this over the finish line, would love to unblock it.

stephenplusplus:
Sorry, I didn't mean to question the test quality of any library-- I never even checked. I was just trying to explain the separation of testing responsibility that made sense to me.

> At the very least, check that callback functions passed as parameters are called; even better is to check that something meaningful is going on (if not, why do we need this `grpc-gcp` at all :) )

I actually don't know :) I'm still not clear why we need to double-check other libraries' behavioral guarantees. If `grpc-gcp` accepts these options, that should be proof that they are valid, and will be appropriately handled by grpc. Under that presumption, all that should be tested is that we pass these parameters, not that they are called (not our code that calls them? not our responsibility to test).

> But we do need a test that verifies that grpc-gcp actually works with Spanner.

This can definitely negate my previous confidence-- what specifically needs to work, and why wouldn't it?

alexander-fenster:
Hey @stephenplusplus,

So, here is how I see it. I totally get the point of having the library tests test the library functionality - that is how our unit tests work. In spanner, unit tests test spanner library up to the point where the request is taken care of by gax; in gax, they test gax features (e.g. if the parameters are passed down to gRPC).

We also have system tests that cover more than just this library's code: they test the functionality end to end: from sending a request to getting response, including all the stack of technologies under the hood (gax, gRPC, actual API implementation). The failure of a system test does not necessarily mean that the spanner library has a problem (the problem might as well be in gax, in gRPC, or in API server-side implementation), but it will likely mean that the end user will have a problem, and we just happen to be the ones who detect it.

So... back to `grpc-gcp` stuff. For now, we take some set of parameters and pass it to the client constructor. The client constructor knows to pass the same set of parameters to gax. Gax is tested and is believed to pass the same set of parameters to gRPC, which is believed to call the function being passed, that changes it behavior (in a way I have no idea about). That's why we have all this `grpc-gcp`: to change the gRPC behavior.

Now let me ask you: does the fact that we unit-test all the puzzle pieces individually means that `grpc-gcp` actually _works_ for Spanner? Quite possibly, it might mean that the puzzle pieces perfectly stick together, but the resulting picture does not make any sense :)

That's why I'm suggesting to:
1. Understand what functionality `grpc-gcp` provides for end users;
2. Have a system test that verifies that adding `grpc-gcp` options and callback functions actually makes this functionality used by gRPC (at the very least, that the callback function that we pass is called).

I don't have any authority here on this library, just trying to be logical: we can't say we _support_ `grpc-gcp` unless we actually checked that it works. Does this approach make sense for you @stephenplusplus? How about you folks @JustinBeckwith @bcoe?

Thanks!

alexander-fenster:
TL;DR: I don't agree with this:

> not our code that calls them? not our responsibility to test

We are the last Google-supported layer of the stack before the actual customer, so we might as well need to test everything :)  You folks might have a different opinion here, I'm OK with that.

stephenplusplus:
I agree with both of us :) Thanks for the time to explain your position. The way we used to think of it was, unit tests test library code and system tests test our docs and examples (not 1:1, but in general, does the lib do what we say it does).

We should get specific on what grpc-gcp does and figure out if we think it's at risk for failing with Spanner. Does anyone know what it is? The README isn't very informative.

alexander-fenster:
@WeiranFang for the rescue!

WeiranFang:
I might miss some of the content, so please correct me if I'm wrong. By reading this thread, it seems like grpc-gcp has been moved out of gax because it's not compatible with grpc-js, and the spanner library would need to pass both the grpc-native-core and grpc-gcp as options into gax. Is that right?

If that's the case, are we planning to move it back to gax once grpc-gcp is compatible with grpc-js? FYI we will start working on this to resolve the incompatibility issue in Q3. 

To answer some of the questions:

> Understand what functionality grpc-gcp provides for end users

grpc-gcp is a grpc extension that supports GCP specific features that are not supported (or may never be) in grpc itself. And those features are normally service specific (i.e. spanner and bigtable may have different requirements). So this library acts as an extension point to carry both the underlying features and API/Service specific configurations when using grpc clients.

Currently one of its major features is to support grpc channel management/pooling. This is to solve the GFE max 100 concurrent streams on TCP connection.

> Have a system test that verifies that adding grpc-gcp options and callback functions actually makes this functionality used by gRPC (at the very least, that the callback function that we pass is called).

Not sure if it helps but the library has an [integration test](https://github.com/GoogleCloudPlatform/grpc-gcp-node/blob/master/grpc-gcp/test/integration/spanner_test.js) to test the channel pool functionality for spanner stub.

Let me know if there's any other questions.

alexander-fenster:
Thanks @WeiranFang! As for moving it back to gax, we can discuss it later when the support for `grpc-js` is implemented. To be fair, I like the idea of passing `grpc-gcp` options all the way down without having any code in gax looks cleaner from the architectural point of view for me. But let's get back to this discussion later when `grpc-gcp` supports `grpc-js`. Until that time, it just cannot be a part of gax because of the extra dependency (`grpc`) it pulls.

AVaksman:
Just referencing to the original issue #622

stephenplusplus:
Pinging for another review. I have added the tests to confirm we are passing through the correct things from grpc-gcp.

PR #656, opened 2019-06-28
docs(samples): Add Datatypes examples to Spanner sample

jsimonweb:


samples/datatypes.js
```diff
@@ -0,0 +1,696 @@
...
+  } catch (err) {
+    console.error('ERROR:', err);
+  } finally {
+    // Close the database when finished.
+    database.close();
+  }
+  // [END spanner_query_with_timestamp_parameter]
+}
+
+require(`yargs`)
```
bcoe: we haven't been using yargs for our new sample format (the idea being that it can cause some confusion for customers, as they need to install a dependency).

Here's the approach we've been taking:

https://github.com/googleapis/nodejs-web-risk/blob/master/samples/quickstart.js#L46

Doesn't need to be a blocker from me, but I'd rather move towards consistency.

PR #692, opened 2019-09-17
feat(database): batch create sessions

callmehiphop:
There's a decrease in code coverage, however I believe this feature is tested at 100%.

- [x] Tests and linter pass


src/session-pool.ts
```diff
@@ -542,62 +548,67 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+    const labels = this.options.labels!;
+
+    this._pending += count;
+
+    let sessions: Session[] | null = null;
+
     try {
-      await this._createSession(type);
-      this.emit('available');
+      [sessions] = await this.database.batchCreateSessions({count, labels});
```
skuruppu: It's possible that the number of sessions returned is less than what we ask for. So we need to do this in a while loop until the number of remaining sessions is zero.

src/session-pool.ts
```diff
@@ -542,62 +548,67 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+          } catch (e) {
+            type = types.ReadOnly;
+          }
+        }
+
+        session.type = type;
+        session.lastUsed = Date.now();
+
+        this._pending -= 1;
+        this._inventory[type].push(session);
```
skuruppu: I'm not familiar with this client library yet so correct me if I'm wrong. Is the structure that gives a session to the user when they want to perform a query a FIFO or a LIFO? As part of this change, we also want to use sessions in LIFO order to improve backend caching performance so that read/write latencies improve for the client.

skuruppu: Does Node.js have multiple gRPC channels? If there are multiple channels then the sessions are meant to be shuffled across the channels.

callmehiphop: Ah, I must have missed that part, this currently is a fifo.

callmehiphop: We use a library called `grpc-gcp` that pools channels for us, so yes!

src/session-pool.ts
callmehiphop: Thanks for the heads up, went ahead and made another commit to do just that.

bcoe:
@callmehiphop I believe the decrease in coverage is related to a slight bug with default parameters showing up as uncovered.

I wouldn't worry about such a small drop in coverage.

src/session-pool.ts
skuruppu: Ah ok. I don't know if this works in the same way in Node.js. But in other languages, if a session is created through a given channel, then that session has an affinity to that channel. The problem when batchCreateSessions is introduced is that you allow a bunch of sessions to be created by a single channel, especially in the pool initialization phase, so that channel ends up being more utilized than other channels as the user makes requests.

This is why the code example in the spec is something like:
```
    // Spread session creation evenly over all available channels.
    int sessions_per_channel = min_sessions / num_channels;
    for (int i = 0; i < num_channels; i++) {
        sessions = batchCreateSessions(sessions_per_channel);
    }
```

Is there a way to get the number of channels you have available and spread the calls to batchCreateSessions over all the channels so that each channel gets its own portion of the sessions that are initialized?

src/database.ts
```diff
@@ -258,6 +273,103 @@ class Database extends ServiceObject {
...
+   * **It is unlikely you will need to interact with sessions directly. By
+   * default, sessions are created and utilized for maximum performance
+   * automatically.**
+   *
+   * Wrapper around {@link v1.SpannerClient#batchCreateSessions}.
+   *
+   * @see {@link v1.SpannerClient#batchCreateSessions}
+   * @see [BatchCreateSessions API Documentation](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.BatchCreateSessions)
+   *
+   * @param {BatchCreateSessionsOptions} [options] Configuration object.
```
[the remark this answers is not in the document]
callmehiphop: I want to make sure the users can customize labels, but I don't want to add more parameters. We don't have overloading, so this is typically achieved by accepting a configuration object. I went ahead and made a commit that lets the user supply either a count or an object containing a count property. PTAL!

src/session-pool.ts
WeiranFang: Hi @skuruppu @callmehiphop ,

The spec of grpc-gcp channel pool is to spin out new channels based on the number of concurrent streams on existing channels. So every time the number of open streams exceed the watermark value (configurable by client) for all the existing channels, a new channel will be created and added to the channel pool, until it reaches the upper bound pool size (also configurable by client). 

Can you help me better understand this batchCreateSessions function?

1. Is this a new Unary gRPC method (i.e. a single request with single response) or just a wrapper of existing `CreateSession` call? If it's a new unary rpc, I assume the response contains a list of created sessions from server. Then in this case, we would need to support batch BIND (binding session affinity to a channel). And I believe there are similar works being done in grpc-gcp-csharp to support batch BIND.

2. Are sessions expected to be evenly distributed across channels? If so, we may need to add this requirement to the spec of grpc-gcp and add this feature in grpc-gcp across all languages .

skuruppu: > 1. Is this a new Unary gRPC method (i.e. a single request with single response) or just a wrapper of existing CreateSession call? 

It's a unary gRPC method so the response contains a list of created sessions from server as you said. So sessions currently don't have an affinity to the channel that it was created on?

> 2. Are sessions expected to be evenly distributed across channels?

Yes because if there are multiple channels, we want to utilize them as evenly as possible. In Java for example, if you call BatchCreateSessions on a particular channel, the returned sessions have an affinity to that channel. So we just need to call BatchCreateSessions using the logic I put in the code snippet above. Based on what you said in point 1, it doesn't work that way in Node.js currently without batch BIND.

I don't want to block this PR based on this discussion since it seems like a few other changes need to be made to support evenly distributing sessions across channels. So I'm happy to file an issue to do this work later.

Let me know if this works for you @WeiranFang.

bcoe:
don't have a super strong opinion on this feature, going to leave it to @jkwlui, @crwilcox, and company to chime in.

src/session-pool.ts
skuruppu: Resolving conversation since the changes that need to be made are out of scope for this PR.

skuruppu (approved):
Approval for the logic in the PR. I will leave it up to repo owners to approve on the Node.js specifics.

src/session-pool.ts
WeiranFang: Thanks for the input! Then I believe we have same requirements (batch bind support & session balanced connection pool) for other languages as well. Please feel free to file an issue to grpc-gcp, also PR is welcomed :)

bcoe (approved):

crwilcox:
@callmehiphop it seems some files are missing license headers?

callmehiphop:
@crwilcox yep, but the file in question is generated so while I could add them, they're likely to magically disappear in the future. @bcoe @alexander-fenster any way around this?

bcoe:
I don't think we should block merging on license headers for generated code, but we've discussed this in chat with @alexander-fenster (and I'm hoping we'll start populating headers appropriately soon).

callmehiphop:
@bcoe SGTM.

@crwilcox did you have a chance to review this before I merge?

src/session-pool.ts
```diff
@@ -542,61 +548,60 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+    const labels = this.options.labels!;
+
+    let needed = reads + writes;
+    this._pending += needed;
+
+    while (needed > 0) {
+      let sessions: Session[] | null = null;
+
+      try {
+        [sessions] = await this.database.batchCreateSessions({
```
crwilcox: I seem to recall that we don't want to request more than 100 sessions at a time. I believe this isn't a hard limit but to do with channel affinity of sessions and there being performance implications when we exceed 100 sessions on a single channel. For this reason we might want to limit the number requested to the minimum of 100 or needed?

crwilcox (changes requested):

src/session-pool.ts
crwilcox: I dug a bit more and for sure the java mock emulates this https://github.com/googleapis/google-cloud-java/blob/d6182e6072baac276c413654094dec77ae0b5d69/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java#L616

The backend might already only send you 100 even if you ask for more so we may not need to do this client side.

/cc: @skuruppu

crwilcox: I found a comment that clarifies this. We can request as many sessions. Returned will be, at most, `min(requested, 100)`. Consider this optional. Though it might be good to document/call out in a comment.

crwilcox (approved):

src/session-pool.ts
callmehiphop: Good to know! I put a comment in that mentions this.

PR #698, opened 2019-09-26
refactor: enable noImplicitAny table.ts 

laljikanjareeya:
towards #223 

- [ ] Tests and linter pass


src/table.ts
```diff
@@ -795,16 +797,17 @@ class Table {
...
     this.database.runTransaction((err, transaction) => {
       if (err) {
         callback(err);
         return;
       }
 
-      transaction[method](this.name, rows);
-      transaction.commit(callback);
+      // tslint:disable-next-line: no-any
+      (transaction as any)![method](this.name, rows);
```
callmehiphop: Why do we need to cast this to `any`?

src/transaction-runner.ts
```diff
@@ -48,8 +48,7 @@ export interface RunTransactionOptions {
  *     already been created, and is ready to be queried and committed against.
  */
 export interface RunTransactionCallback {
-  (err: ServiceError, transaction?: null): void;
-  (err: null, transaction: Transaction): void;
+  (err: ServiceError | null, transaction?: Transaction | null): void;
```
callmehiphop: Could this just be `NormalCallback<Transaction>`?

laljikanjareeya: Done!

src/table.ts
laljikanjareeya: As We don't have any key-value property in `Transaction`, it gives below error:
`No index signature with a parameter of type 'string' was found on type 'Transaction'`

callmehiphop: @laljikanjareeya interesting, what would happen if we changed `method` from `string` to something like

```ts
method: 'delete' | 'insert' | 'etc'
```

laljikanjareeya: Yes, It's working Thanks!

PR #711, opened 2019-10-21
fix: don't wrap SpannerDate so timezone does not affect results

### LINKED ISSUE #710: samples-test - ARRAY query parameter - failing for timezone with positive offset
Running `npm run samples-test` results in the `should use an ARRAY query parameter to query record from the Venues example table` test failing for me.  Looks to be caused by Date timezone offset - my machine is in Australia/Sydney so timezone is AEST/AEDT with offset +10:00/+11:00.

Error message:

```
  52 passing (7m)
  1 failing

  1) Spanner
       should use an ARRAY query parameter to query record from the Venues example table:
     AssertionError: expected 'VenueId: 19, VenueName: Venue 19, AvailableDate: 2020-10-31\nVenueId: 42, VenueName: Venue 42, AvailableDate: 2020-09-30\n' to match /VenueId: 19, VenueName: Venue 19, AvailableDate: 2020-11-01/                                                                                                                                                                                 
      at Context.it (system-test/spanner.test.js:576:12)                                                                                                                                                 



npm ERR! Test failed.  See above for more details.
```

Debugging into this test, in `datatypes.js` I can see in `queryWithArray()` that the result of `JSON.stringify(availableDate).substring(1, 11)` is producing the wrong result, where `JSON.stringify(availableDate)` is showing a date one day previous than expected modified with a time with my timezone offset factored in.

Looking at another test that deals with dates - `queryWithDate()` - this one currently is passing.  Digging a bit deeper, the difference looks to be caused by `queryWithDate()` doing a `JSON.stringify()` on the raw value coming out of the spanner row versus `queryWithArray()` which wraps the row value in a `new Date()`.  The raw value is a `SpannerDate` object which extends `Date`, so there's no need to wrap.  And `SpannerDate` has special logic for conversion which deals with timezones, so it seems better to deal with these rather than wrapping to make this logic get used.

prunge-helix:
This change allows the `should use an ARRAY query parameter to query record from the Venues example table` test to pass when run on machines with time zones with positive offsets.

Fixes #710

### DIFF
--- a/samples/datatypes.js
+++ b/samples/datatypes.js
@@ -190,7 +190,7 @@ async function queryWithArray(instanceId, databaseId, projectId) {
   try {
     const [rows] = await database.run(query);
     rows.forEach(row => {
-      const availableDate = new Date(row[2]['value']);
+      const availableDate = row[2]['value'];
       const json = row.toJSON();
       console.log(
         `VenueId: ${json.VenueId}, VenueName: ${


prunge-helix:
@googlebot I signed it!

bcoe:
@prunge-helix thanks for the patch :smile:

I think this would be a breaking change, so we need to be a bit careful pulling it in. I've looped in @stephenplusplus and @callmehiphop who know more about @google-cloud/spanner than myself.

callmehiphop:
@bcoe why would this be a breaking change? It looks like it is just a fix for a sample

bcoe:
@callmehiphop @prunge-helix my bad, I didn't notice it was a sample file changing :smile: 

please go ahead and land.

PR #733, opened 2019-11-13
Mb.update synthtool

mbril:
Updated with latest protos from Cloud Spanner


mbril:
I was running with the stable version.
Updated to latest, re-running synthtool now

skuruppu:
> this looks like it was generated with an old version of `synthtool`, note that it switches us to an old version of `.jsdoc.js` and reverts various things in `.kokoro`.
> 
> I'd recommend running the version of synthool here:
> 
> https://github.com/googleapis/synthtool

I'm curious @bcoe. Is this not done automatically in this repo with PRs like https://github.com/googleapis/nodejs-spanner/pull/719?

bcoe:
@skuruppu correct, this runs daily at 5AM PDT, sometimes it's reasonable to run the task manually (perhaps you have something you want to land during working hours); but it should be updated within a day either way.

PR #742, opened 2019-12-08
feat: support environment variable SPANNER_EMULATOR_HOST

olavloite:
The Spanner client will now connect to a custom host if the environment variable SPANNER_EMULATOR_HOST has been set to a valid `host:port` combination.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -212,6 +212,39 @@ class Spanner extends GrpcService {
    */
   static COMMIT_TIMESTAMP = 'spanner.commit_timestamp()';
 
+  /**
+   * Gets the configured Spanner emulator host from an environment variable.
+   */
+  static getSpannerEmulatorHost():
+    | {endpoint: string; port?: number}
+    | undefined {
+    const endpointWithPort = process.env.SPANNER_EMULATOR_HOST;
+    if (endpointWithPort) {
+      if (
+        endpointWithPort.startsWith('http:') ||
+        endpointWithPort.startsWith('https:')
+      ) {
+        throw new Error(
+          'SPANNER_EMULATOR_HOST must not start with a protocol specification (http/https)'
+        );
+      }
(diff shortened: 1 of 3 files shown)


olavloite:
@skuruppu PTAL
I haven't been added to this repo yet, so I can't assign you as a reviewer :-)

callmehiphop:
@skuruppu I'm going to defer to your judgement here - I tried looking at the official docs and gcloud cli but found no mention of a Spanner emulator in either.

src/index.ts
```diff
@@ -239,6 +272,16 @@ class Spanner extends GrpcService {
       },
       options || {}
     ) as {}) as SpannerOptions;
+    const emulatorHost = Spanner.getSpannerEmulatorHost();
+    if (
+      emulatorHost &&
+      emulatorHost.endpoint &&
+      emulatorHost.endpoint.length > 0
+    ) {
+      options.servicePath = emulatorHost.endpoint;
```
hengfengli: What is the difference between `apiEndpoint` vs `servicePath ` in options? It seems that `apiEndpoint` has a higher priority and why don't we set `apiEndpoint` here instead of `servicePath`? 

https://github.com/googleapis/nodejs-spanner/blob/32fb6ca547b96fac1d742fc7580e78be069d2c02/src/index.ts#L286-L289

olavloite: To my understanding, the `apiEndpoint` option is intended as a manual override for the endpoint to use that can be specified by the user, and that should take precedence over any default settings. The `servicePath` is the option that can be used to override the default service endpoint of the client. By setting the value of `SPANNER_EMULATOR_HOST` on this option, we ensure that any user that has manually set a value for `apiEndpoint` will get the same behavior regardless of whether `SPANNER_EMULATOR_HOST` has been set.

olavloite:
Closed in favor of #749 which also contains this change.

PR #745, opened 2019-12-08
fix: restore SessionLeakError name after super call

### LINKED ISSUE #744: SessionLeakError has generic Error name and prototype
The `SessionLeakError` that is returned by the session pool when a session is leaked has the generic name `Error`. This is probably a regression caused by https://github.com/Microsoft/TypeScript/issues/13965.

olavloite:
The name of a `SessionLeakError` is set to `Error` by its super constructor. This makes checking specifically for a `SessionLeakError` more difficult.

Fixes #744

### DIFF
--- a/samples/struct.js
+++ b/samples/struct.js
@@ -193,7 +193,8 @@ async function queryWithArrayofStruct(instanceId, databaseId, projectId) {
     sql:
       'SELECT SingerId FROM Singers ' +
       'WHERE STRUCT<FirstName STRING, LastName STRING>(FirstName, LastName) ' +
-      'IN UNNEST(@names)',
+      'IN UNNEST(@names) ' +
+      'ORDER BY SingerId',
     params: {
       names: bandMembers,
     },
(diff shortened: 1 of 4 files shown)


samples/struct.js
```diff
@@ -193,7 +193,8 @@ async function queryWithArrayofStruct(instanceId, databaseId, projectId) {
     sql:
       'SELECT SingerId FROM Singers ' +
       'WHERE STRUCT<FirstName STRING, LastName STRING>(FirstName, LastName) ' +
-      'IN UNNEST(@names)',
+      'IN UNNEST(@names) ' +
+      'ORDER BY SingerId',
```
callmehiphop: Seems like this change is unrelated to what the PR description is aiming to fix. Was this snuck in on accident?

olavloite: Well, not completely on accident. Strangely enough, the system test using this example failed repeatedly for this PR (see the build log of the above commits). I have not been able to find any logical reason why the change to `SessionLeakError` should have any impact on this test, but I made the change to get the build successful.
I do agree however that it would probably be better to move that change into a separate PR. I'll look into that.

If you do have any logical explanation for why the change to `SessionLeakError` could have any (indirect) impact on this test, then I'm all ears. :-)
Or is this test known to be flaky?

callmehiphop: Looks like it is failing in our latest release PR, not really sure what changed? 🤷‍♂

PR #754, opened 2019-12-16
fix: return different databases for different pool options

### LINKED ISSUE #753: Calling instance.database(name, options) twice with same name and different options returns same database
Calling `instance.database(name, poolOptions)` twice for the same database but with different pool options will return the same database instance both times. This means that the session pool options supplied in the second call will not be respected.

#### Environment details

  - OS: Ubuntu 18.04
  - Node.js version: 12.13.1
  - npm version: 6.12.1
  - `@google-cloud/spanner` version: 4.4.0

#### Steps to reproduce

```Typescript
const dbWithDefaultOptions = instance.database('some-database');
const dbWithWriteSessions = instance.database('some-database', {
  writes: 1.0,
});
// The following assert should succeed, but currently fails.
assert.notStrictEqual(dbWithDefaultOptions, dbWithWriteSessions);
```

olavloite:
`Instance.database(name, poolOptions)` should return different databases when called multiple times with the same name but with different pool options. Otherwise, the `SessionPoolOptions` supplied in the second and subsequent calls are not respected, and instead the options of the first call are used.

Fixes #753

### DIFF
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -354,7 +354,7 @@ class Instance extends common.GrpcServiceObject {
     // Only add an additional key for SessionPoolOptions if an options object with at least one value was passed in.
     const optionsKey =
       poolOptions && Object.keys(poolOptions).length > 0
-        ? '/' + JSON.stringify(poolOptions)
+        ? '/' + JSON.stringify(Object.entries(poolOptions).sort())
         : '';
     const key = name.split('/').pop() + optionsKey;
     if (!this.databases_.has(key!)) {
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -351,7 +351,12 @@ class Instance extends common.GrpcServiceObject {
     if (!name) {
       throw new Error('A name is required to access a Database object.');
     }
-    const key = name.split('/').pop();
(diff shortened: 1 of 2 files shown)


src/instance.ts
```diff
@@ -351,7 +351,12 @@ class Instance extends common.GrpcServiceObject {
     if (!name) {
       throw new Error('A name is required to access a Database object.');
     }
-    const key = name.split('/').pop();
+    // Only add an additional key for SessionPoolOptions if an options object with at least one value was passed in.
+    const optionsKey =
+      poolOptions && Object.keys(poolOptions).length > 0
+        ? '/' + JSON.stringify(poolOptions)
```
hengfengli: Would the order of poolOptions affect the optionsKey? E.g., 

```javascript
{
   max: 10,
   min: 5,
}
```

```javascript
{
   min: 5,
   max: 10,
}
```

Will they lead to different optionsKey?

olavloite: That's a good point. Yes, that would lead to different option keys. I've changed it to a version that ignores the order of the properties.

PR #760, opened 2019-12-17
feat: add fieldNames option in instance#getMetadata()

AVaksman:

Enable user to pass a list of `Instance` field names to be returned via `getMetadata()` method
[proto source](https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/instance/v1/spanner_instance_admin.proto#L472)


src/instance.ts
```diff
@@ -697,6 +715,9 @@ class Instance extends common.GrpcServiceObject {
    * @see {@link v1.InstanceAdminClient#getInstance}
    * @see [GetInstance API Documentation](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.instance.v1#google.spanner.admin.instance.v1.InstanceAdmin.GetInstance)
    *
+   * @param {GetInstanceMetadataOptions} [options] Configuration object
+   * @param {string | string[]} [options.fieldNames] A list of `Instance` field
+   *     names to be requested.
```
laljikanjareeya: it's good to have example of `fieldNames`.

test/instance.ts
```diff
@@ -498,6 +532,34 @@ describe('Instance', () => {
...
+          fieldMask: {
+            paths: fieldNames,
+          },
+          name: instance.formattedName_,
+        });
+        done();
+      };
+      instance.get({fieldNames}, assert.ifError);
+    });
+
```
laljikanjareeya: can we add unit test case for nested type as mention in below docs:
https://github.com/googleapis/nodejs-spanner/blob/1c0cf27b6c502fe7cc03f9b4c9191d746761b8b6/protos/google/protobuf/field_mask.proto#L42-L174

AVaksman: I believe there is no case of nested types for `FieldMask` in `getMetadata()` for `Instance`
Available values for `fieldNames` array are:
`'name'`, `'displayName'`, `'endpointUris'`, `'labels'`, `'config'`, `'nodeCount'`, `'state'`

src/instance.ts
AVaksman: Added some samples as well as a list of eligible values that could be passed as `fieldNames`

AVaksman:
While we at it, how about utilize `FieldName` functionality with `Instance#exists()` method.
Perhaps pass `name` into `getMetadata` inside the `exists` [method](https://github.com/AVaksman/nodejs-spanner/blob/instance_getMeta_fieldMaks/src/instance.ts#L477) as optimization effort?

src/instance.ts
```diff
@@ -83,7 +86,9 @@ export type SetInstanceMetadataCallback = ResourceCallback<
   GaxOperation,
   IOperation
 >;
-export interface GetInstanceConfig extends GetConfig {}
+export interface GetInstanceConfig
+  extends GetConfig,
+    GetInstanceMetadataOptions {}
```
hengfengli: nits: can we put `GetInstanceMetadataOptions {}` into previous line? like: 

```typescript
export interface GetInstanceConfig
  extends GetConfig, GetInstanceMetadataOptions {}
```

src/instance.ts
```diff
@@ -542,7 +550,13 @@ class Instance extends common.GrpcServiceObject {
         ? optionsOrCallback
         : ({} as GetInstanceConfig);
 
-    this.getMetadata((err, metadata) => {
+    const getMetadataOptions: GetInstanceMetadataOptions = new Object(null);
+    if (options.fieldNames) {
+      getMetadataOptions.fieldNames = options[`fieldNames`];
```
hengfengli: nits: use single quote? `options['fieldNames']`.

src/instance.ts
AVaksman: I tried, but unfortunately linter is complaining about it.
`npm run fix` breaks it back in two lines.

src/instance.ts
AVaksman: done, thanks

PR #775, opened 2019-12-28
fix: do not use empty string as emulator host

olavloite:
Check whether the `SPANNER_EMULATOR_HOST` environment variable contains an empty string, and if it does, do not use it.


AVaksman:
isn't empty string is a "falsy" value?
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

olavloite:
@AVaksman Thanks, you're completely right, so this is not necessary.

PR #776, opened 2019-12-29
fix: session pool should use push/pop and return sessions lifo

### LINKED ISSUE #767: Using unshift() & splice() for stack in the session pool leads to a poor performance.
#### Environment details

  - OS:macOS 10.14.6
  - Node.js version:v13.3.0
  - npm version:6.13.2
  - `@google-cloud/spanner` version:4.4.1

#### Steps to reproduce

`this._inventory[type]` returns an array. Currently, we use `unshift` to prepend a session to the front of the array. I believe this is very low efficient because every time you have to shift all elements in order to prepend an element into an array (a O(n) operation). 

https://github.com/googleapis/nodejs-spanner/blob/f419e27e24eb8df78633d8245c364a709e58d007/src/session-pool.ts#L867

When we remove it from the array, we use `splice` to modify the original array. If the element is the first one, we also need to move all the rest of elements (a O(n) operation). 

https://github.com/googleapis/nodejs-spanner/blob/f419e27e24eb8df78633d8245c364a709e58d007/src/session-pool.ts#L522-L528

https://github.com/googleapis/nodejs-spanner/blob/f419e27e24eb8df78633d8245c364a709e58d007/src/session-pool.ts#L538-L542

If we use the array as a LIFO stack, we should use push & pop (O(1) operations). 

#### Benchmarking test

```typescript
var array = [];
const runs = 100000;

var start = Date.now();
for (var _i = 0; _i < runs; _i++) {
    array.push(_i);
}
console.log("push time:", Date.now() - start);

start = Date.now();
for (var _i = 0; _i < runs; _i++) {
    array.pop();
}
console.log("pop time:", Date.now() - start);

array = [];
start = Date.now();
for (var _i = 0; _i < runs; _i++) {
    array.unshift(_i);
}
console.log("unshift time:", Date.now() - start);

var index;
start = Date.now();
for (var _i = 0; _i < runs; _i++) {
    index = array.indexOf(_i);
    array.splice(_i, 1);
}
console.log("lookup & splice time:", Date.now() - start);
```

Output: 

```bash
push time: 5
pop time: 4
unshift time: 856
lookup & splice time: 2960
```

olavloite:
Use push/pop for returning and taking sessions from the pool. This also makes the session pool return sessions in LIFO order, which has been shown to perform better in low load situations (i.e. 1 transaction per second).

Fixes #767


src/session-pool.ts
```diff
@@ -532,17 +532,22 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
    *
    * @param {string} type The desired session type.
    * @return {Session}
    */
   _borrowFrom(type: types): Session {
-    const session = this._inventory[type][0];
-    this._borrow(session);
-    return session;
+    const session = this._inventory[type].pop();
+    if (session) {
```
AVaksman: It seems that `borrowFrom` is only called by `_borrowNextAvailableSession` which in turn is only called `_getSession` in cases ether `this.available` is > 0 or on `'available'` event, otherwise a `new Error(errors.Closed)` is thrown.
If so, this step might not be necessary.

WDYT?

src/session-pool.ts
```diff
@@ -532,17 +532,22 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
    * @private
    *
    * @param {string} type The desired session type.
    * @return {Session}
    */
   _borrowFrom(type: types): Session {
-    const session = this._inventory[type][0];
-    this._borrow(session);
-    return session;
+    const session = this._inventory[type].pop();
```
AVaksman: to switch from FIFO to LIFO perhaps we can switch 
from 
```ts
const session = this._inventory[type][0];
```
to 
```ts
const session = this._inventory[type][this._inventory.length - 1];
```
Once [`splice`](https://github.com/googleapis/nodejs-spanner/blob/a1a1ff49ed9036acf0876ae4967ad7f7be2dac27/src/session-pool.ts#L531) gets the last index instead of the first it's efficiency should be same as `pop()`

src/session-pool.ts
olavloite: Yes, you're right about that, but the compiler is not able to see that. So without this check you will get a compiler error. I've added a comment to indicate this.

src/session-pool.ts
olavloite: Yes, you're right about that, but in my opinion, using `push` and `pop` better reflects the LIFO behavior of the session pool that we want. This makes it easier for new developers to automatically understand what the behavior is. It also keeps it more in line with the other client libraries.

src/session-pool.ts
AVaksman: Yes, the come tricky with the type safety monitoring.
I believe the compiler is complaining due to the `undefined` part of the `pop()` typescript return signature of `T | undefined`.
A `!` should be able to take care of that
```ts
    const session = this._inventory[type].pop();

    this._inventory.borrowed.add(session!);
    return session!;
```

olavloite: Ah, yes, that's a lot better. Done.

AVaksman:
LGTM

Issue #779, opened 2019-12-31
add support for spanner client resource base routing 


laljikanjareeya:

skuruppu:
We have decided to implement this functionality on the server side so we no longer need to add this support on the client side.

PR #780, opened 2019-12-31
feat: add support for spanner client resource base routing

laljikanjareeya:
Implement resource based routing.

When preparing Gapic Request for a `SpannerClient`, If Client code does not specify the `apiEndpoint` than check for suggested `endpointUris`.

If `endpointUris` are returned, use the first one as `apiEndpoint`.
In case of no `endpointUris` are returned, use default `apiEndpoint`.

Gapic Clients are cached per instance on the `SpannerClient`.

This feature is disabled by default and is enabled by setting the environment variable to true.

Environment variable to enable this feature:
`GOOGLE_CLOUD_SPANNER_ENABLE_RESOURCE_BASED_ROUTING`

Closes : #779


src/index.ts
```diff
@@ -771,20 +775,93 @@ class Spanner extends GrpcService {
...
+      callback(error);
+      return;
+    }
+    const instanceId = config.instanceId;
+    clientName = `${clientName}-${instanceId}`;
+
+    if (!this.clients_.has(clientName)) {
+      this.instances_
+        .get(instanceId)!
+        .getInstanceEndPointUris((err, endPointUris) => {
```
AVaksman: Please add a comment above this line (to remember to change before merging)
That once #760 PR is merged this will get replaced with 
```ts
getMetadata({fieldNames: 'displayName'}, (err, metadata, apiResponse)) => {...}
```

AVaksman: `endPointUris` => `endpointUris` here and down below please

src/instance.ts
```diff
@@ -737,6 +738,57 @@ class Instance extends common.GrpcServiceObject {
...
+   * instance.getInstanceEndPointUris(function(err, endPointUris) {});
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * instance.getInstanceEndPointUris().then(function(data) {
+   *   const endPointUris = data[0];
+   * });
+   */
+  getInstanceEndPointUris(
```
AVaksman: Just leaving a note here to remember to remove this method once #760 is merged

src/instance.ts
```diff
@@ -92,6 +92,7 @@ interface InstanceRequest {
   ): void;
   <T>(config: RequestConfig, callback: RequestCallback<T>): void;
 }
+export type GetInstanceEndPointUrisCallback = NormalCallback<string[]>;
```
AVaksman: Same note to remember to remove once #760 is merged

src/index.ts
laljikanjareeya: Done!!

src/index.ts
```diff
@@ -771,20 +775,96 @@ class Spanner extends GrpcService {
...
+          reqOpts,
+          config.gaxOpts
+        );
+        callback(null, requestFn);
+      });
+    });
+  }
+
+  /**
+   * get GAX client. This will retrive the end point uris for specific instance and cache the client accordingly.
```
skuruppu:
```suggestion
   * get GAX client. This will retrieve the end point uris for specific instance and cache the client accordingly.
```

src/index.ts
```diff
@@ -771,20 +775,96 @@ class Spanner extends GrpcService {
...
+    if (!this.clients_.has(clientName)) {
+      this.instances_.get(instanceId)!.getMetadata(
+        {
+          fieldNames: ['endpointUris'],
+        },
+        (err, instance) => {
+          if (err) {
+            if (err.code === 7) {
+              process.emitWarning(
+                'spanner.instances.get permission must be added to use resource based routing.'
```
skuruppu: Sorry we didn't quite update the spec on time but the warning message should be the following:

"The client library attempted to connect to an endpoint closer to your Cloud Spanner data but was unable to do so. The client library will fallback to the API endpoint given in the client options, which may result in increased latency. We recommend including the scope https://www.googleapis.com/auth/spanner.admin so that the client library can get an instance-specific endpoint and efficiently route requests."

src/index.ts
```diff
@@ -771,20 +775,96 @@ class Spanner extends GrpcService {
...
+    clientName = `${clientName}-${instanceId}`;
+
+    if (!this.clients_.has(clientName)) {
+      this.instances_.get(instanceId)!.getMetadata(
+        {
+          fieldNames: ['endpointUris'],
+        },
+        (err, instance) => {
+          if (err) {
+            if (err.code === 7) {
```
skuruppu: I'm a bit worried about checking for code 7 since it's not clear what this is. If it's possible to be more explicit about checking for the PERMISSION_DENIED error, that would be great. If not, add a comment to say that this is what's being checked here.

src/index.ts
```diff
@@ -771,20 +775,96 @@ class Spanner extends GrpcService {
...
+        }
       );
-      callback(null, requestFn);
-    });
+    } else {
+      callback(null, this.clients_.get(clientName)!);
+    }
+  }
+  /**
+   * cache the GAX client accordingly.
```
skuruppu:
```suggestion
   * Cache the GAX client accordingly.
```

src/index.ts
```diff
@@ -771,20 +775,96 @@ class Spanner extends GrpcService {
...
+   *
+   * @private
+   *
+   * @param {string} clientName client name to cache.
```
skuruppu:
```suggestion
   * @param {string} clientName Client name to cache.
```

test/index.ts
```diff
@@ -920,6 +922,179 @@ describe('Spanner', () => {
...
+      };
+      asAny(CONFIG).instanceId = instanceId;
+
+      spanner.prepareGapicRequest_(CONFIG, err => {
+        assert.strictEqual(err, error);
+        done();
+      });
+    });
+
+    it('should continue if do not have permission for instance', done => {
```
skuruppu:
```suggestion
    it('should continue if a permission error is returned for the instance', done => {
```

test/index.ts
```diff
@@ -920,6 +922,179 @@ describe('Spanner', () => {
...
+      spanner.prepareGapicRequest_(CONFIG, err => {
+        assert.ifError(err);
+        assert.ok(
+          (process.emitWarning as sinon.SinonStub).calledWith(expectedWarning)
+        );
+        done();
+      });
+    });
+
+    it('should return an error if instanceId does not provided.', done => {
```
skuruppu:
```suggestion
    it('should return an error if instanceId is not provided.', done => {
```

test/index.ts
```diff
@@ -920,6 +922,179 @@ describe('Spanner', () => {
...
+          done();
+          return FAKE_GAPIC_CLIENT;
+        }
+      };
+      spanner.prepareGapicRequest_(CONFIG, assert.ifError);
+    });
+
+    it('should use user-specified endpoint when GetInstance response is empty.', done => {
+      const instanceId = 'instance-id';
+      const customeEndpoint = 'us-central1-spanner.googleapis.com';
```
skuruppu:
```suggestion
      const customEndpoint = 'us-central1-spanner.googleapis.com';
```

test/index.ts
```diff
@@ -920,6 +922,179 @@ describe('Spanner', () => {
...
+            asAny(spanner.options).apiEndpoint
+          );
+          done();
+          return FAKE_GAPIC_CLIENT;
+        }
+      };
+      spanner.prepareGapicRequest_(CONFIG, assert.ifError);
+    });
+
+    it('should override the first endpoint while multiple endpointUri available.', done => {
```
skuruppu:
```suggestion
    it('should override the first endpoint while multiple endpointUris are available.', done => {
```

system-test/spanner.ts
```diff
@@ -4393,6 +4396,37 @@ describe('Spanner', () => {
...
+      spanner.options.enableResourceBasedRouting = true;
+    });
+    afterEach(() => {
+      spanner.options.enableResourceBasedRouting = false;
+    });
+
+    it('should list the session while resource based routing is enabled', done => {
+      database.getSessions((err, session) => {
+        assert.ifError(err);
+        assert.equal(session!.length, 0);
```
skuruppu: Why would the number of sessions be zero?

system-test/spanner.ts
```diff
@@ -4393,6 +4396,37 @@ describe('Spanner', () => {
...
+        done();
+      });
```
skuruppu: In the spec we also wanted a test for user overriding the endpoint in the client options.

User-specified endpoint config

1. Set up the client to point at an instance-specific endpoint
2. Start the client / make a Spanner api call
3. call Spanner api (e.g. ListSessions). Expected to call the configured instance-config endpoint (GetInstance response should be empty, resulting in no override).

src/index.ts
laljikanjareeya: @skuruppu Looks like the link is broken or not working, cloud you look into that.

skuruppu: In https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.instance.v1 this is how the scopes are named (scroll down to GetInstance).

system-test/spanner.ts
```diff
@@ -4393,6 +4396,37 @@ describe('Spanner', () => {
...
+        Name STRING(MAX)
+      ) PRIMARY KEY (Key)
+    `;
+
+    before(async () => {
+      await onPromiseOperationComplete(await database.create());
+      await onPromiseOperationComplete(await table.create(schema));
+    });
+    beforeEach(() => {
+      spanner.options.enableResourceBasedRouting = true;
```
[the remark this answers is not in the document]
laljikanjareeya: yes, apart from this we have disabled the resource based routing.

src/session.ts
```diff
@@ -273,6 +274,7 @@ export class Session extends GrpcServiceObject {
       {
         client: 'SpannerClient',
         method: 'deleteSession',
+        instanceId: ((this.parent.parent as {}) as Instance).id,
```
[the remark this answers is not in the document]
laljikanjareeya: we are just trying to retrieve the `instanceId`. in above case `this` has `session` scope.
`session` has parent `database` which contains parent `instance`.

system-test/spanner.ts
laljikanjareeya: because it's newly created database.

src/index.ts
laljikanjareeya: Done!

src/index.ts
laljikanjareeya: I was looking into `enum` but it's not there for `status`, just created `const` variable name with `PERMISSION_DENIED`.

laljikanjareeya: Done!!

test/index.ts
```diff
@@ -982,11 +982,15 @@ describe('Spanner', () => {
       });
     });
 
-    it('should continue if do not have permission for instance', done => {
+    it('should continue if a permission error is returned for the instance.', done => {
+      const PERMISSION_DENIED_MESSAGE = `The client library attempted to connect to
```
skuruppu: It would be great if there was a way to share this constant across the two index.ts files so that you don't have to repeat this large block of text.

system-test/spanner.ts
skuruppu: Interesting. I thought we initialized the session pool with a min number of sessions when the the client is first initialized. Is this not the case?

skuruppu:
One more request @laljikanjareeya. Please be more descriptive in the PR description. Follow https://github.com/googleapis/google-cloud-python/pull/10183 as an example.

olavloite:
@skuruppu Regarding https://github.com/googleapis/nodejs-spanner/pull/780#discussion_r371551887 on the minimum number of sessions in the pool: The Nodejs client currently still uses `MinSessions=0` as the default value for the session pool.

See https://github.com/googleapis/nodejs-spanner/blob/f9af78a4dffef4d84e5b14a528474e44adb2e48a/src/session-pool.ts#L128

skuruppu:
> @skuruppu Regarding [#780 (comment)](https://github.com/googleapis/nodejs-spanner/pull/780#discussion_r371551887) on the minimum number of sessions in the pool: The Nodejs client currently still uses `MinSessions=0` as the default value for the session pool.
> 
> See
> 
> https://github.com/googleapis/nodejs-spanner/blob/f9af78a4dffef4d84e5b14a528474e44adb2e48a/src/session-pool.ts#L128

Oh got it. Thanks for pointing this out @olavloite.

system-test/spanner.ts
skuruppu: @olavloite explained in a separate comment that the Node.js library initializes the session pool with min=0.

system-test/spanner.ts
laljikanjareeya: @skuruppu Added above system test case.

system-test/spanner.ts
```diff
@@ -4420,12 +4420,50 @@ describe('Spanner', () => {
...
+        (err, instanceMetadata) => {
+          assert.ifError(err);
+          // tslint:disable-next-line: no-any
+          const resolvedEndpoint: string[] = (instanceMetadata as any)
+            .endpointUris;
+          if (resolvedEndpoint.length === 0) {
+            done(); //no resolved endpoint.
+          }
+          //user specific apiEndpoint
+          spanner.options.apiEndpoint = resolvedEndpoint[0];
```
skuruppu: Isn't this using the resolved endpoint instead of a user-specified endpoint?

system-test/spanner.ts
```diff
@@ -4420,12 +4420,50 @@ describe('Spanner', () => {
...
+          spanner.options.enableResourceBasedRouting = true;
```
skuruppu: I think at line 4417, resource routing is already enabled so don't think you need this.

skuruppu:
We have decided to implement this functionality on the server side so we no longer need to add this support on the client side.

PR #781, opened 2019-12-31
feat!: add typings for top level object

### LINKED ISSUE #748: Add typings for top level object
Per #747 I noticed that the top level object has quite a few `any` types using tslint to ignore them. We should add the correct types/overloads there.

olavloite:
Adds input and return types for (public) methods in `index.ts`.

BREAKING CHANGE: `GetInstancesCallback` and `GetInstanceConfigsCallback` now require checks that the `results` callback parameter is defined before it can be used.

Fixes #748


src/index.ts
```diff
@@ -453,15 +506,21 @@ class Spanner extends GrpcService {
       },
       (err, operation, resp) => {
         if (err) {
-          callback(err, null, null, resp);
+          callback!(err, null, null, resp);
```
hengfengli: If `callback` is undefined, I think here it will throw an error. 

```typescript
function createInstance(callback?: Function) {
    callback!("hello, world!");
}

createInstance();
```

This gives me an error: `VM52:3 Uncaught TypeError: callback is not a function`. 

From the [doc](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), I feel like that `!` is used to mute the type checker instead of having any actual effect (stop the execution if the variable is undefined). 

I tried the following code as well: 

```typescript
interface IPerson {
   firstName: string;
   lastName: string;
}

function createInstance(person?: IPerson) {
    console.log(person!.firstName);
}

createInstance();
```

It throws an error: `VM58:6 Uncaught TypeError: Cannot read property 'firstName' of undefined`.

src/index.ts
```diff
@@ -748,12 +828,11 @@ class Spanner extends GrpcService {
    * const spanner = new Spanner();
    * const operation = spanner.operation('operation-name');
    */
-  operation(name) {
+  operation(name): GrpcOperation {
```
hengfengli: Missing a `string` type for `name`.

src/index.ts
AVaksman: the `callback` is optional from the user's point of view. If user omits the callback, a promise will be returned.

From the API's point of view the callback will always be there, it will be provided by `PromissifyAll`

hengfengli: @AVaksman Thanks for explanation.

src/index.ts
olavloite: Done.

system-test/spanner.ts
```diff
@@ -677,7 +677,7 @@ describe('Spanner', () => {
     it('should list the instances', done => {
       spanner.getInstances((err, instances) => {
         assert.ifError(err);
-        assert(instances.length > 0);
+        assert(instances!.length > 0);
```
olavloite: I wonder whether we need to consider this PR as a breaking change, as it does require users to make changes to their code in some cases. The callback for `getInstances` was untyped. It is now typed, and the `instances` field is declared as optional as it is not filled when we return an error. This in turn requires users to change their code to verify that `instances` is actually defined.
@skuruppu and @hengfengli WDYT?

hengfengli: I agree that it should be considered as a breaking change if it requires users to make changes on their existing code. 

> This in turn requires users to change their code to verify that `instances` is actually defined.

When the callback for `getInstances` was untyped, I think users also need to check if `instances` is defined, but this is not forced by the type checker. It is based on the convention about how the callback is defined, `callback(err)` for a failure and `callback(null, results)` for a success. 

Now, due to the typings, it forces users to do this check, otherwise the type checker will give an error. Am I right?

olavloite: > Now, due to the typings, it forces users to do this check, otherwise the type checker will give an error. Am I right?

Correct.

skuruppu: I'm happy with this breaking change if it helps users to make fewer mistakes in the future. @bcoe, what is the process for releasing breaking changes in the client libraries? Is it unacceptable, or is it ok to do so as long as we make it very clear?

callmehiphop: Just an fyi, I think there a major bump was already planned when support for Node 8 is dropped. You may just want to piggyback off of that.

skuruppu:
@olavloite, now that we're able to do a major version bump, we can proceed with this. Could I please ask you to resolve the conflicts? I looked at it but realized that I may not know what I'm doing.

olavloite:
> @olavloite, now that we're able to do a major version bump, we can proceed with this. Could I please ask you to resolve the conflicts? I looked at it but realized that I may not know what I'm doing.

Yes, I'll look into it.

PR #783, opened 2019-12-31
fix: year zero was not accepted for SpannerDate

### LINKED ISSUE #782: Constructing a SpannerDate with year=0 fails
Creating a `SpannerDate` with y/m/d values, where year is 0 (zero) will return an invalid date:

```typescript
// The following will return a 'NaN-NaN-NaN' date.
const date = new codec.SpannerDate(0, 2, 22);
```

olavloite:
Creating a SpannerDate with year/month/date values with year=0 would return an invalid date.

Fixes #782

### DIFF
--- a/src/codec.ts
+++ b/src/codec.ts
@@ -58,8 +58,8 @@ type DateFields = [number, number, number];
  * @extends Date
  *
  * @param {string|number} [date] String representing the date or number
- *     representing the year. If year is a number between 0 and 99 is used, then
- *     year is assumed to be 1900 + year.
+ *     representing the year. If year is a number between 0 and 99, then year is
+ *     is assumed to be 1900 + year.
  * @param {number} [month] Number representing the month (0 = January).
  * @param {number} [date] Number representing the date.
  *
--- a/src/codec.ts
+++ b/src/codec.ts
@@ -59,7 +59,7 @@ type DateFields = [number, number, number];
  *
  * @param {string|number} [date] String representing the date or number
  *     representing the year. If year is a number between 0 and 99, then year is
- *     is assumed to be 1900 + year.
+ *     assumed to be 1900 + year.
  * @param {number} [month] Number representing the month (0 = January).
  * @param {number} [date] Number representing the date.
  *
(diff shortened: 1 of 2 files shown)


test/codec.ts
```diff
@@ -75,6 +75,14 @@ describe('codec', () => {
         assert.strictEqual(json, '1986-03-22');
       });
 
+      it('should accept year zero in y/m/d number values', () => {
+        const d = new codec.SpannerDate(null!);
```
AVaksman: is value planned to be used?

src/codec.ts
```diff
@@ -58,8 +58,9 @@ type DateFields = [number, number, number];
  * @extends Date
  *
  * @param {string|number} [date] String representing the date or number
- *     representing the year.
- * @param {number} [month] Number representing the month.
+ *     representing the year. If year is a number between 0 and 99, then year is
+ *     is assumed to be 1900 + year.
```
AVaksman: `If the year` ... `then the year is`
an extra `is` on line 62

AVaksman:
LGTM

test/codec.ts
```diff
@@ -75,6 +75,14 @@ describe('codec', () => {
...
+        const date = new codec.SpannerDate(0, 2, 22);
```
hengfengli: why is `(0, 2, 22)` instead of `(0, 3, 22)`? did this test pass? `json` should not be equal to `1900-03-22`.

AVaksman: The month are 0 based index ([link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth/))
0 = January

hengfengli: Thanks. Got it. The built-in `Date` uses zero-based index. This is awful, 😭.

test/codec.ts
olavloite: I added the missing assert.

src/codec.ts
olavloite: Fixed.

PR #789, opened 2020-01-05
fix: transaction runner should not timeout before first attempt

### LINKED ISSUE #786: test 'transaction-runner - should throw a DeadlineError if the timeout is exceeded' seems flaky
The test `transaction-runner - should throw a DeadlineError if the timeout is exceeded` seems flaky:

https://source.cloud.google.com/results/invocations/fa1053c0-054c-4f8a-ba0a-a3f5a21d7d74/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-spanner%2Fnode12%2Ftest/log

olavloite:
The transaction runner could timeout before the first transaction attempt had been executed, if a very short timeout value had been set. The transaction runner should always execute at least one transaction attempt before throwing a `DeadlineError`.

Fixes #786.

### DIFF
--- a/src/transaction-runner.ts
+++ b/src/transaction-runner.ts
@@ -186,7 +186,7 @@ export abstract class Runner<T> {
 
     let lastError: ServiceError;
 
-    // The transaction runner should always executed at least one attempt before
+    // The transaction runner should always execute at least one attempt before
     // timing out.
     while (this.attempts === 0 || Date.now() - start < timeout) {
       const transaction = await this.getTransaction();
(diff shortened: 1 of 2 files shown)


src/transaction-runner.ts
```diff
@@ -186,7 +186,9 @@ export abstract class Runner<T> {
 
     let lastError: ServiceError;
 
-    while (Date.now() - start < timeout) {
+    // The transaction runner should always executed at least one attempt before
```
hengfengli: a type `should always execute at...`

olavloite: Fixed.

PR #791, opened 2020-01-07
fix: session pool should only create session if pending<=waiters

### LINKED ISSUE #790: Session pool should not create new sessions when pending < numWaiters
The session pool should not initiate the creation of new sessions if the current number of sessions being created is greater than current number of waiters for a session when an application requests a session. Instead, the requester should be placed in a waiter queue and wait for one of the pending sessions to become available.

The following example will create a session pool with 101 sessions instead of 100, as there is no session available yet when the query is being executed. Instead of waiting for the session pool initialization to return one of the 100 sessions being created, the query will also initiate the creation of a session.

```typescript
const database = instance.database({
  writes: 0.2,
  min: 100,
  max: 200,
});
const [rows] = await database.run('SELECT * FROM FOO');
```

olavloite:
The session pool would always initiate the creation of a new session if one was not available at the moment that an application requested a session, even when a large number of sessions were already pending creation. This could cause the pool to create more sessions than necessary for an application.

Fixes #790.

### DIFF
--- a/src/session-pool.ts
+++ b/src/session-pool.ts
@@ -779,7 +779,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     }
 
     // Only create a new session if there are more waiters than sessions already
-    // being created.
+    // being created. The current requester will be waiter number _numWaiters+1.
     if (
       !this.isFull &&
       this._pending + this._pendingPrepare <= this._numWaiters
--- a/src/session-pool.ts
+++ b/src/session-pool.ts
@@ -235,6 +235,8 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
   _inventory: SessionInventory;
   _onClose!: Promise<void>;
   _pending = 0;
+  _pendingPrepare = 0;
+  _numWaiters = 0;
   _pingHandle!: NodeJS.Timer;
   _requests: PQueue;
   _traces: Map<string, trace.StackFrame[]>;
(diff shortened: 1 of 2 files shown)


src/session-pool.ts
```diff
@@ -772,7 +778,12 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       );
     }
 
-    if (!this.isFull) {
+    // Only create a new session if there are more waiters than sessions already
+    // being created.
+    if (
+      !this.isFull &&
+      this._pending + this._pendingPrepare <= this._numWaiters
```
hengfengli: Shouldn't we change `<=` to `<`? When `this._pending + this._pendingPrepare == this._numWaiters`, should we create a new session?

src/session-pool.ts
```diff
@@ -781,10 +792,13 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     }
 
     try {
+      this._numWaiters++;
```
hengfengli: Should we move this line before the check in Line 783? 

E.g., assume that `sessionPool._pending = 1` and `sessionPool._getSession()` is called, when we reach the check in Line 783, we have a pending session and a waiter (the sessionPool._getSession() thread itself should be a waiter and `_numWaiters` is supposed to be `1`). Then, when `_pending + _pendingPrepare` is equal to `_numWaiters`, we should not create a new session. 

What do you think about this logic?

src/session-pool.ts
olavloite: The current requester is at this point not included in `this._numWaiters`, so if we already have `this._pending === this._numWaiters`, the current requester will not be served by any of the pending sessions. So the check here should be `<=`.

src/session-pool.ts
olavloite: I would prefer to have it at line 795, as that is the moment it actually becomes a waiter (i.e. the moment it actually starts waiting). We don't have to worry about other threads calling `sessionPool._getSession()` between lines 783 and 795 because of the single-threaded nature of the event loop in Node.

hengfengli: At my first look, I would think `_numWaiters` would include the thread itself, but actually not. It's not a big problem. 

It might be good to say something in the comment that for that `if` check, the current thread is not included in the `_numWaiters` so when `this._pending + this._pendingPrepare === this._numWaiters`, we need to create a new session for the current thread.

olavloite: I added a comment about that to clarify it.

PR #792, opened 2020-01-07
fix: fixed part of backoff should be between 1 and 32 seconds + max 1 second randomized jitter

### LINKED ISSUE #653: Spanner: Ensure client backoff is implemented to respect client configuration.
There was an issue in the client library for python where requests to retry could occur without a proper retry object. In this case, rather than defaulting to an exponential backoff, we defaulted to immediate retry. Ensure that for this client library we respect the min, max, and multiplier as specified in the gapic configuration. 

Original Bug: https://github.com/googleapis/google-cloud-python/issues/7303

olavloite:
The Node client did implement a client side backoff if no backoff information was provided by the server, but it did not maximize the fixed part of the backoff at 32 seconds. In addition, each backoff is added a randomized max 1 second jitter.

Fixes #653

### DIFF
--- a/src/transaction-runner.ts
+++ b/src/transaction-runner.ts
@@ -152,7 +152,11 @@ export abstract class Runner<T> {
       return secondsInMs + nanosInMs;
     }
 
-    return Math.pow(2, this.attempts) * 1000 + Math.floor(Math.random() * 1000);
+    // Max backoff should be 32 seconds.
+    return (
+      Math.pow(2, Math.min(this.attempts, 5)) * 1000 +
+      Math.floor(Math.random() * 1000)
+    );
   }
   /**
    * Retrieves a transaction to run against.
--- a/test/transaction-runner.ts
+++ b/test/transaction-runner.ts
@@ -163,7 +163,22 @@ describe('TransactionRunner', () => {
       it('should create a backoff when `retryInfo` is absent', () => {
         const random = Math.random();
 
-        runner.attempts = 5;
+        runner.attempts = 3;
+        sandbox.stub(global.Math, 'random').returns(random);
+
+        const badError = new Error('err') as ServiceError;
+        badError.metadata = new Metadata();
+
+        const expectedDelay = Math.pow(2, 3) * 1000 + Math.floor(random * 1000);
+        const delay = runner.getNextDelay(badError);
+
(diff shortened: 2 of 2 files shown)


src/transaction-runner.ts
```diff
@@ -152,7 +152,11 @@ export abstract class Runner<T> {
       return secondsInMs + nanosInMs;
     }
 
-    return Math.pow(2, this.attempts) * 1000 + Math.floor(Math.random() * 1000);
+    // Max backoff should be 32 seconds.
+    return (
+      Math.pow(2, Math.min(this.attempts, 5)) * 1000 +
```
hengfengli: According to the code, the max is a random value between 32 secs and 33 secs. Not strictly maxing at 32 secs.

olavloite: Good point. I checked the Java client that is [referenced in the issue](https://github.com/googleapis/google-cloud-python/issues/7303#issuecomment-464153384) as the reference, and that client also uses 32 seconds max for the fixed part of the backoff in addition to randomized jitter.
I've updated the title and description of this PR to indicate this.

hengfengli: I didn't see the extra 1-second jitter in java client lib: 

https://github.com/googleapis/java-spanner/blob/c9c059b3b68e5a4362102c7e4d13e6bd015e320c/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L113-L119

But I do see it in python client lib: 

https://github.com/IlyaFaer/google-cloud-python/blob/8d7c503cc5f6e28d4bc1801bca8d5a84e0d8e0f9/spanner/google/cloud/spanner_v1/session.py#L374-L380

Which one should we use as a standard reference?

olavloite: The `ExponentialBackoff` class that is used in Java automatically adds the randomized jitter: https://github.com/googleapis/google-http-java-client/blob/290e3ef6197b56b73256d38f551e3aa044c1c8ff/google-http-client/src/main/java/com/google/api/client/util/ExponentialBackOff.java#L48

hengfengli: It's a different approach to generate the exponential backoff - pick a random value within a range around the retry interval.

PR #795, opened 2020-01-12
fix: retry executeStreamingSql when error code is retryable

### LINKED ISSUE #620: 14 UNAVAILABLE: GOAWAY received
#### Environment details

  - OS: Linux
  - Node.js version: 10
  - yarn version: 1.15
  - `@google-cloud/spanner` version: 3.1.0

#### Steps to reproduce

  1. Run a web service using nodejs-spanner
  2. Observe random 14 UNAVAILABLE: GOAWAY errors

It happens 1-5 times a day

The full stacktrace is 
```
14 UNAVAILABLE: GOAWAY received Error: 14 UNAVAILABLE: GOAWAY received
    at Object.exports.createStatusError (/app/node_modules/grpc/src/common.js:91:15)
    at ClientReadableStream._emitStatusIfDone (/app/node_modules/grpc/src/client.js:233:26)
    at ClientReadableStream._receiveStatus (/app/node_modules/grpc/src/client.js:211:8)
    at Object.onReceiveStatus (/app/node_modules/grpc/src/client_interceptors.js:1272:15)
    at InterceptingListener._callNext (/app/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/app/node_modules/grpc/src/client_interceptors.js:618:8)
    at Object.onReceiveStatus (/app/node_modules/grpc-gcp/build/src/index.js:93:25)
    at InterceptingListener._callNext (/app/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/app/node_modules/grpc/src/client_interceptors.js:618:8)
    at /app/node_modules/grpc/src/client_interceptors.js:1029:24
```

Interestingly, we had a very similar error (`13 INTERNAL: GOAWAY`) when using node-spanner v2.x. We were hoping to see this error disappear by upgrading, which it did, but it apparently got replaced by this one ^^

It might be related to https://github.com/googleapis/nodejs-spanner/issues/234 or to any of these issues https://github.com/grpc/grpc-node/issues?utf8=%E2%9C%93&q=is%3Aissue+GOAWAY+

olavloite:
The streaming call executeStreamingSql is not automatically retried by gax, as the gapic configuration for the call does not specify any error codes that should automatically be retried. Instead, the PartialResultStream is responsible for retrying these calls with the appropriate resume token. Until now, the call was only retried when a valid resume token had been seen for the stream, meaning that if the initial call failed with a retryable error code (e.g. UNAVAILABLE), the
stream would fail with this error. This fix ensures that the call is also retried when the error occurs for the initial call or before the stream has returned a valid resume token.

See also https://github.com/googleapis/nodejs-spanner/blob/09b5c56ed7844514283754235f7203364f01d6ab/test/spanner.ts#L199 for additional tests for retrying streaming errors.

Fixes #620.

### DIFF
--- a/src/partial-result-stream.ts
+++ b/src/partial-result-stream.ts
@@ -398,9 +398,7 @@ export function partialResultStream(
 
   // need types for events-intercept
   // tslint:disable-next-line no-any
-  (requestsStream as any).intercept('error', err => {
-    retry(err);
-  });
+  (requestsStream as any).intercept('error', retry);
 
   return (
     requestsStream
--- a/src/partial-result-stream.ts
+++ b/src/partial-result-stream.ts
@@ -345,7 +345,7 @@ export function partialResultStream(
 ): PartialResultStream {
   const retryableCodes = [status.UNAVAILABLE];
   let lastResumeToken: ResumeToken;
-  let lastErr: ServiceError | undefined;
+  let lastRetriedErr: ServiceError | undefined;
   let lastRequestStream: Readable;
 
   // mergeStream allows multiple streams to be connected into one. This is good;
@@ -362,7 +362,7 @@ export function partialResultStream(
   // This listener ensures that the last request that executed successfully
   // after one or more retries will end the requestsStream.
   const endListener = () => {
-    if (lastErr) {
+    if (lastRetriedErr) {
       setImmediate(() => requestsStream.end());
     }
   };
@@ -385,6 +385,9 @@ export function partialResultStream(
     }
 
     // We're going to retry from where we left off.
+    // Keep track of the fact that we retried an error in order to end the
+    // merged result stream.
+    lastRetriedErr = err;
     // Empty queued rows on the checkpoint stream (will not emit them to user).
     batchAndSplitOnTokenStream.reset();
     makeRequest();
@@ -396,7 +399,6 @@ export function partialResultStream(
   // need types for events-intercept
   // tslint:disable-next-line no-any
   (requestsStream as any).intercept('error', err => {
-    lastErr = err;
     retry(err);
   });
 
--- a/src/partial-result-stream.ts
+++ b/src/partial-result-stream.ts
@@ -25,6 +25,7 @@ import * as streamEvents from 'stream-events';
 
 import {codec, JSONOptions, Json, Field, Value} from './codec';
 import {SpannerClient as s} from './v1';
+import {ServiceError, status} from 'grpc';
 
 export type ResumeToken = string | Uint8Array;
 
@@ -342,7 +343,10 @@ export function partialResultStream(
   requestFn: RequestFunction,
   options?: RowOptions
 ): PartialResultStream {
+  const retryableCodes = [status.UNAVAILABLE];
   let lastResumeToken: ResumeToken;
+  let lastErr: ServiceError | undefined;
+  let lastRequestStream: Readable;
 
   // mergeStream allows multiple streams to be connected into one. This is good;
   // if we need to retry a request and pipe more data to the user's stream.
@@ -355,13 +359,25 @@ export function partialResultStream(
     },
   });
 
+  // This listener ensures that the last request that executed successfully
+  // after one or more retries will end the requestsStream.
+  const endListener = () => {
+    if (lastErr) {
+      setImmediate(() => requestsStream.end());
+    }
+  };
   const makeRequest = (): void => {
-    requestsStream.add(requestFn(lastResumeToken));
+    if (lastRequestStream) {
+      lastRequestStream.removeListener('end', endListener);
+    }
+    lastRequestStream = requestFn(lastResumeToken);
+    lastRequestStream.on('end', endListener);
+    requestsStream.add(lastRequestStream);
   };
 
-  const retry = (err: Error): void => {
-    if (!lastResumeToken) {
-      // We won't retry the request, so this will flush any rows the
+  const retry = (err: ServiceError): void => {
+    if (!(err.code && retryableCodes!.includes(err.code))) {
+      // This is not a retryable error, so this will flush any rows the
       // checkpoint stream has queued. After that, we will destroy the
       // user's stream with the same error.
       setImmediate(() => batchAndSplitOnTokenStream.destroy(err));
@@ -379,7 +395,10 @@ export function partialResultStream(
 
   // need types for events-intercept
   // tslint:disable-next-line no-any
-  (requestsStream as any).intercept('error', retry);
+  (requestsStream as any).intercept('error', err => {
+    lastErr = err;
+    retry(err);
+  });
 
   return (
     requestsStream
--- a/test/partial-result-stream.ts
+++ b/test/partial-result-stream.ts
@@ -275,7 +275,6 @@ describe('PartialResultStream', () => {
             code: status.UNAVAILABLE,
             message: 'Error.',
           } as ServiceError);
-          firstFakeRequestStream.end();
         }, 50);
 
         return firstFakeRequestStream;
@@ -338,7 +337,6 @@ describe('PartialResultStream', () => {
               code: status.UNAVAILABLE,
               message: 'Error.',
             } as ServiceError);
-            firstFakeRequestStream.end();
           }, 50);
         }, 50);
 
@@ -393,7 +391,6 @@ describe('PartialResultStream', () => {
               code: status.DATA_LOSS,
               message: 'Non-retryable error.',
             } as ServiceError);
-            fakeRequestStream.end();
           }, 50);
         }, 50);
 
--- a/test/partial-result-stream.ts
+++ b/test/partial-result-stream.ts
@@ -25,6 +25,8 @@ import * as through from 'through2';
 
 import {codec} from '../src/codec';
 import * as prs from '../src/partial-result-stream';
+import {ServiceError, status} from 'grpc';
+import {Row} from '../src/partial-result-stream';
 
 describe('PartialResultStream', () => {
   const sandbox = sinon.createSandbox();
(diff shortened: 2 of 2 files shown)


src/partial-result-stream.ts
```diff
@@ -355,13 +359,25 @@ export function partialResultStream(
...
+    if (lastErr) {
+      setImmediate(() => requestsStream.end());
+    }
+  };
   const makeRequest = (): void => {
-    requestsStream.add(requestFn(lastResumeToken));
+    if (lastRequestStream) {
+      lastRequestStream.removeListener('end', endListener);
+    }
+    lastRequestStream = requestFn(lastResumeToken);
```
hengfengli: I wonder if a retryable error happens at an initial call, `lastResumeToken` is empty here, so what would `lastRequestStream` be here? In other words, what will `requestFn` return for an empty `lastResumeToken`?

src/partial-result-stream.ts
```diff
@@ -355,13 +359,25 @@ export function partialResultStream(
     },
   });
 
+  // This listener ensures that the last request that executed successfully
+  // after one or more retries will end the requestsStream.
+  const endListener = () => {
+    if (lastErr) {
+      setImmediate(() => requestsStream.end());
```
hengfengli: I don't quite understand this. Why do we need to do this? `endListener` is triggered by an `end` event in `lastRequestStream`, which means this request is successfully executed. Then, we can close the entire `requestsStream` (a merged stream). Why don't we need to close `requestsStream` in the previous code but now we do?

src/partial-result-stream.ts
```diff
@@ -355,13 +359,25 @@ export function partialResultStream(
     },
   });
 
+  // This listener ensures that the last request that executed successfully
+  // after one or more retries will end the requestsStream.
+  const endListener = () => {
+    if (lastErr) {
```
hengfengli: Currently, `lastErr` includes retryable and non-retryable errors. Shouldn't `lastErr` just be retryable errors?

test/partial-result-stream.ts
```diff
@@ -251,7 +253,60 @@ describe('PartialResultStream', () => {
...
+            message: 'Error.',
+          } as ServiceError);
+          firstFakeRequestStream.end();
+        }, 50);
+
+        return firstFakeRequestStream;
+      });
+
+      requestFnStub.onCall(1).callsFake(resumeToken => {
+        assert.ok(!resumeToken, 'Retry called with resume token');
```
hengfengli: ... with `empty` resume token?

src/partial-result-stream.ts
olavloite: This code is also used for the initial call to `requestStreamingSql`, which is also always called without a resume token. Calling `requestFn` another time without a resume token will start a new stream. So:
* The first time this is called (the initial call), `lastRequestStream` is `undefined` at first (i.e. the condition of the `if` statement will be `false`).
* If the initial call fails with a retryable error, this method is called again with `lastRequestStream` still pointing to the initial stream. The `endListener` will then be removed from the initial call and placed on the second call.

src/partial-result-stream.ts
olavloite: Yes, this was the part that took me hours to figure out as well. The basic answer is: The original solution seems to never have worked as intended in the first place.

The problem is that the first request stream that causes an error will never indicate to the merged `requestsStream` that it has ended. This can be illustrated with the following test case (not added to the code, as this test case will actually fail):

```typescript
    it('should end merged stream', done => {
      // This test case will never finish when this flag is set to true.
      const simulateRetryAfterError = true;
      const mergedStream = mergeStream();

      eventsIntercept.patch(mergedStream);
      // tslint:disable-next-line no-any
      (mergedStream as any).intercept('error', err => {
        // By intercepting the error, the underlying input stream that caused
        // the error will never end. That in turn will cause the merged stream
        // also to never end.
        console.log('intercepted error');
      });

      const inputStream1 = new Readable();
      inputStream1._read = (size: number): void => {
        setImmediate(() => {
          inputStream1.emit('data', 'one');
          inputStream1.emit('data', 'two');
          // inputStream1.emit('end'); // Uncomment this line and comment out the next line to make the test case work.
          inputStream1.emit('error', new Error('test error'));
        });
      };
      inputStream1.on('end', () => console.log('Input stream 1 ended'));
      if (simulateRetryAfterError) {
        // Add a stream that ends with an error. That error is then intercepted by the merged
        // stream, which causes the merged stream to think that the underlying
        // stream has never ended, which in turn prevents the mergedStream from
        // finishing.
        mergedStream.add(inputStream1);
      }

      const inputStream2 = new Readable();
      inputStream2._read = (size: number): void => {
        setImmediate(() => {
          inputStream2.emit('data', 'three');
          inputStream2.emit('data', 'four');
          inputStream2.emit('end');
        });
      };
      inputStream2.on('end', () => console.log('Input stream 2 ended'));
      mergedStream.add(inputStream2);
      mergedStream.on('finish', () => {
        console.log('merged stream ended');
        done();
      });
    });
```

What basically happens in the above example and also in the `PartialResultStream` in case of a retry is the following:
1. An initial request stream is opened (`inputStream1`) and added to the merged result stream.
2. `inputStream1` emits a retryable error. This error bubbles up to the merged stream, which intercepts it and executes a retry. Intercepting this error means that the mergedStream does not know that `inputStream1` ended with an error, and it erroneously thinks it should still receive data from it.
3. `inputStream2` is started, returns its data and then finishes. The 'end' event bubbles up to the merged stream, but that does not end the merged stream as it is still waiting for input from `inputStream1`.

The above also happened when testing `PartialResultStream` with the mock server. If you remove this `endListener` that was added, you will see that retrying errors does not actually work as intended: All data is returned to the user stream, but the stream is never ended, causing the test cases to hang.

The above example can be made to work by setting `simulateRetryAfterError = false` or by replacing the error from `inputStream1` with an end event for the input stream.

src/partial-result-stream.ts
olavloite: Yes, that's true. We don't need it for non-retryable errors. Fixed.

alexander-fenster:
Just one question: since

> as the gapic configuration for the call does not specify any error codes that should automatically be retried

should we talk to API team to reconsider this?

src/partial-result-stream.ts
```diff
@@ -379,7 +398,9 @@ export function partialResultStream(
 
   // need types for events-intercept
   // tslint:disable-next-line no-any
-  (requestsStream as any).intercept('error', retry);
+  (requestsStream as any).intercept('error', err => {
```
hengfengli: nits: probably better to keep the original one. No need to change the code here.

src/partial-result-stream.ts
hengfengli: Thanks for the detailed explanation :clap:. This means when an error happens in `inputStream1`, it will emit an `error` event, but there is no `end` event after that. The merged stream is not ended, so it will not emit `finish` event.

test/partial-result-stream.ts
```diff
@@ -251,7 +253,63 @@ describe('PartialResultStream', () => {
...
+      const requestFnStub = sandbox.stub();
+
+      requestFnStub.onCall(0).callsFake(() => {
+        setTimeout(() => {
+          // This causes a new request stream to be created.
+          firstFakeRequestStream.emit('error', {
+            code: status.UNAVAILABLE,
+            message: 'Error.',
+          } as ServiceError);
+          firstFakeRequestStream.end();
```
hengfengli: In order to simulate the real case as we discussed, I think we should remove this. I tried it and it works fine.

test/partial-result-stream.ts
```diff
@@ -309,6 +370,50 @@ describe('PartialResultStream', () => {
...
+          fakeCheckpointStream.emit('checkpoint', RESULT_WITH_TOKEN);
+          fakeRequestStream.push(RESULT_WITH_TOKEN);
+          fakeCheckpointStream.emit('checkpoint', RESULT_WITH_TOKEN);
+
+          setTimeout(() => {
+            fakeRequestStream.emit('error', {
+              code: status.DATA_LOSS,
+              message: 'Non-retryable error.',
+            } as ServiceError);
+            fakeRequestStream.end();
```
hengfengli: Same above. It can be deleted.

test/partial-result-stream.ts
```diff
@@ -276,7 +334,10 @@ describe('PartialResultStream', () => {
 
           setTimeout(() => {
             // This causes a new request stream to be created.
-            firstFakeRequestStream.emit('error', new Error('Error.'));
+            firstFakeRequestStream.emit('error', {
+              code: status.UNAVAILABLE,
+              message: 'Error.',
+            } as ServiceError);
             firstFakeRequestStream.end();
```
hengfengli: Same above.

src/partial-result-stream.ts
hengfengli: FYI, when I changed your test case from `finish` to `end`, like this: 

```typescript
      mergedStream.on('end', () => {
        console.log('merged stream ended');
        done();
      });
```

I commented the error and uncommented the `end`: 

```typescript
          ...
          inputStream1.emit('end'); // Uncomment this line and comment out the next line to make the test case work.
          // inputStream1.emit('error', new Error('test error'));
          ...
```

This does not work any more. I was wondering why this happens, :sweat_smile:. Then, I found this link: https://nodejs.org/api/stream.html#stream_events_finish_and_end.

olavloite:
> Just one question: since
> 
> > as the gapic configuration for the call does not specify any error codes that should automatically be retried
> 
> should we talk to API team to reconsider this?

That is certainly an option, but it is something that we in that case should do in sync with all client libraries. I know that the Java and Go client libraries also have custom implementations that retry UNAVAILABLE errors for the executeStreamingSql RPC, and I assume that it is also the case of other languages.

test/partial-result-stream.ts
olavloite: `end()` call removed from test case.

In case of a non-retryable error, we circumvent the `requestsStream` and send an error directly to the stream that is returned to the user (technically, by sending it first to the `batchAndSplitOnTokenStream`). So it is not necessary to end the `requestsStream` in that case.

See https://github.com/googleapis/nodejs-spanner/blob/569570fe3ed7a8ebc10b0ed2d528435102be5e82/src/partial-result-stream.ts#L383

src/partial-result-stream.ts
olavloite: Done.

test/partial-result-stream.ts
olavloite: Done

test/partial-result-stream.ts
olavloite: Done.

src/partial-result-stream.ts
olavloite: > Thanks for the detailed explanation . This means when an error happens in `inputStream1`, it will emit an `error` event, but there is no `end` event after that. The merged stream is not ended, so it will not emit `finish` event.

Correct. And the error that is emitted by `inputStream1` is intercepted, so that also keeps the merged stream think that `inputStream1` is still alive.

PR #797, opened 2020-01-13
types: add types for createInstance

AVaksman:

Towards #748


skuruppu:
@AVaksman if you could check the build failures, that would be great.

src/index.ts
```diff
@@ -434,15 +465,15 @@ class Spanner extends GrpcService {
         {
           name: formattedName,
           displayName: shortName,
+          nodeCount: config.nodes || 1,
```
stephenplusplus: Is it possible the user would want `0`?

AVaksman: [Docs](https://cloud.google.com/spanner/docs/instances#node_count)
`0` is not a valid value. The request with `node_count` as `0` will be rejected with 
`INVALID_ARGUMENT: Invalid CreateInstance request.` error.

AVaksman:
It seems to be a bug with coverage report.
closing this PR in favor of #805 which has the same exact changes

PR #798, opened 2020-01-13
feat: expose displayName in createInstance

AVaksman:
This PR is dependent on #797 
 
### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -99,6 +99,7 @@ export interface RequestConfig {
 export interface CreateInstanceRequest {
   config: string;
   nodes?: number;
+  displayName?: string;
   labels?: {[k: string]: string} | null;
 }
 /**
@@ -324,6 +325,11 @@ class Spanner extends GrpcService {
    *     be used to control how resource metrics are aggregated. And they can
    *     be used as arguments to policy management rules (e.g. route,
    *     firewall, load balancing, etc.).
+   * @property {string} [displayName] The descriptive name for this instance
+   *     as it appears in UIs. Must be unique per project and between 4 and 30
+   *     characters in length.
+   *     Defaults to the instance unique identifier '<instance>' of the full
+   *     instance name of the form 'projects/<project>/instances/<instance>'.
    */
   /**
    * @typedef {array} CreateInstanceResponse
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -426,23 +463,23 @@ class Spanner extends GrpcService {
...
     const reqOpts = {
       parent: 'projects/' + this.projectId,
-      instanceId: shortName,
+      instanceId: displayName,
       instance: extend(
         {
           name: formattedName,
-          displayName: shortName,
+          displayName,
+          nodeCount: config.nodes || 1,
```
olavloite: As far as I can tell, the `createInstance` method used to accept the node count to be specified both using the property `nodeCount` as well as `nodes`. If both were set, `nodes` would take precedence over `nodeCount`. This implementation seems to only accept `nodes`, which could break existing client code. Or am I missing something here?

src/index.ts
```diff
@@ -65,6 +70,12 @@ export interface RequestConfig {
   reqOpts: any;
   gaxOpts?: {};
 }
+export interface CreateInstanceRequest {
```
olavloite: Wouldn't it be better to let this interface extend the `IIinstance` interface, as they share almost all properties? This request only adds the `nodes` property.

AVaksman: `IIinstance` interface provides misleading as well as unnecessary  properties for the purpose of `CreateInstance` function. I believe it is an error for `CreateInstanceRequest` interface to extend from `IIinstance` interface due to the following:

- `name` should not be part of the `config` param, it is passed in separately.
- `config` property is not optional.
- `nodeCount` should not be part of the `config` param. `nodeCount` is not sighted anywhere in the documentation and only erroneously sneaked in during typescript conversion when `CreateInstanceRequest` interface was set to extend `IIinstance` interface.
- `state` can not be set during instance creation, it is output only. 
- `endpointUris` can not be set during instance creation, it is output only.

src/index.ts
AVaksman: As per my comment above, I believe `nodeCount` became a part of `config` by an error when `CreateInstanceRequest` interface was set to extend from `IIinstance` interface. `nodeCount` is not listed in the documentation (which predated the introduction of types), so it is not legacy code. I believe it is confusing to have both `nodes` and `nodeCount` properties.

src/index.ts
AVaksman: Once we set default value for `nodes`, as introduced in the PR, `nodes` parameter becomes optional.

src/index.ts
olavloite: I agree that having both is confusing (and unnecessary), and if it was only recently introduced by the addition of types, then I agree that we should remove it. We should however flag it in the release notes to notify potential users that have used the property in the meantime.

src/index.ts
olavloite: Sounds reasonable, thanks for the elaborate explanation.

src/index.ts
skuruppu: Resolving this since we submitted the nodeCount change in the breaking changes PR.

PR #801, opened 2020-01-15
test: add mock server for testing

olavloite:
Adds a mock Spanner server for testing purposes. The mock server implements the gRPC interface of the Spanner backend and implements simple in-mem functionality for creating sessions and transactions. The results of query and update statements must be mocked in advance before they can be executed on the server. The server does not implement any storage, and executing an update statement on the server does not change the state of the server.

The purpose of the server is to allow simple in-mem (integration) testing for specific use cases and error conditions.


protos/google/type/expr.proto
```diff
@@ -0,0 +1,51 @@
...
+    string title = 2;
+
+    // An optional description of the expression. This is a longer text which
+    // describes the expression, e.g. when hovered over it in a UI.
+    string description = 3;
+
+    // An optional string indicating the location of the expression for error
+    // reporting, e.g. a file name and a position in the file.
+    string location = 4;
+}
```
wyardley: nit: missing line endings on some of the files

skuruppu:
> Interesting idea, wonder if it makes sense to make this usable by consumers of the library (for tests). I assume the Spanner emulator would eventually be preferred for integration testing, once it’s available?

That's actually a really good point @wyardley. @olavloite, would you be able to list the advantages of using this over the emulator?

olavloite:
@skuruppu @wyardley 

This mock server and the emulator have slightly different use cases.

### Emulator
* The emulator is good for running integration tests that might span multiple queries, updates and transactions, and will keep track of the actual state at any time.
* The emulator is good for quickly checking the validity of a statement or the validity of a certain operation on Cloud Spanner.
* The emulator is a separate binary that must be started in addition to your own test code.

### Mock server
* The mock server is good for (unit) testing specific return values or specific (error) conditions that are difficult to trigger on Cloud Spanner or an emulator, such as for example an aborted transaction, an `UNAVAILABLE` error response or a broken result set stream.
* The mock server is quick and simple to use to unit test basic scenarios such as simple queries and small transactions where you know what the response of a query/update should be (or don't care for the specific test case).
* The mock server cannot be used to check the validity of a query or statement.
* The mock server can be started in-process with the test code.

### General
* Use the emulator to verify that you are using the Cloud Spanner backend in the correct way.
* Use the mock server to verify that you are using the Cloud Spanner client library in the correct way.

protos/google/type/expr.proto
olavloite: Done.

skuruppu:
Thanks for the explanation @olavloite, I think it's reasonable to have a mock server for certain types of tests as you mentioned.

protos/google/api/client.proto
```diff
@@ -0,0 +1,100 @@
+// Copyright 2019 Google LLC.
```
mayurkale22: minor: `s/2019/2020`

protos/google/api/field_behavior.proto
```diff
@@ -0,0 +1,79 @@
...
+    //   google.protobuf.Duration ttl = 1
+    //     [(google.api.field_behavior) = INPUT_ONLY];
+    //   google.protobuf.Timestamp expire_time = 1
+    //     [(google.api.field_behavior) = OUTPUT_ONLY,
+    //      (google.api.field_behavior) = IMMUTABLE];
+    repeated google.api.FieldBehavior field_behavior = 1052;
+}
+
+// An indicator of the behavior of a given field (for example, that a field
+// is required in requests, or given as output but ignored as input).
```
mayurkale22:
```suggestion
// is required in requests, or given as output but ignored as an input).
```

test/mockserver/mockdatabaseadmin.ts
```diff
@@ -0,0 +1,205 @@
...
+  longs: String,
+  enums: String,
+  defaults: true,
+  oneofs: true,
+  includeDirs: [IMPORT_PATH, PROTO_DIR],
+});
+const protoDescriptor = grpc.loadPackageDefinition(packageDefinition);
+const databaseAdminProtoDescriptor =
+  protoDescriptor['google']['spanner']['admin']['database']['v1'];
+export const TEST_DATABASE_NAME =
```
mayurkale22: I think it is safe to remove `export` from this and next `const`. WDYT?

test/mockserver/mockspanner.ts
```diff
@@ -0,0 +1,730 @@
...
+  oneofs: true,
+  includeDirs: [IMPORT_PATH, PROTO_DIR],
+});
+const protoDescriptor = grpc.loadPackageDefinition(packageDefinition);
+const spannerProtoDescriptor = protoDescriptor['google']['spanner']['v1'];
+const jsonProtos = require('../../../protos/protos.json');
+
+const RETRY_INFO = 'google.rpc.retryinfo-bin';
+// // tslint:disable-next-line variable-name
+// const RetryInfo = Root.fromJSON(jsonProtos).lookup('google.rpc.RetryInfo');
```
mayurkale22: I am not sure if we want to merge comment out code.

test/mockserver/mockspanner.ts
```diff
@@ -0,0 +1,730 @@
...
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+
+  private async simulateExecutionTime(functionName: string): Promise<void> {
+    while (this.frozen > 0) {
+      await MockSpanner.sleep(10);
+    }
+    const execTime = this.executionTimes.get(functionName);
+    if (execTime) {
+      if (
```
mayurkale22: What do you think about merging this `if` block with previous one? Something like:

```js
if (execTime && execTime.errors && execTime.errors.length && !execTime.errors[0].streamIndex) {
 // ...
}
```

test/spanner.ts
```diff
@@ -0,0 +1,1209 @@
...
+        sql: selectSql,
+      };
+      const database = newTestDatabase();
+      try {
+        const [rows] = await database.run(query);
+        assert.strictEqual(rows.length, 3);
+        let i = 0;
+        rows.forEach(row => {
+          i++;
+          assert.strictEqual(row[0].name, 'NUM');
```
mayurkale22: optional: you can destructure instead of calling by index

protos/google/api/client.proto
olavloite: The proto files are direct copy-paste from the googleapis GitHub project. I'd rather not make any changes to them.

protos/google/api/field_behavior.proto
olavloite: Same as above.

test/mockserver/mockdatabaseadmin.ts
olavloite: Done

test/mockserver/mockspanner.ts
olavloite: Removed :-)

test/mockserver/mockspanner.ts
olavloite: I agree, that is better.

test/spanner.ts
olavloite: Thanks for pointing that out (I'm quite new to Typescript / JavaScript). I tried it, but it turns out that the declared return type of `database.run(...)` is incorrect. It claims to return a `Promise<Row[]>`, but in reality it returns a `Promise<RowResponse>`, where `RowResponse` is a `[Row[], ResultSetStats]`. That prevents me from using destructuring here, as the typing is incorrect. I'll open a separate issue for this and look into the destructuring later for this part of the code.

PR #805, opened 2020-01-21
fix(types)!: types for createInstance

AVaksman:

Towards #748


src/index.ts
```diff
@@ -341,25 +351,42 @@ class Spanner extends GrpcService {
...
+   *     Cloud Labels can be used to filter collections of resources. They can
+   *     be used to control how resource metrics are aggregated. And they can
+   *     be used as arguments to policy management rules (e.g. route,
+   *     firewall, load balancing, etc.).
    */
   /**
    * @typedef {array} CreateInstanceResponse
    * @property {Instance} 0 The new {@link Instance}.
-   * @property {Operation} 1 An {@link Operation} object that can be used to check
+   * @property {Operation} 1 An operation object that can be used to check
```
skuruppu: Why are the links removed here and below?

src/index.ts
```diff
@@ -434,15 +465,15 @@ class Spanner extends GrpcService {
         {
           name: formattedName,
           displayName: shortName,
+          nodeCount: config.nodes || 1,
```
olavloite: As also mentioned [here](https://github.com/googleapis/nodejs-spanner/pull/798#discussion_r369383379): I think we should flag in the release notes that users that have been using `config.nodeCount` must switch to using `config.nodes`. As `config.nodeCount` was only introduced by the adding of types relatively recently, and it is not mentioned in the documentation, I think it's fair to assume that it should not affect many users.

src/index.ts
AVaksman: `Operation` is not defined, link it does not provide anything during docs generation.

src/index.ts
AVaksman: To summarize,

- `nodeCount` was never sighted in any docs. 
- It erroneously became a part of the `config` when types where [introduced](https://github.com/googleapis/nodejs-spanner/blob/da8c2f85b55346d99a9eacc4249e6d6cd1a14556/src/instance.ts#L63) and `CreateInstanceRequest` interface extended `IInstance` interface
- This PR removes `nodeCount` and some other misleading params (`name`, `state`, `endpointUris`) from the `CreateInstance` `config` interface.
Bottom line, if anyone used `nodeCount` option, this change will break their code.

@bcoe what should be the proper course of action here?

skuruppu: From our meeting notes:

On Node meeting @bcoe suggested to lump this together with other breaking changes in the pipeline anyways. Microgenerators with types are coming.

bcoe:
@alexander-fenster @xiaozhenliu-gg5 bringing this back to the top of our minds, might be good to prioritize this library for TypeScript conversion since it seems like it might be a bit finicky.

alexander-fenster:
@bcoe It's in progress, I'm half way there.

JustinBeckwith:
Have we determined this is a breaking change then?  If so, let's please use the `fix!: <stuff>` format to make sure we don't accidentally let a breaking change slip in as a patch fix :)

alexander-fenster:
@JustinBeckwith It's not supposed to be a breaking change. The code just works, I have problems with unit tests (basically, either making `proxyquire` work in the new setting, or get rid of it). Need some more time on it but slowly progressing :)

olavloite:
Regarding the different questions about whether this should be a breaking change: I think there is some confusion between two different PRs. This PR (#805) is not considered a breaking change. The PR that adds typings for index.ts (#781) is considered a breaking change. That is also the PR that was suggested that could be combined with other breaking changes.

skuruppu:
@AVaksman, please wait for #781 to be submitted and resolve the conflicts afterwards since both PRs are changing some of same parts of the code. After the conflicts are resolved, I don't think this will be a breaking change. But correct me if I'm wrong.

skuruppu:
#781 is now merged so after you resolve the conflicts, I'll be happy to do a review.

PR #807, opened 2020-01-22
fix: use PreciseDate instead of Date for min read timestamp

agasheesh:
Without this fix, TimestampBounds defaults min read timestamp to empty google timestamp proto, which correspond to min unix time i.e., 1970:01:01-T:00:00:00Z.


system-test/spanner.ts
```diff
@@ -3457,7 +3458,7 @@ describe('Spanner', () => {
         const query = 'SELECT * FROM TxnTable';
 
         const options = {
-          minReadTimestamp: new Date(),
+          minReadTimestamp: new PreciseDate(),
```
skuruppu: I couldn't really tell from the [docs](https://github.com/googleapis/nodejs-precise-date) what PreciseDate() defaults to for the empty ctor. Is it some value or is it just an empty object?

agasheesh: PreciseDate() will call ctor for super() i.e., Date() which returns Now() timestamp. TimestampBounds object however expects an object of PreciseDate and will convert it correctly to a string representation.

agasheesh: https://screenshot.googleplex.com/X54iBrh6WFQ

agasheesh: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

skuruppu: Ah got it, thanks for the pointer @agasheesh.

PR #810, opened 2020-01-22
fix: fixed wrong return type of Database.run(..)

### LINKED ISSUE #809: Return type of Database.run(query) is incorrect
The return type of the function `Database.run(query: ExecuteSqlRequest | string)` is defined as `Promise<Row[]>` (when no callback function is passed to the function). The actual return type of the method is however `Promise<[Array<Row | Json>, spannerClient.spanner.v1.ResultSetStats]`.

olavloite:
The Database.run(query) method declared a return type of `Promise<Row[]>`, while its actual return type was `Promise<[Array<Row | Json>, ResultSetStats]>`.

Fixes #809

### DIFF
--- a/test/spanner.ts
+++ b/test/spanner.ts
@@ -26,7 +26,7 @@ import * as mockDatabaseAdmin from './mockserver/mockdatabaseadmin';
 import * as sinon from 'sinon';
 import {google} from '../protos/protos';
 import {types} from '../src/session';
-import {ExecuteSqlRequest} from '../src/transaction';
+import {ExecuteSqlRequest, RunResponse} from '../src/transaction';
 import {PartialResultStream, Row} from '../src/partial-result-stream';
 import {
   SessionLeakError,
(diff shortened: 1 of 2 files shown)


test/spanner.ts
```diff
@@ -126,12 +127,29 @@ describe('Spanner with mock server', () => {
...
         assert.strictEqual(rows.length, 3);
         let i = 0;
-        rows.forEach(row => {
+        (rows as Row[]).forEach(row => {
           i++;
-          assert.strictEqual(row[0].name, 'NUM');
-          assert.strictEqual(row[0].value.valueOf(), i);
-          assert.strictEqual(row[1].name, 'NAME');
-          assert.strictEqual(row[1].value.valueOf(), numberToEnglishWord(i));
+          const [numCol, nameCol] = row;
```
mayurkale22:
```suggestion
          const [row1, row2] = row;
```

olavloite: The array `row` contains the columns of the current row, so I don't think we should change the name of these variables to `row1` and `row2`. The iteration over the array of rows is done on [line 130](https://github.com/googleapis/nodejs-spanner/blob/d4590bc8054de38d8436aa9fcb0d74d8c9de5235/test/spanner.ts#L130).

PR #812, opened 2020-01-23
fix: retry Session not found for Database.run(..)

olavloite:
This PR adds retry protection for `Session not found` errors for the `Database.run(..)` and `Database.runStream(..)` functions.

### Background
The session pool should normally keep all sessions alive by pinging the them regularly with a simple `SELECT  1`, but it is known from other clients (Java, Go) that users can sporadically experience `Session not found` errors as sessions can be dropped by the backend. These clients have therefore added retry functionality for these errors.


test/spanner.ts
```diff
@@ -512,6 +513,109 @@ describe('Spanner with mock server', () => {
...
+        }
+        assert.strictEqual(rows!.length, 3);
+        db.getSessions((err, results) => {
+          if (err) {
+            assert.fail(err);
+            done();
+            return;
+          }
+          // The server should have exactly 2 sessions. The first one that was
+          // created was removed from the session pool because of the simulated
```
hengfengli: Change to `The first session was removed ...`?

test/spanner.ts
```diff
@@ -512,6 +513,109 @@ describe('Spanner with mock server', () => {
...
+          // 'Session not found' error, but it was not removed from the server.
```
hengfengli: "but it was not removed from the server" I don't quite understand this comment. The session was not removed from the server, why?

test/spanner.ts
olavloite: Done

test/spanner.ts
olavloite: We only simulate the `Session not found` error by telling the mock server to throw this error. The test case does not delete the session on the mock server, so if we check which sessions are still on the mock server, we will get both the original session and the session created by the retry. The first session is however removed from the session pool in the client, as the client thinks that the session is no longer valid.

I've tried to clarify the comment.

PR #815, opened 2020-01-25
tests: cleanup mock server tests

olavloite:
This PR removes warnings, cleans up and improves the test methods using the mock server:
* Removes calls to the deprecated method assert.fail(Error).
* Adds a `Database.close()` call to all test methods to check for session leaks (if a session leak is detected, the test will now fail).
* Adds `Transaction.commit()` calls where these were missing to prevent session leaks.
* Compacts several test cases by using `assert.ifError` and sending errors directly to the done function by adding `.catch(done)` to promise chains.


test/spanner.ts
```diff
@@ -225,12 +225,13 @@ describe('Spanner with mock server', () => {
...
-          assert.fail(err);
-        } else {
-          assert.strictEqual(rows!.length, 3);
-        }
-        done();
+        assert.ifError(err);
+        assert.strictEqual(rows!.length, 3);
+        database
+          .close()
+          // This will cause done to be called with an error and fail the test.
```
skuruppu: Since you're not adding the same comment in all the other places with the same sequence, I would remove this comment.

olavloite: Done.

PR #819, opened 2020-01-30
fix: retry 'Session not found' errors on getSnapshot

olavloite:
If the `BeginTransaction` RPC of a read-only transaction fails with a 'Session not found' error, we can safely discard the invalid session, get a new session from the pool and retry the BeginTransaction call on the new session.

### Background
The session pool should normally keep all sessions alive by pinging them regularly with a simple SELECT 1, but it is known from other clients (Java, Go) that users can sporadically experience Session not found errors as sessions can be dropped by the backend. These clients have therefore added retry functionality for these errors.

### DIFF
--- a/test/spanner.ts
+++ b/test/spanner.ts
@@ -615,6 +615,60 @@ describe('Spanner with mock server', () => {
         done();
       });
     });
+
+    it('should retry "Session not found" errors for Database.getSnapshot() with callbacks', done => {
+      const db = newTestDatabase();
+      const sessionNotFound = {
+        code: status.NOT_FOUND,
+        message: 'Session not found',
+      } as MockError;
+      // The beginTransaction call will fail 3 times with 'Session not found'
+      // before succeeding.
+      spannerMock.setExecutionTime(
+        spannerMock.beginTransaction,
+        SimulatedExecutionTime.ofErrors([
+          sessionNotFound,
+          sessionNotFound,
+          sessionNotFound,
+        ])
+      );
+      db.getSnapshot((err, snapshot) => {
+        assert.ifError(err);
+        snapshot!.run(selectSql, (err, rows) => {
+          assert.ifError(err);
+          assert.strictEqual(rows.length, 3);
+          snapshot!.end();
+          db.close(done);
+        });
+      });
+    });
+
+    it('should retry "Session not found" errors for Database.getSnapshot()', done => {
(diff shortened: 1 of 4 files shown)


test/spanner.ts
```diff
@@ -615,6 +615,60 @@ describe('Spanner with mock server', () => {
...
+        .then(response => {
+          const [snapshot] = response;
+          snapshot
+            .run(selectSql)
+            .then(response => {
+              const [rows] = response;
+              assert.strictEqual(rows.length, 3);
+              snapshot.end();
+              db.close()
+                .then(() => done())
```
hengfengli: Just curious: why can't we write `.then(done)`?

test/spanner.ts
```diff
@@ -615,6 +615,60 @@ describe('Spanner with mock server', () => {
...
+                .catch(done);
+            })
+            .catch(done);
```
hengfengli: If there is an error, should we mark this test `done`? or we should call `assert.fail('there is an error when running select sql.');`?

test/spanner.ts
```diff
@@ -615,6 +615,60 @@ describe('Spanner with mock server', () => {
...
+        })
+        .catch(done);
```
hengfengli: Same here. If it fails to get a snapshot, we probably should mark this test failed.

test/spanner.ts
olavloite: This will automatically fail the test. `done` is a receiver method that will receive the error that is caught, and when done is called with an error, it fails the test.

test/spanner.ts
olavloite: See above.

test/spanner.ts
olavloite: `db.close(done)` is indeed a lot better. I didn't know that would work as well, but I've verified that it does (i.e. the test ends successfully if `db.close()` succeeds, and the test ends with a failure if `db.close()` returns an error). Thanks for the tip.

olavloite:
Created an issue for the flaky failure in https://source.cloud.google.com/results/invocations/06b43955-ccf2-4b0b-aa4e-2394df58f59c/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-spanner%2Fnode10%2Fsystem-test/log

PR #824, opened 2020-02-07
fix: retry 'Session not found' for r/w tx

olavloite:
Read/write transactions already automatically retry `ABORTED` errors. This PR extends the same retry logic to also retry `Session not found` errors.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -1869,6 +1869,10 @@ class Database extends GrpcServiceObject {
         : {};
 
     this.pool_.getWriteSession((err, session?, transaction?) => {
+      if (err && isSessionNotFoundError(err)) {
+        this.runTransaction(options, runFn!);
+        return;
+      }
       if (err) {
         runFn!(err);
         return;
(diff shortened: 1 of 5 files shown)


src/database.ts
```diff
@@ -1962,18 +1971,27 @@ class Database extends GrpcServiceObject {
...
-      runFn,
-      options
-    );
-
-    try {
-      return await runner.run();
-    } finally {
-      this.pool_.release(session);
+    // Loop to retry 'Session not found' errors.
+    while (true) {
```
hengfengli: Should we have a sleep interval before retrying? Also, is it expected to retry it indefinitely?

olavloite: Correct, there's no need to wait, as there's no (known) contention on the backend. The client should just try to get a valid session and execute the transaction as quickly as possible. It should not be possible that it will retry indefinitely, as it will first try with all the sessions that are available in the pool, and once those have all been tried, it will create a new session.

PR #825, opened 2020-02-08
fix: Correctly parse metadata from BatchDML response + fix flaky system test

### LINKED ISSUE #822: System test 'Transactions - batchDml - should stop executing statements if an error occurs' is flaky
The system test 'Transactions - batchDml - should stop executing statements if an error occurs' is flaky. It expects the transaction to fail with error code 3 (`INVALID_ARGUMENT`), but the error code can also be 10 (`ABORTED`).

The `ABORTED` error is not handled automatically by the transaction runner, as the error is caught by the test itself.

https://source.cloud.google.com/results/invocations/06b43955-ccf2-4b0b-aa4e-2394df58f59c/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-spanner%2Fnode10%2Fsystem-test/log

olavloite:
The system test `Spanner/Transactions/batch dml/should stop executing statements if an error occurs` could cause flaky failures, as it caught all errors that could happen while executing a batch DML statement. The test assumed that the error code would always be `INVALID_ARGUMENT`, as the second statement was an invalid statement, but if the transaction was aborted by Spanner before that statement could be reached, the error code would be `ABORTED`.

Further investigation of this problem also showed that the status information that is returned by Spanner did not correspond with the `StatusObject` type that was used by the client. Instead, this is a `grpc.Status` object. This object does not include a `metadata` field. Instead, it has an array of `IAny` which could contain any proto object. One of these proto objects could be a `RetryInfo` object that should be used to determine the wait before retrying the transaction.

The previous implementation that used the field `status.details` to create the error message and `status.metadata` to set the error metadata would lead to an error where:
* The error message would be garbled.
* No metadata would be included.

Fixes #822


olavloite:
See https://github.com/googleapis/nodejs-spanner/issues/829#issuecomment-585900129

JustinBeckwith:
@olavloite should this be ready to go?

olavloite:
> @olavloite should this be ready to go?

@JustinBeckwith Yes, this was only stuck on the build error caused by the failing unit tests on master.

PR #827, opened 2020-02-12
build: add GitHub actions config for unit tests

bcoe:


bcoe:
@olavloite I'm making an effort to standardize our test config, it looks like there are potentially some actual failing tests in Node 12, any thoughts?

olavloite:
@bcoe See https://github.com/googleapis/nodejs-spanner/issues/829#issuecomment-585900129

PR #830, opened 2020-02-13
deps: remove @types/extend

### LINKED ISSUE #829: Unit tests are broken on master
Just cloning the repo and running the tests, looks like the unit tests are somehow broken.  PRs are starting to pile up - we should prioritize ASAP.  

@olavloite @skuruppu

olavloite:
Remove @types/extend as it seems to be broken in combination with Node 12.16.0. `Object.assign(...)` can be used instead of extend. Only the deep copy option of extend is not natively supported by `Object.assign`.

Fixes #829

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
   "devDependencies": {
     "@grpc/proto-loader": "^0.5.1",
     "@types/concat-stream": "^1.6.0",
+    "@types/extend": "^3.0.0",
     "@types/is": "0.0.21",
     "@types/lodash.snakecase": "^4.1.4",
     "@types/merge-stream": "^1.1.2",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,6 @@
   "devDependencies": {
     "@grpc/proto-loader": "^0.5.1",
     "@types/concat-stream": "^1.6.0",
-    "@types/extend": "^3.0.0",
     "@types/is": "0.0.21",
     "@types/lodash.snakecase": "^4.1.4",
     "@types/merge-stream": "^1.1.2",
--- a/test/database.ts
+++ b/test/database.ts
@@ -1592,7 +1592,7 @@ describe('Database', () => {
     it('should send labels correctly', done => {
       const labels = {a: 'b'};
       const options = {a: 'b', labels};
-      const originalOptions = extend({}, options, labels);
+      const originalOptions = extend(true, {}, options);
 
       database.request = config => {
         assert.deepStrictEqual(config.reqOpts.session, {labels});
@@ -1878,7 +1878,7 @@ describe('Database', () => {
       const gaxOpts = {};
       const options = {a: 'a', gaxOptions: gaxOpts};
 
-      const expectedReqOpts = Object.assign({}, options, {
+      const expectedReqOpts = extend({}, options, {
         database: database.formattedName_,
       });
 
--- a/test/database.ts
+++ b/test/database.ts
@@ -28,7 +28,6 @@ import {Instance} from '../src';
 import {TimestampBounds} from '../src/transaction';
 import {ServiceError, status} from 'grpc';
 import {MockError} from './mockserver/mockspanner';
-import extend = require('extend');
 
 let promisified = false;
 const fakePfy = Object.assign({}, pfy, {
--- a/test/database.ts
+++ b/test/database.ts
@@ -17,7 +17,6 @@
 import * as assert from 'assert';
 import {describe, it} from 'mocha';
 import {EventEmitter} from 'events';
-import * as extend from 'extend';
 import {ApiError, util} from '@google-cloud/common';
 import * as proxyquire from 'proxyquire';
 import * as sinon from 'sinon';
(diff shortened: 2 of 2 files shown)


ljharb:
While this is a fine change, the issue isn't `extend`; it's that you're doing this: https://github.com/googleapis/nodejs-spanner/blob/master/test/database.ts#L35 which mutates, and should cause `Object.assign` to throw as well in strict mode.

package.json
```diff
@@ -78,7 +78,6 @@
   "devDependencies": {
     "@grpc/proto-loader": "^0.5.1",
     "@types/concat-stream": "^1.6.0",
-    "@types/extend": "^3.0.0",
```
JustinBeckwith: I'm surprised this works - we are using the actual `extend` package in many other places.  The TypeScript types at `@types/extend` should have no material impact on the code at runtime.  I think we probably want to add this back?

test/database.ts
```diff
@@ -1590,7 +1589,7 @@ describe('Database', () => {
     it('should send labels correctly', done => {
       const labels = {a: 'b'};
       const options = {a: 'b', labels};
-      const originalOptions = extend(true, {}, options);
+      const originalOptions = Object.assign({}, options, labels);
```
JustinBeckwith: This is technically different.  The `true` flag passed into the first argument of the `extend` method signals a *deep* object copy.  `Object.assign` only does shallow clones. For the purposes of the tests it may not matter, just be aware this is a behavior change, and not one we should make make in the `src/` dir without heavy, heavy testing.

olavloite:
> While this is a fine change (converting to Object.assign), the issue isn't `extend`; it's that you're doing this: https://github.com/googleapis/nodejs-spanner/blob/master/test/database.ts#L35 which mutates, and should cause `Object.assign` to throw as well in strict mode.
> 
> "promisify all" patterns that mutate should be avoided.

@ljharb Thanks for your insights on this one, I really appreciate it.
Would you mind elaborating a little bit on what you mean by 'which mutates, and should cause `Object.assign` to throw as well in strict mode'? What confuses me here is the following:
1. The construction that is being used in `Database.ts` is also used in other tests in the same project without problems. See for example https://github.com/googleapis/nodejs-spanner/blob/c0fb78f7d47b2b713d2441f92dca1f4ffd9e7e33/test/instance.ts#L43
1. There are also test files using `Object.assign` with furthermore the same construction as above, also without problems. Example: https://github.com/googleapis/nodejs-spanner/blob/c0fb78f7d47b2b713d2441f92dca1f4ffd9e7e33/test/common/service-object.ts#L23
1. The line that is actually triggering the problem is this one: https://github.com/googleapis/nodejs-spanner/blob/c0fb78f7d47b2b713d2441f92dca1f4ffd9e7e33/test/database.ts#L213

I've reverted most of my change except for the `extend` call on line 213. Using `Object.assign` instead of `extend` here fixes the problem. Also note that running the same test case on Node 12.15.0 works regardless whether I use `Object.assign` or `extend`, but in 12.16.0 it's broken with `extend`.

ljharb:
@olavloite looking at the commented line remaining in the PR, I’m quite confused - the implication is that it’s a bug in extend combined with node 12.16, but i can’t conceive of what it’d be.

If you can replicate the runtime error with extend alone, it’d be awesome if you could file an issue on it!

olavloite:
> @olavloite looking at the commented line remaining in the PR, I’m quite confused - the implication is that it’s a bug in extend combined with node 12.16, but i can’t conceive of what it’d be.
> 
> If you can replicate the runtime error with extend alone, it’d be awesome if you could file an issue on it!

@ljharb I think I've figured out what's happening here. The test code that caused the failure was doing two calls to `extend`, and it was the combination of those two calls with the introduction of a new field on `EventEmitter` in 12.16.0 that caused it:
1. https://github.com/nodejs/node/pull/30932 was added in 12.16.0. This PR adds a non-writable field `errorMonitor` to `EventEmitter`.
1. The relevant class hierarchy in the test case was `Database extends GrpcServiceObject extends ServiceObject extends EventEmitter`.
1. `extend` tries to write the field `EventEmitter.errorMonitor` on the second call, which fails as it is read-only. I'll leave it to the experts whether this should be considered a bug in `extend`, or whether this works as designed and the user should handle this in advance. I don't know enough about `extend` to have a strong opinion on that.

The simplest example that I've been able to define that replicates this problem is listed below. The line `(foo2 as any).errorMonitor = undefined;` fixes the issue. If you remove that line, the code will fail in Node 12.16.0 and succeed in 12.15.0.

```Typescript
import * as extend from 'extend';
import {EventEmitter} from 'events';
import * as assert from 'assert';

class Foo extends EventEmitter {
  constructor() {
    super();
  }
}

describe('extend', () => {
  let foo1: typeof Foo;
  let foo2: typeof Foo;

  it('should work', () => {
    foo1 = class Bar extends EventEmitter {
      constructor() {
        super();
      }
    };
    foo2 = extend({}, foo1);

    // Set errorMonitor to undefined to prevent extend from trying to set it.
    // Without this, the following extend call will fail in Node 12.16.0.
    // In 12.15.0 and earlier, this is not necessary as there is no read-only
    // errorMonitor field.
    // tslint:disable-next-line:no-any
    (foo2 as any).errorMonitor = undefined;
    extend(foo1, foo2);
    assert.ok(foo1);
  });
});
```

ljharb:
This is indeed something user code has to handle; it's not ever safe to blindly mutate code you don't own, which includes anything that you *do* own that extends something you don't own, precisely because you could be extending a nonconfigurable/nonwritable object, or a Proxy, or anything :-)

Thanks for figuring it out!

PR #831, opened 2020-02-14
fix: set read-only field to undefined to prevent error

### LINKED ISSUE #829: Unit tests are broken on master

olavloite:
See https://github.com/googleapis/nodejs-spanner/pull/830#issuecomment-586477413

Fixes #829


test/database.ts
```diff
@@ -210,13 +210,15 @@ describe('Database', () => {
         AsyncTransactionRunner: FakeAsyncTransactionRunner,
       },
     }).Database;
-    // The following commented out line is the one that will trigger the error.
-    // DatabaseCached = extend({}, Database);
-    DatabaseCached = Object.assign({}, Database);
+    DatabaseCached = extend({}, Database);
```
JustinBeckwith: I am totally on board with us switching every usage of `extend` to `Object.assign` in this file - *except* for ones that use deep copy, ie `extend(true, {}, somethin);`

I actually thought I had already made that change in the module, but I guess I missed this one :). The only places we kept `extend` around were for ones that had deep copies in many other modules. 

So I guess I'm saying I'm ok with this change, but I'd really prefer we use `Object.assign` unless we have to use `extend`.  Sorry if this was confusing/poorly communicated in the last review.

skuruppu:
@olavloite should we close this PR since #829 is already fixed?

olavloite:
@skuruppu Yep, this is no longer needed.

PR #838, opened 2020-02-27
build: add publish.yml enabling GitHub app for publishes

bcoe:


skuruppu:
No idea why the automerge is not working so manually merging in.

PR #846, opened 2020-03-04
feat: add support for QueryOptions

olavloite:
Adds the ability to set `QueryOptions` when running Cloud Spanner queries. For now, only setting the `query_optimizer_version` is added.

QueryOptions can be configured through the following mechanisms:

1. Through the `SPANNER_OPTIMIZER_VERSION` environment variable.
1. At `Database` level using `spanner.instance('instance-name').database('database-name', sessionPoolOptions, queryOptions)`.
1. At query level using `ExecuteSqlRequest.queryOptions`.

If the options are configured through multiple mechanisms then:

1. Options set at an environment variable level will override options configured at the `Database` level.
1. Options set at a query-level will override options set at either the `Database` or environment variable level.

If no options are set, the optimizer version will default to:

1. The optimizer version the database is pinned to in the backend.
1. If the database is not pinned to a specific version, then the Cloud Spanner backend will use the "latest" version.

### DIFF
--- a/proto/spanner.d.ts
+++ b/proto/spanner.d.ts
@@ -1146,16 +1146,19 @@ export namespace google {
                 paramTypes?: ({ [k: string]: google.spanner.v1.IType }|null);
 
                 /** ExecuteSqlRequest resumeToken */
-                resumeToken?: (Uint8Array|null);
+                resumeToken?: (Uint8Array|string|null);
 
                 /** ExecuteSqlRequest queryMode */
-                queryMode?: (google.spanner.v1.ExecuteSqlRequest.QueryMode|null);
+                queryMode?: (google.spanner.v1.ExecuteSqlRequest.QueryMode|keyof typeof google.spanner.v1.ExecuteSqlRequest.QueryMode|null);
 
                 /** ExecuteSqlRequest partitionToken */
-                partitionToken?: (Uint8Array|null);
+                partitionToken?: (Uint8Array|string|null);
 
                 /** ExecuteSqlRequest seqno */
-                seqno?: (number|Long|null);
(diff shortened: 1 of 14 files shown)


proto/spanner.d.ts
```diff
@@ -1146,16 +1146,19 @@ export namespace google {
                 paramTypes?: ({ [k: string]: google.spanner.v1.IType }|null);
 
                 /** ExecuteSqlRequest resumeToken */
-                resumeToken?: (Uint8Array|null);
+                resumeToken?: (Uint8Array|string|null);
```
skuruppu: I assume these are auto-generated changes?

olavloite: Yes, although it could be that this change is only 'generated' because of the way I manually generated the proto changes. It should however not make any difference for this PR once the actual proto change has been merged.

olavloite:
> LGTM
> 
> If you can update the PR description to:
> 
> 1. List the precedence
> 2. Mention the env var name
>    that would be great.

Done.

skuruppu:
Thanks for the changes. One more request @olavloite, would you be able to upload any samples you have? I'm not sure whether it should be in this PR or in a separate one but I'll let you decide.

olavloite:
> Thanks for the changes. One more request @olavloite, would you be able to upload any samples you have? I'm not sure whether it should be in this PR or in a separate one but I'll let you decide.

I've added the samples to this PR.

proto/spanner.d.ts
skuruppu: It seems that the generated code change is on master so I think you should be able to revert this file.

PR #847, opened 2020-03-05
fix!: params issue in executeSql

### LINKED ISSUE #845: spanner: client throws error when introducing params: field to request
#### Environment details

  - OS: unknown
  - Node.js version: unknown
  - npm version: unknown
  - `@google-cloud/spanner` version: unknown

#### Steps to reproduce

Customer was interested in getting query execution plan using [Node.js client library for Spanner](https://googleapis.dev/nodejs/spanner/latest/index.html).

The following code works:
```ts
const spanner = require('@google-cloud/spanner');

const client = new spanner.v1.SpannerClient({
  projectID: '[PROJECT-NAME]',
});

var sessionName = '';
var projectName = '[PROJECT-NAME]';
var instanceName = '[INSTANCE-NAME]';
var databaseName = '[DATABASE-NAME]'; 

const formattedDatabase = client.databasePath(projectName, instanceName, databaseName);

client.createSession({database: formattedDatabase})
  .then(responses => {
    const response = responses[0];
    var pathlist = response.name.split('/');
    sessionName = pathlist[7];
    const formattedSession = client.sessionPath(projectName, instanceName, databaseName, sessionName);

    //Create SQL Query and Issue the request with queryMode set to 0 to get only the statement restult, 1 to get the statement result + query plan 2 to get the statement result + query plan + stats
    const sql = 'SELECT tracking_id FROM trackings WHERE tracking_id = 1';
    const request = {
      session: formattedSession,
      sql: sql,
      queryMode: 2,
    };

    client.executeSql(request)
      .then(responses => {
        const response = responses[0];
        console.log(response.stats.queryPlan)
      });

      .catch(err => {
        console.error(err);
      });

      //Delete the session
      const formattedName = client.sessionPath(projectName, instanceName, databaseName, sessionName);
      client.deleteSession({name: formattedName}).catch(err => {
        console.error(err);
      });
  })

  .catch(err => {
    console.error(err);
  })
```

But it fails when a params: field is introduced on the request section that serves as input to the [executeSQL](https://googleapis.dev/nodejs/spanner/latest/v1.SpannerClient.html#executeSql) method:

```ts
...
    const request = {
       session: formattedSession,
       sql: `SELECT tracking_id FROM trackings WHERE tracking_id=@tracking_id`,
       params: {
          tracking_id: '1'
       },
       queryMode: 2,
     };

const [response] = await client.executeSql(request)
...
```

When introducing this field, the code will fail to return any response and results in the following error message:
```ts
{ Error: 3 INVALID_ARGUMENT: No parameter found for binding: tracking_id

at Object.callErrorFromStatus (/home/[XXXX]/nodejs-spanner/samples/node_modules/@grpc/grpc-js/build/src/call.js:30:26)

at Http2CallStream.call.on (/home/[XXXXX]/nodejs-spanner/samples/node_modules/@grpc/grpc-js/build/src/client.js:96:33)

at Http2CallStream.emit (events.js:187:15)

at process.nextTick (/home/[XXXXX]/nodejs-spanner/samples/node_modules/@grpc/grpc-js/build/src/call-stream.js:100:22)

at process._tickCallback (internal/process/next_tick.js:61:11)

code: 3,
details: 'No parameter found for binding: tracking_id',
metadata:
Metadata {
internalRepr: Map { 'grpc-server-stats-bin' => [Array] },
options: {} },
note:
'Exception occurred in retry method that was not classified as transient' }
```

laljikanjareeya:
This is PR is Address to encode the params as it does in `database.run` using [encodeParams](https://github.com/googleapis/nodejs-spanner/blob/master/src/transaction.ts#L992) function, but it breaks the existing code.
Fixes #845

### DIFF
--- a/src/v1/spanner_client.js
+++ b/src/v1/spanner_client.js
@@ -17,6 +17,7 @@
 const gapicConfig = require('./spanner_client_config.json');
 const gax = require('google-gax');
 const path = require('path');
+const snapshot = require('../transaction').Snapshot;
 
 const VERSION = require('../../../package.json').version;
 
@@ -832,6 +833,9 @@ class SpannerClient {
       session: request.session,
     });
 
+    const {params, paramTypes} = snapshot.encodeParams(request);
+    request.params = params;
+    request.paramTypes = paramTypes;
     return this._innerApiCalls.executeSql(request, options, callback);
   }
 
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -34,6 +34,9 @@ import {Row} from '../src/partial-result-stream';
 import {GetDatabaseConfig} from '../src/database';
 import {status} from 'grpc';
 
+const spannerModule = require('../src');
+const v1Client = new spannerModule.v1.SpannerClient();
+
 const PREFIX = 'gcloud-tests-';
 const RUN_ID = shortUUID();
 const LABEL = `gcloud-tests-${RUN_ID}`;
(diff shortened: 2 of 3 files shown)


laljikanjareeya:
> Thanks for the fix @laljikanjareeya. Could you please also add a unit test to make sure that this is fixed?

Added Unit Test as well as system test.

src/v1/spanner_client.js
```diff
@@ -17,6 +17,7 @@
 const gapicConfig = require('./spanner_client_config.json');
 const gax = require('google-gax');
 const path = require('path');
+const snapshot = require('../transaction').Snapshot;
```
olavloite: This is a generated file that should not be edited manually. The generated code in the other languages (that I know) also include that in the comments, but unfortunately that is not the case for NodeJS, so I can perfectly understand the confusion.

skuruppu:
> While this change would make the specific example work, the real problem in this case is that the user is using the generated gapic client. When doing so, the user must also take care to specify the parameters in encoded form in the `ExecuteSqlRequest`. See also my comment in the related issue.

@olavloite correct me if I'm wrong but I think this means we don't need this PR at all since we don't want to modify generated code? I asked the folks dealing with the customer issue to forward your suggestion from the issue.

olavloite:
> @olavloite correct me if I'm wrong but I think this means we don't need this PR at all since we don't want to modify generated code? I asked the folks dealing with the customer issue to forward your suggestion from the issue.

Correct, so closing this PR.

PR #879, opened 2020-03-31
fix: remove src/common-grpc/operation.ts

### LINKED ISSUE #878: Delete `src/common-grpc/operation.ts`
The `@google-cloud/common-grpc` package was deleted but before the deletion, the code was moved to `src/common-grpc`. Since this code hasn't been maintained, `src/common-grpc/operation.ts` is no longer functional.

@alexander-fenster said that it would be better to use the Gax operations directly. 
Since no code in the repo is using the `GrpcOperation` exported by `operation.ts`, we will delete it.

skuruppu:
Fixes #878 🦕


stephenplusplus:
Nice! 👍

This technically deprecates the `spanner.operation('op-name')` method, which would be a breaking change. It would be nice if we could still support it through gax. [gapic-generator#1778](https://github.com/googleapis/gapic-generator/issues/1778) & [gapic-generator-typescript#170](https://github.com/googleapis/gapic-generator-typescript/issues/170) was looking into this.

alexander-fenster:
@stephenplusplus We're looking into this! @xiaozhenliu-gg5 This is one more example where instantiating an operation by name is useful (we talked today about this use case).

skuruppu:
Thanks @alexander-fenster and @stephenplusplus :)

@stephenplusplus, if you could also approve the change, that would be wonderful.

As for the breaking change aspect of it, I'm happy to put a "do not merge" on this until the node 8 deprecation. We already have several PRs in this repo waiting for it.

PR #884, opened 2020-04-03
fix: update samples-test script

### LINKED ISSUE #866: spanner: unable to run samples tests locally
#### Environment details

  - OS: Linux
  - Node.js version: v10.17.0
  - npm version: 5.8.0
  - `@google-cloud/spanner` version: 4.8.0

#### Steps to reproduce

  1. Checkout the master branch of the repo.
  2. `npm install`
  3. `npm run samples-test`

#### Expected output

The samples tests to run.

#### Actual output

```
  Spanner
    1) "before all" hook for "should create an example database"


  0 passing (12s)
  1 failing

  1) Spanner
       "before all" hook for "should create an example database":
     TypeError: Cannot read property 'filter' of undefined
      at Promise.all.instances.map (system-test/spanner.test.js:82:14)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

This doesn't seem to be a problem on Kokoro since I see that the tests pass there ([example run](https://source.cloud.google.com/results/invocations/338ce8f7-0863-4f5c-abfd-63c280462582/targets)).

AVaksman:

Fixes #866


package.json
```diff
@@ -30,7 +30,7 @@
     "docs": "jsdoc -c .jsdoc.js",
     "predocs": "npm run compile",
     "lint": "gts fix",
-    "samples-test": "cd samples/ && npm link ../ && npm test && cd ../",
+    "samples-test": "cd samples/ && rm -rf node_modules package-lock.json && npm link ../ && npm install && npm test && cd ../",
```
bcoe, line 33: removing `node_modules` and `package-lock.json` as part of `sample-tests` seems weird to me. In CI/CD these files don't exist, in a local environment, this may result in you needing to reinstall dependencies repeatedly as you test.

stephenplusplus: @AVaksman is the idea to force the latest code in src/ to compile and be relinked? Does just this work?:

```suggestion
    "samples-test": "cd samples/ && npm link ../ && npm install && npm test",
```

npm install should noop relatively quickly if nothing needs to be done. Removing package-lock.json for the user makes sense, although is there a way to disable package-lock.json on a repo level?

alexander-fenster: I am strongly opposed to hiding `rm -rf` in scripts.

AVaksman: adding `npm install` to the script should be enough. `rm -rf` is way to harsh.
to disable package-lock.json on a repo level an `.npmrc` with `package-lock=false` could be added.
WDYT?

stephenplusplus: I'd be okay with disabling package-lock.json. There's not a good time I can think of that a maintainer should be okay using stale dependencies to run the tests against.

PR #925, opened 2020-04-24
fix(types)!: properly format listing methods with gaxOptions

AVaksman:

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import {ServiceError, CallOptions} from 'grpc';
-import {Operation as GaxOperation} from 'google-gax';
+import {ServiceError} from 'grpc';
+import {Operation as GaxOperation, CallOptions} from 'google-gax';
 import {google as instanceAdmin} from '../protos/protos';
 import {google as databaseAdmin} from '../protos/protos';
 
@@ -33,9 +33,7 @@ export interface ResourceCallback<Resource, Response> {
     response?: Response
   ): void;
 }
-export type PagedResponse<Item, Response> =
-  | [Item[]]
-  | [Item[], {} | null, Response];
+export type PagedResponse<Item, Response> = [Item[], {} | null, Response];
 
 export type RequestCallback<T, R = void> = R extends void
   ? NormalCallback<T>
@@ -63,8 +61,8 @@ export interface LongRunningCallback<Resource> {
   ): void;
 }
 
-export type PagedRequest<P> = P & {
-  autoPaginate?: boolean;
-  maxApiCalls?: number;
+export interface PagedOptions {
+  pageSize?: number;
+  pageToken?: string;
   gaxOptions?: CallOptions;
-};
+}
--- a/src/common.ts
+++ b/src/common.ts
@@ -66,3 +66,7 @@ export interface PagedOptions {
   pageToken?: string;
   gaxOptions?: CallOptions;
 }
+
+export interface PagedOptionsWithFilter extends PagedOptions {
+  filter?: string;
+}
--- a/src/database.ts
+++ b/src/database.ts
@@ -1176,7 +1176,7 @@ class Database extends GrpcServiceObject {
   /**
    * @typedef {array} GetSessionsResponse
    * @property {Session[]} 0 Array of {@link Session} instances.
-   * @param {object} nextQuery A query object to to receive more results.
+   * @param {object} nextQuery A query object to receive more results.
    * @property {object} 1 The full API response.
    */
   /**
--- a/src/database.ts
+++ b/src/database.ts
@@ -1175,8 +1175,8 @@ class Database extends GrpcServiceObject {
   /**
    * @typedef {array} GetSessionsResponse
    * @property {Session[]} 0 Array of {@link Session} instances.
-   * @property {object} nextQuery A query object to receive more results.
-   * @property {object} 1 The full API response.
+   * @property {object} 1 A query object to receive more results.
+   * @property {object} 2 The full API response.
    */
   /**
    * @callback GetSessionsCallback
--- a/src/database.ts
+++ b/src/database.ts
@@ -1243,10 +1243,26 @@ class Database extends GrpcServiceObject {
       typeof optionsOrCallback === 'object'
         ? optionsOrCallback
         : ({} as GetSessionsOptions);
-    const reqOpts = extend({}, options, {
+    const gaxOpts = extend(true, {}, options.gaxOptions);
+    let reqOpts = extend({}, options, {
       database: this.formattedName_,
     });
     delete reqOpts.gaxOptions;
+
+    // Copy over pageSize and pageToken values from gaxOptions.
+    // However values set on options take precedence.
+    if (gaxOpts) {
+      reqOpts = extend(
+        {},
+        {
+          pageSize: gaxOpts.pageSize,
+          pageToken: gaxOpts.pageToken,
+        },
+        reqOpts
+      );
+      delete gaxOpts.pageSize;
+      delete gaxOpts.pageToken;
+    }
 
     this.request<
       google.spanner.v1.ISession,
@@ -1256,7 +1272,7 @@ class Database extends GrpcServiceObject {
         client: 'SpannerClient',
         method: 'listSessions',
         reqOpts,
-        gaxOpts: options.gaxOptions,
+        gaxOpts,
       },
       (err, sessions, ...args) => {
         let sessionInstances: Session[] | null = null;
--- a/src/database.ts
+++ b/src/database.ts
@@ -71,11 +71,11 @@ import {
   IOperation,
   Schema,
   RequestCallback,
-  PagedOptions,
   ResourceCallback,
   PagedResponse,
   NormalCallback,
   LongRunningCallback,
+  PagedOptionsWithFilter,
 } from './common';
 import {ServiceError} from 'grpc';
 import {Readable, Transform, Duplex} from 'stream';
@@ -120,9 +120,7 @@ type PoolRequestCallback = RequestCallback<Session>;
 
 type ResultSetStats = spannerClient.spanner.v1.ResultSetStats;
 
-interface GetSessionsOptions extends PagedOptions {
-  filter?: string;
-}
+type GetSessionsOptions = PagedOptionsWithFilter;
 
 /**
  * IDatabase structure with database state enum translated to string form.
@@ -589,6 +587,8 @@ class Database extends GrpcServiceObject {
    *   * Label values must be between 0 and 63 characters long and must conform
    *     to the regular expression `([a-z]([-a-z0-9]*[a-z0-9])?)?`.
    *   * No more than 64 labels can be associated with a given session.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} CreateSessionResponse
--- a/src/database.ts
+++ b/src/database.ts
@@ -1152,9 +1152,7 @@ class Database extends GrpcServiceObject {
   /**
    * Options object for listing sessions.
    *
-   * @typedef {object} GetSessionsRequest
-   * @property {boolean} [autoPaginate=true] Have pagination handled
-   *     automatically.
+   * @typedef {object} GetSessionsOptions
    * @property {string} [filter] An expression for filtering the results of the
    *     request. Filter rules are case insensitive. The fields eligible for
    *     filtering are:
@@ -1169,21 +1167,23 @@ class Database extends GrpcServiceObject {
    *     - **`labels.env:dev`** The instance's label env has the value dev.
    *     - **`name:howl labels.env:dev`** The instance's name is howl and it has
    *       the label env with value dev.
-   * @property {number} [maxApiCalls] Maximum number of API calls to make.
-   * @property {number} [maxResults] Maximum number of items to return.
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetSessionsResponse
    * @property {Session[]} 0 Array of {@link Session} instances.
+   * @param {object} nextQuery A query object to to receive more results.
    * @property {object} 1 The full API response.
    */
   /**
    * @callback GetSessionsCallback
    * @param {?Error} err Request error, if any.
    * @param {Session[]} instances Array of {@link Session} instances.
+   * @param {object} nextQuery A query object to to receive more results.
    * @param {object} apiResponse The full API response.
    */
   /**
@@ -1221,7 +1221,7 @@ class Database extends GrpcServiceObject {
    * }
    *
    * database.getInstances({
-   *   autoPaginate: false
+   *   gaxOptions: {autoPaginate: false}
    * }, callback);
    *
    * //-
@@ -1238,19 +1238,16 @@ class Database extends GrpcServiceObject {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     const self = this;
     const callback =
-      typeof optionsOrCallback === 'function'
-        ? (optionsOrCallback as GetSessionsCallback)
-        : cb;
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
     const options =
       typeof optionsOrCallback === 'object'
-        ? (optionsOrCallback as GetSessionsOptions)
-        : {gaxOptions: {}};
-    const gaxOpts: CallOptions = options.gaxOptions as CallOptions;
+        ? optionsOrCallback
+        : ({} as GetSessionsOptions);
     const reqOpts = extend({}, options, {
       database: this.formattedName_,
     });
-
     delete reqOpts.gaxOptions;
+
     this.request<
       google.spanner.v1.ISession,
       google.spanner.v1.IListSessionsResponse
@@ -1259,7 +1256,7 @@ class Database extends GrpcServiceObject {
         client: 'SpannerClient',
         method: 'listSessions',
         reqOpts,
-        gaxOpts,
+        gaxOpts: options.gaxOptions,
       },
       (err, sessions, ...args) => {
         let sessionInstances: Session[] | null = null;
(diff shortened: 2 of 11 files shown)


src/database.ts
```diff
@@ -1158,21 +1167,23 @@ class Database extends GrpcServiceObject {
...
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetSessionsResponse
    * @property {Session[]} 0 Array of {@link Session} instances.
+   * @param {object} nextQuery A query object to to receive more results.
```
stephenplusplus, line 1179:
```suggestion
   * @param {object} nextQuery A query object to receive more results.
```

src/index.ts
```diff
@@ -597,7 +598,7 @@ class Spanner extends GrpcService {
    *     config.
    * @param {string} instanceConfigs.displayName The name of the instance config
    *     as it appears in UIs.
-   * @param {object} apiResponse The full API response.
+   * @param {object} nextQuery A query object to to receive more results.   * @param {object} apiResponse The full API response.
```
stephenplusplus, line 601:
```suggestion
   * @param {object} nextQuery A query object to to receive more results.
   * @param {object} apiResponse The full API response.
```

src/instance.ts
```diff
@@ -365,22 +365,31 @@ class Instance extends common.GrpcServiceObject {
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
```
stephenplusplus, line 368: How does it work that we have top-level "pageSize" and "pageToken", but they're also in `gaxOptions` (e.g. "gaxOptions.pageSize")?

AVaksman: Thank you!

The existing behavior is as follows:

Current options for listing methods typically look something like this
```ts
{
	parent?: string;
	pageSize?: number;
	pageToken?: string;
	filter?: string;
	autoPaginate?: boolean;
	maxApiCalls?: number;
	gaxOptions?: CallOptions;
}
```
Where `gaxOptions` (among others) contain the following options as well
```ts
{
	autoPaginate?: boolean;
    	pageToken?: string;
    	pageSize?: number;
}
```
Which is utilized as 
```ts
instance.getBackups({
        pageSize: 3,
        pageToken,
        gaxOptions: {autoPaginate: false},
});
```
in [system tests](https://github.com/googleapis/nodejs-spanner/blob/5624b7bafb585adcbd0a7c9d53f728ff77afb1fa/system-test/spanner.ts#L1136-L1140) as well as [samples](https://github.com/googleapis/nodejs-spanner/blob/5624b7bafb585adcbd0a7c9d53f728ff77afb1fa/samples/backups-get.js#L100-L104).


In terms of `reqOpts` and `gaxOpts` of the `request()` method the following options are respected
```ts
reqOpts = {
	parent: string;
	pageSize?: number;
	pageToken?: string;
	filter?: string;
}
gaxOpts = {
	autoPaginate?: boolean;
}
```
having `pageSize` and `pageToken` in two places is confusing and misleading. Should we just remove them from the top-level and pass to the `reqOts` from the the `gaxOptions` object, or accept and respect `pageSize` and `pageToken` values from both top-level and `gaxOptios`, or another option?
WDYT @stephenplusplus

stephenplusplus: This is indeed tricky! I don't think we have the option to remove it from the top-level, since that would conflict with the raw API documentation, e.g. https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.database.v1#google.spanner.admin.database.v1.ListBackupsRequest which accepts "pageSize" and "pageToken".

Do you know where & what gax is doing with "pageSize" and "pageToken". It seems like the wrong place to send a native request parameter, as opposed to the other options in "gaxOptions" which are all custom helpers like "autoPaginate", "bundleOptions", etc.

AVaksman: > Do you know where & what gax is doing with "pageSize" and "pageToken". It seems like the wrong place to send a native request parameter, as opposed to the other options in "gaxOptions" which are all custom helpers like "autoPaginate", "bundleOptions", etc.

Perhaps these values where used by [`paginator`](https://github.com/googleapis/nodejs-paginator/blob/cabc0489190cfae73178354841286a3e89779d3f/src/index.ts#L41-L53).

Speaking of which, I believe it (`paginator.streamify`) could be dropped as both [paginated](https://github.com/googleapis/nodejs-spanner/blob/5624b7bafb585adcbd0a7c9d53f728ff77afb1fa/src/v1/database_admin_client.ts#L2039) and [streaming](https://github.com/googleapis/nodejs-spanner/blob/5624b7bafb585adcbd0a7c9d53f728ff77afb1fa/src/v1/database_admin_client.ts#L2150) calls can be handled now.

stephenplusplus: Should gax stop listing those under CallOptions?

AVaksman: refactored to pass `pageSize` and `pageToken` as `reqOpts`.
PTAL

src/database.ts
```diff
@@ -1227,19 +1238,32 @@ class Database extends GrpcServiceObject {
...
+        ? optionsOrCallback
+        : ({} as GetSessionsOptions);
+    const gaxOpts = extend(true, {}, options.gaxOptions);
+    let reqOpts = extend({}, options, {
       database: this.formattedName_,
     });
-
     delete reqOpts.gaxOptions;
+
+    // Copy over pageSize and pageToken values from gaxOptions.
```
stephenplusplus, line 1252: Since we're in a breaking change, let's not even worry about supporting this improper usage. Most people have probably been using it in `reqOpts` already anyway.

src/instance.ts
```diff
@@ -264,12 +259,22 @@ class Instance extends common.GrpcServiceObject {
...
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetBackupsResponse
    * @property {Backup[]} 0 Array of {@link Backup} instances.
+   * @param {object} nextQuery A query object to to receive more results.
```
skuruppu, line 268:
```suggestion
   * @param {object} nextQuery A query object to receive more results.
```

src/instance.ts
skuruppu: Thanks for trying to clean up this mess @AVaksman. I'm not a fan of having to support `pageSize` and `pageToken` from `CallOptions` but I feel like we have no choice :( Given that it's in the documentation, it's very likely someone will try to set it in `gaxOptions` and wonder why it's not being respected.

src/database.ts
skuruppu: I completely agree with this but I think we have no choice but to do this while `CallOptions` still has `pageSize` and `pageToken` parameters (see my reply [here](https://github.com/googleapis/nodejs-spanner/pull/925/files#r416565445)). Someone is bound to set it here instead of in `reqOpts`.

test/database.ts
```diff
@@ -1894,18 +1895,85 @@ describe('Database', () => {
...
+        assert.notDeepStrictEqual(config.gaxOpts, gaxOptions);
+        assert.deepStrictEqual(config.gaxOpts, expectedGaxOpts);
+
+        done();
+      };
+
+      database.getSessions(options, assert.ifError);
+    });
+
+    it('pageSize and pageToken in options should take precedence over over gaxOptions', done => {
```
skuruppu, line 1933: Here and below:

```suggestion
    it('pageSize and pageToken in options should take precedence over gaxOptions', done => {
```

test/database.ts
```diff
@@ -1894,18 +1895,85 @@ describe('Database', () => {
...
-        assert.strictEqual(config.gaxOpts, gaxOpts);
+        assert.deepStrictEqual(config.gaxOpts, gaxOpts);
         done();
       };
 
       database.getSessions(options, assert.ifError);
     });
 
-    it('should not require a query', done => {
+    it('should respect pageSize & pageToken on gaxOptions, pass values in reqOpts only', done => {
```
skuruppu, line 1905: Here and below.

```suggestion
    it('should pass pageSize and pageToken from gaxOptions into reqOpts', done => {
```

src/common.ts
```diff
@@ -63,8 +61,8 @@ export interface LongRunningCallback<Resource> {
   ): void;
 }
 
-export type PagedRequest<P> = P & {
-  autoPaginate?: boolean;
-  maxApiCalls?: number;
+export interface PagedOptions {
+  pageSize?: number;
+  pageToken?: string;
```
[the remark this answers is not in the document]
AVaksman: Done!

src/database.ts
```diff
@@ -576,6 +578,18 @@ class Database extends GrpcServiceObject {
...
+   *
+   * @typedef {object} CreateSessionOptions
+   * @property {Object.<string, string>} [labels] The labels for the session.
+   *
+   *   * Label keys must be between 1 and 63 characters long and must conform to
+   *     the following regular expression: `[a-z]([-a-z0-9]*[a-z0-9])?`.
+   *   * Label values must be between 0 and 63 characters long and must conform
+   *     to the regular expression `([a-z]([-a-z0-9]*[a-z0-9])?)?`.
+   *   * No more than 64 labels can be associated with a given session.
+   */
```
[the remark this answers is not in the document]
AVaksman: added, thanks

src/index.ts
```diff
@@ -94,7 +87,7 @@ export interface RequestConfig {
   method: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   reqOpts: any;
-  gaxOpts?: {};
+  gaxOpts?: CallOptions;
```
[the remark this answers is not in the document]
AVaksman: I've seen this usage across other products as well.

src/instance.ts
```diff
@@ -946,33 +1073,33 @@ class Instance extends common.GrpcServiceObject {
...
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetDatabasesResponse
    * @property {Database[]} 0 Array of {@link Database} instances.
-   * @property {object} 1 The full API response.
+   * @param {object} nextQuery A query object to receive more results.
```
skuruppu, line 1095:
```suggestion
   * @property {object} nextQuery A query object to receive more results.
```

src/instance.ts
```diff
@@ -263,12 +255,22 @@ class Instance extends common.GrpcServiceObject {
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetBackupsResponse
    * @property {Backup[]} 0 Array of {@link Backup} instances.
+   * @param {object} nextQuery A query object to receive more results.
```
skuruppu, line 264:
```suggestion
   * @property {object} nextQuery A query object to receive more results.
```

src/instance.ts
```diff
@@ -364,12 +447,23 @@ class Instance extends common.GrpcServiceObject {
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetBackupOperationsResponse
    * @property {IOperation[]} 0 Array of {@link IOperation} instances.
+   * @param {object} nextQuery A query object to receive more results.
```
skuruppu, line 456:
```suggestion
   * @property {object} nextQuery A query object to receive more results.
```

src/instance.ts
```diff
@@ -459,12 +564,23 @@ class Instance extends common.GrpcServiceObject {
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetDatabaseOperationsResponse
    * @property {IOperation[]} 0 Array of {@link IOperation} instances.
+   * @param {object} nextQuery A query object to receive more results.
```
skuruppu, line 573:
```suggestion
   * @property {object} nextQuery A query object to receive more results.
```

src/index.ts
```diff
@@ -463,21 +421,21 @@ class Spanner extends GrpcService {
...
-   * @property {number} [maxApiCalls] Maximum number of API calls to make.
-   * @property {number} [maxResults] Maximum number of items to return.
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
    */
   /**
    * @typedef {array} GetInstancesResponse
    * @property {Instance[]} 0 Array of {@link Instance} instances.
+   * @param {object} nextQuery A query object to receive more results.
```
skuruppu, line 431:
```suggestion
   * @property {object} nextQuery A query object to receive more results.
```

src/index.ts
```diff
@@ -556,32 +539,98 @@ class Spanner extends GrpcService {
...
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+
    */
   /**
    * @typedef {array} GetInstanceConfigsResponse
    * @property {object[]} 0 List of all available instance configs.
    * @property {string} 0.name The unique identifier for the instance config.
    * @property {string} 0.displayName The name of the instance config as it
    *     appears in UIs.
+   * @param {object} nextQuery A query object to receive more results.
```
skuruppu, line 633:
```suggestion
   * @property {object} nextQuery A query object to receive more results.
```

src/database.ts
```diff
@@ -1157,21 +1166,23 @@ class Database extends GrpcServiceObject {
...
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetSessionsResponse
    * @property {Session[]} 0 Array of {@link Session} instances.
+   * @param {object} nextQuery A query object to receive more results.
```
skuruppu, line 1178:
```suggestion
   * @property {object} nextQuery A query object to receive more results.
```

src/instance.ts
AVaksman: Done, thanks

src/instance.ts
AVaksman: Done

src/instance.ts
AVaksman: Done

src/index.ts
AVaksman: Done

src/index.ts
AVaksman: Done

src/database.ts
AVaksman: And done

test/database.ts
AVaksman: Rephrased everywhere

test/database.ts
AVaksman: Removed double everywhere

src/instance.ts
AVaksman: Removed double everywhere

src/index.ts
AVaksman: Fixed, thanks

src/database.ts
AVaksman: Removed doubles everywhere

src/database.ts
stephenplusplus: That's a good point! I sent a PR to gax to remove those options. After that, we can re-generate the docs, and remove them. I would be surprised if we had any users putting those options in there. Already, turning off auto-pagination and setting `pageSize` and `pageToken` manually is rare. Setting options through gax (a very unfamiliar concept for a user) would be even more unlikely.

I would be okay with still dropping our backup support while we wait for the changes from the linked PR to take effect. Otherwise, I think waiting is worth it, to avoid the complexity of support. Of course, we can always come back and remove the code later, so whatever everyone thinks is best :)

skuruppu: @stephenplusplus do you know how long it would take to get the gax PR merged and into a release?

The main concern I have is that this PR needs to go into v5.0.0. We need v5.0.0 to go out asap since it's got the backup & restore support which is already a GA'ed feature but is currently not supported by a Node.js release.

So if we think the gax change will take more than a week to get through, then we should merge this as is then remove the code later.

stephenplusplus: I'm not sure how long that would take. Hopefully @alexander-fenster can take a look at that PR (https://github.com/googleapis/gax-nodejs/pull/821) when he's available and give us a better idea.

alexander-fenster: We just made a semver major for gax a few weeks ago. From one hand, major releases are cheap, but from another hand, having too many majors makes it harder to hot fix security issues so I would better wait for some time before making gax v3.0.0. So I would probably suggest you go to v5.0.0 here as is. What do you think @skuruppu @stephenplusplus?

stephenplusplus: Thanks @alexander-fenster. It's okay with me to wait. The temporary code won't be too bad. Alternatively, I still think we're safe to remove support for the options anyway, and if anybody in the world notices, we can release a hotfix. But that's completely your call @skuruppu.

skuruppu: I agree that it's so very unlikely that anyone would be trying to set these options through CallOptions. But just in case, we'll leave the temporary code. It won't be too hard to remove it when gax v3.0.0 is ready. I'm open to doing another major release whenever we need to.

src/database.ts
```diff
@@ -1157,21 +1166,23 @@ class Database extends GrpcServiceObject {
...
-   * @property {number} [maxResults] Maximum number of items to return.
    * @property {number} [pageSize] Maximum number of results per page.
    * @property {string} [pageToken] A previously-returned page token
    *     representing part of the larger set of results to view.
+   * @property {object} [gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
    */
   /**
    * @typedef {array} GetSessionsResponse
    * @property {Session[]} 0 Array of {@link Session} instances.
```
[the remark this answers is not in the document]
AVaksman: Good catch, thanks!
Done

skuruppu:
@stephenplusplus, let us know if there's anything else we should change in this PR. Your requested changes status is blocking merging this PR :)

stephenplusplus:
@skuruppu sorry about that!

PR #930, opened 2020-04-28
fix: update spanner package in sample

### LINKED ISSUE #926: Mocha Tests: should restore database from a backup failed
This test failed!

To configure my behavior, see [the Build Cop Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop).

If I'm commenting on this issue too often, add the `buildcop: quiet` label and
I will stop commenting.

---

commit: 1222528c8aa09593aeec21583062c696aad55610
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/a0045f89-b10b-4396-a4e4-031b894edea3), [Sponge](http://sponge2/a0045f89-b10b-4396-a4e4-031b894edea3)
status: failed

laljikanjareeya:
Fixes #926

### DIFF
--- a/samples/package.json
+++ b/samples/package.json
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "@google-cloud/precise-date": "^1.0.0",
-    "@google-cloud/spanner": "^4.8.0",
+    "@google-cloud/spanner": "^5.0.0-beta.0",
     "yargs": "^15.0.0"
   },
   "devDependencies": {


bcoe:
@skuruppu @laljikanjareeya 👋 easy to miss this, but it's important that the version in `samples/package.json` and `package.json` lineup, because of the way sample tests work (otherwise samples are tested against a published version of the library, rather than the local version). _tldr; @alexander-fenster dealt with a bug today related to the discrepancy._

I think we did this because we're blocked on Node 10 right now? hopefully we won't end up with a log jam like this frequently. But, if we do again, we probably need to find another approach to test an unreleased library version.

skuruppu:
Oh yeah that makes sense @bcoe. Sorry we missed that. This should be rare though. Normally we'd be doing much more frequent releases without breaking changes.

PR #933, opened 2020-05-01
fix: skip some tests when run against the emulator

### LINKED ISSUE #932: Skip some system tests when run against the emulator
We want the system tests to be run against the [Cloud Spanner Emulator](https://cloud.google.com/spanner/docs/emulator). But it has some limitations so some tests need to be skipped.

skuruppu:

Fixes #932 🦕

### DIFF
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -686,23 +686,22 @@ describe('Spanner', () => {
     });
 
     it('should respect the FieldMask', async function () {
-      if (!emulatorEnabled) {
-        const fieldNames = ['name', 'displayName'];
-
-        const [metadata] = await instance.getMetadata({fieldNames});
-        assert.deepStrictEqual(metadata['endpointUris'], []);
-        assert.deepStrictEqual(metadata['labels'], {});
-        assert.strictEqual(metadata.name, instance.formattedName_);
-        assert.ok(!metadata['config']);
-        assert.strictEqual(
-          metadata['displayName'],
-          instance.formattedName_.split('/').pop()
-        );
-        assert.strictEqual(metadata['nodeCount'], 0);
-        assert.strictEqual(metadata['state'], 'STATE_UNSPECIFIED');
-      } else {
+      if (emulatorEnabled) {
         this.skip();
       }
+      const fieldNames = ['name', 'displayName'];
+
+      const [metadata] = await instance.getMetadata({fieldNames});
+      assert.deepStrictEqual(metadata['endpointUris'], []);
+      assert.deepStrictEqual(metadata['labels'], {});
+      assert.strictEqual(metadata.name, instance.formattedName_);
+      assert.ok(!metadata['config']);
+      assert.strictEqual(
+        metadata['displayName'],
+        instance.formattedName_.split('/').pop()
+      );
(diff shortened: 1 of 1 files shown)


hengfengli:
Hi @AVaksman, @bcoe, do you have any idea why `ci/windows` fails? I can see it also fails in another PR for the same reason: https://github.com/googleapis/nodejs-spanner/pull/931

hengfengli:
@skuruppu I have skipped all needed tests and fixed related errors. I am not sure: 

1) why "ci / windows" fails. It happens in other PRs as well. 
2) Who should I assign this as a reviewer? I guess you're the PR creator and I can't assign it to you.

system-test/spanner.ts
```diff
@@ -64,6 +64,10 @@ describe('Spanner', () => {
     },
   };
 
+  let emulatorEnabled: string | undefined;
+
+  beforeEach(() => (emulatorEnabled = process.env.SPANNER_EMULATOR_HOST));
```
stephenplusplus, line 69: This check should be able to be set outside of a `before`/`beforeEach` block. And just to be a little more accurate with the var name's meaning, can we convert this value to a boolean?:

```ts
describe('Spanner', () => {
  const IS_EMULATOR_ENABLED = typeof process.env.SPANNER_EMULATOR_HOST !== 'undefined';
```

system-test/spanner.ts
```diff
@@ -1051,13 +1070,23 @@ describe('Spanner', () => {
     });
 
     after(async () => {
-      await restoreDatabase.delete();
```
stephenplusplus, line 1054: Would doing this here work:?

```ts
if (emulatorEnabled) {
  this.skip();
}
```

hengfengli: This doesn't work. I tried this at the beginning.

hengfengli: I got the following error: 

```bash
     Error: `this.skip` forbidden
```

system-test/spanner.ts
hengfengli: Done.

system-test/spanner.ts
skuruppu: I did some research. The [documentation](https://mochajs.org/#inclusive-tests) says to just do a `return` instead of a `skip`. I updated the code.

hengfengli: Good catch. Thanks for making this more elegant.

hengfengli:
@skuruppu I think we need to clean up the leftover instances in the GCP project. It fails to add a new instance.

skuruppu:
I agree, it's weird though because the tests actually try to clean up the instances before it creates a new one. I'm not sure why that's not working. I suppose it's because we try to [create an instance](https://github.com/googleapis/nodejs-spanner/blob/43765d83d2835eda78a144e4139703e62c4a2266/samples/system-test/spanner.test.js#L132) before deleting the old ones?

hengfengli:
I guess the instance ID is generated by date: 

```ts
const INSTANCE_ID = process.env.SPANNERTEST_INSTANCE || `test-instance-${date}`;
```

The old one from yesterday is not deleted.

skuruppu:
I created #955, let's see if it runs.

PR #936, opened 2020-05-05
fix: pause request stream on backpressure

### LINKED ISSUE #934: Spanner *runStream* cause a lot memory consumption
#### Environment details

  - windows:
  - Node.js 10.x:
  - npm version: 6.9.0
  - `@google-cloud/spanner` version: @google-cloud/spanner@4.8.0

#### Steps to reproduce

```
const { Spanner } = require("@google-cloud/spanner");
const fs = require("fs");
const util = require('util');
const stream = require('stream');
const through = require('through');

const pipeline = util.promisify(stream.pipeline);

const spanner = {
  "instanceId": "..",
  "databaseId": "..",
  "projectId": ".."
};
const filePath = "./users.data.txt";
const keyFilename = "service-account.json";

const spannerDatabase = new Spanner({
  projectId: spanner.projectId,
  keyFilename,
})
  .instance(spanner.instanceId)
  .database(spanner.databaseId);

const spannerStream = spannerDatabase.runStream({
  sql: 'select * from user',
  json: true,
  gaxOptions: {
    pageSize: 1000,
  },
});

const run = async () => {
  await pipeline(
    spannerStream,
    through(function (data) {
      return this.queue(`${JSON.stringify(data)}\n`);
    }),
    fs.createWriteStream(filePath)
  );

  process.exit(0);
};

run();
```
The stream consume a lot of memory (RAM) (some time it's take 600 Mb)
The `gaxOptions` (pageSize, maxResults, retry) doesn't work.

Thanks!

olavloite:
The request stream should be paused if the downstream is indicating that it cannot handle any more data at the moment. The request stream should be resumed once the downstream does accept data again. This reduces memory consumption and potentially out of memory errors when
a result stream is piped into a slow writer.

Fixes #934

### DIFF
--- a/src/partial-result-stream.ts
+++ b/src/partial-result-stream.ts
@@ -231,6 +231,7 @@ export class PartialResultStream extends Transform implements ResultEvents {
             `Stream is still not ready to receive data after ${this._numPushFailed} attempts to resume.`
           )
         );
+        return;
       }
       setTimeout(() => {
         const nextTimeout = Math.min(timeout * 2, 1024);
--- a/src/partial-result-stream.ts
+++ b/src/partial-result-stream.ts
@@ -136,11 +136,13 @@ export class PartialResultStream extends Transform implements ResultEvents {
   private _options: RowOptions;
   private _pendingValue?: p.IValue;
   private _values: p.IValue[];
-  constructor(options = {}) {
+  private _requestStream: NodeJS.ReadWriteStream;
+  constructor(options = {}, requestStream: NodeJS.ReadWriteStream) {
     super({objectMode: true});
 
     this._destroyed = false;
     this._options = options;
+    this._requestStream = requestStream;
     this._values = [];
   }
   /**
@@ -187,12 +189,39 @@ export class PartialResultStream extends Transform implements ResultEvents {
         .fields as google.spanner.v1.StructType.Field[];
     }
 
+    let res = true;
     if (!is.empty(chunk.values)) {
-      this._addChunk(chunk);
+      res = this._addChunk(chunk);
     }
 
-    next();
+    if (res || !this._requestStream) {
+      next();
+    } else {
+      // Wait a little before we push any more data into the pipeline as a
+      // component downstream has indicated that a break is needed. Pause the
+      // request stream to prevent it from filling up the buffer while we are
+      // waiting.
+      setTimeout(() => {
+        this._tryResume(next, 4);
+      }, 2);
+    }
   }
+
+  private _tryResume(next: Function, timeout: number) {
+    // Try to push an empty chunk to check whether more data can be accepted.
+    if (this.push(undefined)) {
+      this._requestStream!.resume();
+      next();
+    } else {
+      // Downstream returned false indicating that it is still not ready for
+      // more data.
+      setTimeout(() => {
+        const nextTimeout = Math.min(timeout * 2, 1024);
+        this._tryResume(next, nextTimeout);
+      }, timeout);
+    }
+  }
+
   /**
    * Manages any chunked values.
    *
@@ -200,7 +229,7 @@ export class PartialResultStream extends Transform implements ResultEvents {
    *
    * @param {object} chunk The partial result set.
    */
-  private _addChunk(chunk: google.spanner.v1.PartialResultSet): void {
+  private _addChunk(chunk: google.spanner.v1.PartialResultSet): boolean {
     const values: Value[] = chunk.values.map(GrpcService.decodeValue_);
 
     // If we have a chunk to merge, merge the values now.
@@ -223,7 +252,14 @@ export class PartialResultStream extends Transform implements ResultEvents {
       this._pendingValue = values.pop();
     }
 
-    values.forEach(value => this._addValue(value));
+    let res = true;
+    values.forEach(value => {
+      res = this._addValue(value) && res;
+      if (!res && !this._requestStream.isPaused()) {
+        this._requestStream.pause();
+      }
+    });
+    return res;
   }
   /**
    * Manages complete values, pushing a completed row into the stream once all
@@ -233,13 +269,13 @@ export class PartialResultStream extends Transform implements ResultEvents {
    *
    * @param {*} value The complete value.
    */
-  private _addValue(value: Value): void {
+  private _addValue(value: Value): boolean {
     const values = this._values;
 
     values.push(value);
 
     if (values.length !== this._fields.length) {
-      return;
+      return true;
     }
 
     this._values = [];
@@ -247,11 +283,10 @@ export class PartialResultStream extends Transform implements ResultEvents {
     const row: Row = this._createRow(values);
 
     if (this._options.json) {
-      this.push(row.toJSON(this._options.jsonOptions));
-      return;
+      return this.push(row.toJSON(this._options.jsonOptions));
     }
 
-    this.push(row);
+    return this.push(row);
   }
   /**
    * Converts an array of values into a row.
@@ -373,7 +408,8 @@ export function partialResultStream(
   // mergeStream allows multiple streams to be connected into one. This is good;
   // if we need to retry a request and pipe more data to the user's stream.
   const requestsStream = mergeStream();
-  const userStream = streamEvents(new PartialResultStream(options));
+  const partialRSStream = new PartialResultStream(options, requestsStream);
+  const userStream = streamEvents(partialRSStream);
   const batchAndSplitOnTokenStream = checkpointStream.obj({
     maxQueued: 10,
     isCheckpointFn: (row: google.spanner.v1.PartialResultSet): boolean => {
--- a/src/partial-result-stream.ts
+++ b/src/partial-result-stream.ts
@@ -43,10 +43,16 @@ interface RequestFunction {
  * @property {boolean} [json=false] Indicates if the Row objects should be
  *     formatted into JSON.
  * @property {JSONOptions} [jsonOptions] JSON options.
+ * @property {number} [maxResumeRetries=20] The maximum number of times that the
+ *     stream will retry to push data downstream, when the downstream indicates
+ *     that it is not ready for any more data. Increase this value if you
+ *     experience 'Stream is still not ready to receive data' errors as a
+ *     result of a slow writer in your receiving stream.
  */
 export interface RowOptions {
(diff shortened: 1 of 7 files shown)


skuruppu:
This change looks good to me @olavloite. Thanks for working on this. I can't say I have the expertise to review it properly. @bcoe could I kindly ask you to take a look to see if this is reasonable?

skuruppu:
I'm also very confused by these test failures that are happening on all PRs. I can't reproduce it locally.

olavloite:
I've been doing some additional testing with and without this fix using the function below. It does a select from a Spanner database, transforms the rows into JSON strings and writes that to a file. There is also a custom transformer in the pipeline that artificially slows the write progress every 50 rows to simulate a slow flush.

Running this script with/without the change in this PR on mid-sized (87MB) and a huge (2GB) result set gives the following results for maximum memory usage:

### Normal result set (87MB)
* Without this fix: 324MB
* With this fix: 330MB

### Huge result set (2GB)
* Without this fix: out-of-memory error (max heap size was set to 1.5GB)
* With this fix: 918MB

The exact effect of this fix will depend a lot on multiple factors:
* The size of the result set that is streamed.
* The size of the individual partial result sets that Cloud Spanner generates.
* The delay in the pipeline.

Larger values for any of the above will result in higher memory usage in all cases, and without this fix it can cause the entire result set to be loaded into memory.

### Without slow flush
It should also be noted that streaming result sets will in most cases be extremely efficient. Running the same test without the simulated slow flushes show the same results both with and without this fix. It also shows the effect of how Spanner chunks the result set. The memory consumption when streaming the huge result set is a lot lower than when streaming the normal result set (test results are equal with/without this fix):

* Normal result set: 299MB
* Huge result set: 33MB

### Test script
```javascript
async function queryWithMemUsage(instanceId, databaseId, projectId) {
  // Imports the Google Cloud client library

  // eslint-disable-next-line node/no-unpublished-require
  // const {Spanner} = require('../build/src');
  const fs = require('fs');
  const stream = require('stream');
  const util = require('util');
  // eslint-disable-next-line node/no-unsupported-features/node-builtins
  const pipeline = util.promisify(stream.pipeline);

  // Creates a client
  const spanner = new Spanner({
    projectId: projectId,
  });

  // Gets a reference to a Cloud Spanner instance and database
  const instance = spanner.instance(instanceId);
  const database = instance.database(databaseId);

  const query = {
    sql: `SELECT *
          FROM TableWithAllColumnTypes
          ORDER BY ColInt64`,
  };

  let count = 0;
  let maxMemMeasured = 0;
  const fileStream = fs.createWriteStream('/home/loite/rs.txt');
  const rs = database.runStream(query);

  console.time('process result set');
  // eslint-disable-next-line node/no-unsupported-features/node-builtins
  await pipeline(
    rs,
    new stream.Transform({
      objectMode: true,
      highWaterMark: 100,
      transform(chunk, encoding, callback) {
        count++;
        if (count % 100 === 0) {
          console.log(`Processed ${count} rows so far`);
          global.gc();
          const used = process.memoryUsage().heapUsed / 1024 / 1024;
          const memUsed = Math.round(used * 100) / 100;
          console.log(`Current mem usage: ${memUsed} MB`);
          maxMemMeasured = Math.max(maxMemMeasured, memUsed);
        }
        this.push(`${JSON.stringify(chunk.toJSON({wrapNumbers: true}))}\n`);
        callback();
      },
    }),
    // Create an artificially slow transformer to simulate network latency.
    new stream.Transform({
      highWaterMark: 100,
      transform(chunk, encoding, callback) {
        // Simulate a slow flush every 50 records.
        if (count % 50 === 0) {
          setTimeout(() => {
            this.push(chunk, encoding);
            callback();
          }, Math.random() * 200 + 100);
        } else {
          this.push(chunk, encoding);
          callback();
        }
      },
    }),
    fileStream
  );
  console.timeEnd('process result set');
  console.log(`Max memory used: ${maxMemMeasured} MB`);
  console.log('Finished writing file');
  await database.close();
}
```

stephenplusplus:
@olavloite this is pretty great. We've been carrying around the subtle issue where we have a single data event that needs to be split apart and forwarded to the next stream as multiple data events. That's the core issue we're attacking here, right?

One thing I noticed-- it looks like the transform stream would indefinitely try if the consumer never becomes ready. We probably want a cap on the maximum amount of attempts to avoid that.

I have a module "split-array-stream" which currently has the same problem. We use split-array-stream throughout various libraries, maybe even this one, although it looks like not from this file. I haven't merged this change yet, but would [this class](https://github.com/stephenplusplus/split-array-stream/blob/d921995fb4ea187122bf83e202cea7e1d1ff67ed/src/index.ts) either be plug-and-playable here, or be useful in some way to incorporate in PartialResultStream? The description of this PR shows how it can be used: https://github.com/stephenplusplus/split-array-stream/pull/4

Note that it's not merged and released, as it has not yet had a formal review. However, at the time, I had put it through similar tests as you did for this change. Since it seems like we're attacking the same issue, I thought it could be worth checking out.

Let me know what you think!

olavloite:
@stephenplusplus 
Thanks for having a look at this.

> We've been carrying around the subtle issue where we have a single data event that needs to be split apart and forwarded to the next stream as multiple data events. That's the core issue we're attacking here, right?

Correct. Cloud Spanner returns a stream of `PartialResultSets` which each contains a set of rows. The `PartialResultSets` are split into the individual rows which are forwarded to the next stream.

> One thing I noticed-- it looks like the transform stream would indefinitely try if the consumer never becomes ready. We probably want a cap on the maximum amount of attempts to avoid that.

Good point. I'll add an escape for that possibility.

Regarding `split-array-stream`: That's interesting. And also quite a bit to digest to really understand what it does :-) I'll look into it.

It does however seem that it's solving much of the same problem that we are having here, but there's one thing that might be different (or that I'm missing): During my testing of the Spanner client library without this PR I was able to make it go out of memory. That seems to have been caused by request stream that kept pushing `PartialResultSets` into the `PartialResultSetStream`. To prevent that, this change also explicitly pauses and resumes the request stream as needed. Does the `split-array-stream` also support that, either in this way or some other way?

olavloite:
@stephenplusplus Would you mind taking a second look at this?

src/partial-result-stream.ts
```diff
@@ -187,20 +196,57 @@ export class PartialResultStream extends Transform implements ResultEvents {
...
     }
 
+    let res = true;
     if (!is.empty(chunk.values)) {
-      this._addChunk(chunk);
+      res = this._addChunk(chunk);
     }
 
-    next();
+    if (res || !this._requestStream) {
```
stephenplusplus, line 204: In what scenario will there not be a `_requestStream` defined?

src/partial-result-stream.ts
```diff
@@ -187,20 +196,57 @@ export class PartialResultStream extends Transform implements ResultEvents {
...
+      next();
+    } else {
+      // Wait a little before we push any more data into the pipeline as a
+      // component downstream has indicated that a break is needed. Pause the
+      // request stream to prevent it from filling up the buffer while we are
+      // waiting.
+      setTimeout(() => {
```
stephenplusplus, line 211: `2` and `4` are a bit magical. Code comments to explain why those numbers were chosen or descriptive variable names would be helpful to make sure someone doesn't screw it up in the future :)

src/partial-result-stream.ts
```diff
@@ -187,20 +196,57 @@ export class PartialResultStream extends Transform implements ResultEvents {
...
+      this._numPushFailed = 0;
+      this._requestStream!.resume();
+      next();
+    } else {
+      // Downstream returned false indicating that it is still not ready for
+      // more data.
+      this._numPushFailed++;
+      setTimeout(() => {
+        const nextTimeout = Math.min(timeout * 2, 1024);
+        if (this._numPushFailed === this._options.maxResumeRetries) {
```
stephenplusplus, line 229: We might as well bump this out of the setTimeout to fail as quick as possible when we know we're done trying.

stephenplusplus:
@olavloite looks good to me. Just a thought that I know would be inconvenient to implement, and potentially not worth it because of that-- after this change, the stream philosophy of "each stream doesn't concern itself with another stream" is broken when we pass `PartialResultStream` needs to know about `requestStream`.

Possibly a way around this would be to have the PRS emit events to indicate it needs a break, and concerned streams could react as a result.

Something like:

```diff
values.forEach(value => {
  res = this._addValue(value) && res;
  if (!res && !this._requestStream.isPaused()) {
-    this._requestStream.pause();
+    this.emit('paused');
  }
});
```
```diff
requestsStream
  .pipe(batchAndSplitOnTokenStream)
  // If we get this error, the checkpoint stream has flushed any rows
  // it had queued. We can now destroy the user's stream, as our retry
  // attempts are over.
  .on('error', (err: Error) => userStream.destroy(err))
  .on('checkpoint', (row: google.spanner.v1.PartialResultSet) => {
    lastResumeToken = row.resumeToken;
  })
  .pipe(userStream)
+  .on('paused', () => {
+    requestsStream.pause();
+  })
```

In fact, writing that made me realize that should work by default, shouldn't it? The only reason we don't get built-in backpressure is because we do the one-to-many split of data events. But now that we push each data event singularly, you should be able to pause the PRS stream itself and have the streams before it react properly automatically.

src/partial-result-stream.ts
olavloite: Good point, that was a left-over from a previous version of the code where it could have been missing, but that is no longer possible.

src/partial-result-stream.ts
olavloite: Added a variable and a comment to clarify it.

src/partial-result-stream.ts
olavloite: Erm... Yeah, that was quite useless to have it inside the `setTimeout`. I've moved it out.

olavloite:
> Just a thought that I know would be inconvenient to implement, and potentially not worth it because of that-- after this change, the stream philosophy of "each stream doesn't concern itself with another stream" is broken when we pass PartialResultStream needs to know about requestStream.

I like the idea of this. It does make it more idiomatic while the end result is the same. We do need two events, though: `paused` and `resumed`. The implementation is straightforward.

stephenplusplus:
Thanks for doing that!

PR #955, opened 2020-05-07
fix: delete old instances then create new instance

### LINKED ISSUE #954: samples tests can't run due to old instances not being cleaned up
We've been getting errors like "Error: Project 1046198160504 cannot add 1 nodes in region us-central1." This happens when the project has run out of nodes to create new instances.

The solution might be to try to delete old instances before creating new ones.

skuruppu:

Fixes #954  🦕

### DIFF
--- a/samples/system-test/spanner.test.js
+++ b/samples/system-test/spanner.test.js
@@ -130,16 +130,6 @@ describe('Spanner', () => {
     }
 
     if (!INSTANCE_ALREADY_EXISTS) {
-      const [, operation] = await instance.create({
-        config: 'regional-us-central1',
-        nodes: 1,
-        labels: {
-          'gcloud-sample-tests': 'true',
-        },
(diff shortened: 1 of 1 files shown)


skuruppu:
Totally not tested yet, let's see if this runs.

hengfengli:
I don't think this would work. `INSTANCE_ID` is generated by `date` and I guess yesterday's instance was not deleted. All the deleting instances calls use the instance ID based on today not yesterday or past days. 

```ts
const INSTANCE_ID = process.env.SPANNERTEST_INSTANCE || `test-instance-${date}`;
const INSTANCE_ALREADY_EXISTS = !!process.env.SPANNERTEST_INSTANCE;
```

skuruppu:
It worked! Except for the fact that a backup couldn't be completed within the test timeout period :(

PR #959, opened 2020-05-13
chore: improve system-tests

AVaksman:
- Re-use initially created `database` to lighten up the rate of Administrative requests
- Improve resources clean up to avoid exceeding nodes quota for a project

### DIFF
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -3499,7 +3499,7 @@ describe('Spanner', () => {
             });
           }
 
-          table.read(query, (err, rows) => {
+          table.read(query as ReadRequest, (err, rows) => {
             test.assertions(err, rows);
             done();
           });
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -76,10 +76,6 @@ describe('Spanner', () => {
       await operation.promise();
       RESOURCES_TO_CLEAN.push(instance);
     } else {
-      // Update 'created' label timestamp on an existing instance
-      const [meta] = await instance.getMetadata();
-      const labels = Object.assign({}, meta.labels, {created: CURRENT_TIME});
-      await instance.setMetadata({labels});
       console.log(
         `Not creating temp instance, using + ${instance.formattedName_}...`
       );
@@ -932,9 +928,12 @@ describe('Spanner', () => {
 
           DATABASE.getSchema((err, statements) => {
             assert.ifError(err);
-            assert.strictEqual(
-              replaceNewLinesAndSpacing(statements![1]),
-              replaceNewLinesAndSpacing(createTableStatement)
+            assert.ok(
+              statements!.some(
+                s =>
+                  replaceNewLinesAndSpacing(s) ===
+                  replaceNewLinesAndSpacing(createTableStatement)
+              )
             );
             done();
           });
@@ -3583,7 +3582,7 @@ describe('Spanner', () => {
     const table = DATABASE.table(TABLE_NAME);
 
     it('should insert and query a row', done => {
-      const id = generateName('id0');
+      const id = generateName('id');
       const name = generateName('name');
 
       table.insert(
@@ -3594,17 +3593,17 @@ describe('Spanner', () => {
         err => {
           assert.ifError(err);
 
-          DATABASE.run(
-            `SELECT * FROM ${TABLE_NAME}  ORDER BY SingerId`,
-            (err, rows) => {
-              assert.ifError(err);
-              assert.deepStrictEqual(rows!.pop()!.toJSON(), {
-                SingerId: id,
-                Name: name,
-              });
-              done();
-            }
-          );
+          DATABASE.run(`SELECT * FROM ${TABLE_NAME}`, (err, rows) => {
+            assert.ifError(err);
+            assert.ok(
+              rows!.some(
+                r =>
+                  JSON.stringify(r.toJSON()) ===
+                  JSON.stringify({SingerId: id, Name: name})
+              )
+            );
+            done();
+          });
         }
       );
     });
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -110,7 +110,15 @@ describe('Spanner', () => {
         ).map(instance => instance.delete())
       );
     } else {
-      await Promise.all(RESOURCES_TO_CLEAN.map(resource => resource.delete()));
+      /**
+       * Limit the number of concurrent 'Administrative requests per minute'
+       * Not to exceed quota
+       * @see {@link https://cloud.google.com/spanner/quotas#administrative_limits}
+       */
+      const limit = pLimit(5);
+      await Promise.all(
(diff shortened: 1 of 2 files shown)


system-test/spanner.ts
```diff
@@ -1002,15 +999,7 @@ describe('Spanner', () => {
...
-        schema: `
-              CREATE TABLE Singers (
-                SingerId STRING(1024) NOT NULL,
-                Name STRING(1024),
-              ) PRIMARY KEY(SingerId)`,
-      });
-      await operation.promise();
+      database1 = DATABASE;
 
       await database1.table('Singers').insert({
```
stephenplusplus, line 1004: We should have a global `TABLE_NAME` defined so we don't have to assume this name exists.

system-test/spanner.ts
```diff
@@ -1020,6 +1009,7 @@ describe('Spanner', () => {
       // Create a second database since only one pending backup can be created
       // per database.
       database2 = instance.database(generateName('database'));
+      RESOURCES_TO_CLEAN.push(database2);
```
stephenplusplus, line 1012: Could you move these pushes to `RESOURCES_TO_CLEAN` to after the creation operation resolves? I liked how you did that in the global before hook. It makes sense to continue the trend.

system-test/spanner.ts
```diff
@@ -1702,8 +1663,8 @@ describe('Spanner', () => {
         err => {
           assert.ifError(err);
 
-          database.run(
-            'SELECT * FROM Singers ORDER BY SingerId',
+          DATABASE.run(
+            'SELECT * FROM SingersTables ORDER BY SingerId',
```
stephenplusplus, line 1667: Could we create a "TABLE_NAME" for this describe block too (and any others that work off of a unique table)? Now that the Database instance is being re-used and hosting multiple tables, it would be an easy mistake for someone to add a new test in a particular describe block and query against the wrong table.

system-test/spanner.ts
```diff
@@ -3606,22 +3565,10 @@ describe('Spanner', () => {
...
-              Name STRING(1024),
-            ) PRIMARY KEY(SingerId)`,
-      });
-      await operation.promise();
-    });
+    const table = DATABASE.table('Singers');
 
     it('should insert and query a row', done => {
-      const id = generateName('id');
+      const id = generateName('id0');
```
stephenplusplus, line 3571: Why id0?

system-test/spanner.ts
```diff
@@ -65,29 +65,45 @@ describe('Spanner', () => {
...
     if (generateInstanceForTest) {
       await deleteOldTestInstances();
       const [, operation] = await instance.create(INSTANCE_CONFIG);
       await operation.promise();
+      RESOURCES_TO_CLEAN.push(instance);
     } else {
+      // Update 'created' label timestamp on an existing instance
+      const [meta] = await instance.getMetadata();
+      const labels = Object.assign({}, meta.labels, {created: CURRENT_TIME});
+      await instance.setMetadata({labels});
```
stephenplusplus, line 82: Why add the "created" label?

system-test/spanner.ts
```diff
@@ -39,7 +39,7 @@ import CreateBackupMetadata = google.spanner.admin.database.v1.CreateBackupMetad
 
 const PREFIX = 'gcloud-tests-';
 const RUN_ID = shortUUID();
-const LABEL = `gcloud-tests-${RUN_ID}`;
+const LABEL = `gcloud-system-tests-${RUN_ID}`;
```
stephenplusplus, line 42: While we're changing this, could we go more distinct with something like "node-spanner-systests-"?

system-test/spanner.ts
```diff
@@ -921,10 +918,10 @@ describe('Spanner', () => {
             return str.replace(/\n\s*/g, '').replace(/\s+/g, ' ');
           }
 
-          database.getSchema((err, statements) => {
+          DATABASE.getSchema((err, statements) => {
             assert.ifError(err);
             assert.strictEqual(
-              replaceNewLinesAndSpacing(statements![0]),
+              replaceNewLinesAndSpacing(statements![1]),
```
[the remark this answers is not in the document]
stephenplusplus: Ah, this is tricky then. I don't think any test should depend on a previous execution to be kept in order. For example, if we add a test before this one that adds to the schema history, this would be bumped out of position. Other than creating a new database just for this test, will "statements.pop()" wrok?

system-test/spanner.ts
stephenplusplus: Thank you!

system-test/spanner.ts
stephenplusplus: I think the query just grabs the lot of them, and then we pop off the most recently inserted, or at least that's what I was thinking when I wrote the other test that uses `id1` and `id2`. Was it returning the row from the Backups test when it was still just "id"? We should find a way to be certain not to depend on the other test in any way.

system-test/spanner.ts
stephenplusplus: Thanks for the explanation! When I think of a user providing a pre-existing instance, I don't think we would ever want to delete it (unless they're just providing one that we created, then we make no such guarantee).

system-test/spanner.ts
stephenplusplus: That's cool. Thanks for doing that!

system-test/spanner.ts
AVaksman: Removed

system-test/spanner.ts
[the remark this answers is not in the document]
AVaksman: Adjusting it to check that the statement of interest does exist in the the `statements` array.

system-test/spanner.ts
AVaksman: Switched to check if the `row` on question is present in the `rows` array.

system-test/spanner.ts
AVaksman: Removed the logic, thanks!

system-test/spanner.ts
```diff
@@ -65,29 +65,45 @@ describe('Spanner', () => {
...
+            Name STRING(1024),
+          ) PRIMARY KEY(SingerId)`,
+    });
+    await operation.promise();
   });
 
   after(async () => {
     if (generateInstanceForTest) {
       await deleteTestInstances();
+    } else {
```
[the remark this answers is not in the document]
stephenplusplus: That makes sense! Would you mind leaving a comment in the code to explain why we exclude those?

Also, it might be cleaner to get it down to one filter & map:

```js
await Promise.all(
  RESOURCES_TO_CLEAN.filter(
    // No need to delete databases as we will delete the instance
    resource => !(resource instanceof Database)
  ).map(resource => resource.delete())
);
```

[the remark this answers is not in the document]
stephenplusplus: Good point. Adding in a "sort()" would probably be overkill, so no worries. Change looks good, thank you!

PR #963, opened 2020-05-18
feat: increase sessions in the pool in batches

olavloite:
Increase the number of sessions in the pool in batches of 25 (configurable) for better performance and less RPCs.

This also adds four benchmarks for to verify the working of the above change, and for comparison with other supported platforms. The simulated execution times and network latency is not intended to be 100% equal to a production situation, but to simulate a realistic and repeatable situation that can be used for multiple platforms.

### DIFF
--- a/src/session-pool.ts
+++ b/src/session-pool.ts
@@ -130,6 +130,8 @@ export interface SessionPoolInterface extends EventEmitter {
  *     any given time.
  * @property {number} [writes=0.0] Percentage of sessions to be pre-allocated as
  *     write sessions represented as a float.
+ * @property {number} [incStep=25] The number of new sessions to create when at
+ *     least one more session is needed.
  */
 export interface SessionPoolOptions {
   acquireTimeout?: number;
@@ -142,6 +144,7 @@ export interface SessionPoolOptions {
   maxIdle?: number;
   min?: number;
   writes?: number;
+  incStep?: number;
 }
 
 const DEFAULTS: SessionPoolOptions = {
@@ -155,6 +158,7 @@ const DEFAULTS: SessionPoolOptions = {
   maxIdle: 1,
   min: 0,
   writes: 0,
+  incStep: 25,
 };
 
 /**
@@ -615,7 +619,9 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    */
   _createSession(type: types): Promise<void> {
     const kind = type === types.ReadOnly ? 'reads' : 'writes';
-    const options = {[kind]: 1};
+    const options = {
+      [kind]: 1,
+    };
 
     return this._createSessions(options);
   }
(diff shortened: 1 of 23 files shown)


skuruppu:
Sorry @olavloite, I ran out of time to take a proper look at this. So it'll be next week before I get to this. Hope that it's ok.

src/session-pool.ts
```diff
@@ -813,9 +817,22 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       !this.isFull &&
       this._pending + this._pendingPrepare <= this._numWaiters
     ) {
+      let reads = this.options.incStep
+        ? this.options.incStep
+        : DEFAULTS.incStep!;
+      let writes = 0;
+      // Only create 1 write session at a time.
```
skuruppu, line 824: Is there a specific reason for this? Are we worried that we might overdo the number of write sessions and go over the write fraction?

olavloite:
> Thanks for this change @olavloite. I just had one minor question.
> 
> Another thought is the possibility of initializing the client with some sessions. As I understand it, default min sessions is 0 so I think we would wait for the first request before attempting to create sessions. Your change helps since at least now we won't be creating one session at a time when your default min sessions is zero. But maybe we can set min to something like 10 so the pool opens with something. Thoughts?

Yes, I do think it would be good to set a non-zero default value for `min` sessions. I've updated it to the suggested default of minimum 10 sessions. These sessions are created asynchronously when a new database is opened.

src/session-pool.ts
olavloite: The reasoning behind this is to keep it consistent with the current behavior of the node session pool, which treats the configured fraction for write sessions slightly different from what the Java and Go pools do. The basic working (at the moment) is:
1. Use the write fraction to determine the number of write sessions to create during the initial filling of the session pool. This means that if you configure `min=100` and `write=0.2` you'll end up with an initial session pool with 20 write sessions.
1. The write fraction is __not__ used to determine whether any further write sessions should be created when additional sessions are needed. Instead, write sessions are only created on-demand when the user application requests one, and there is none available in the pool. That is what is happening here.
1. A session that has been marked as a write session, either during the initial filling of the pool, or later because it was created on-demand, will remain a write session during the course of its lifetime. This means that the node session pool will maintain a number of write sessions, even if a write fraction of 0.0 (the default) has been configured, if the user application requires write sessions.

The last point is something we should consider a breaking change if we want to change it, but it is probably better not to change it at all. It is a very nice way of auto-configuring the number of write sessions in the pool.

#1031 further optimizes the way the pool maintains write sessions, but retains the behavior from point 3 above.

olavloite:
@skuruppu I figured out why the system tests (and emulator tests) were failing, and the reason behind it might also be a reason that it might not be a good idea to change the default for `min` to a non-zero value. The following lines of code is the normal way to create a new database:

```typescript
      database2 = instance.database(generateName('database'));
      const [, database2CreateOperation] = await database2.create({
        schema: `
        CREATE TABLE Albums (
          AlbumId STRING(1024) NOT NULL,
          AlbumTitle STRING(1024) NOT NULL,
          ) PRIMARY KEY(AlbumId)`,
      });
      await database2CreateOperation.promise();
```

The first line will however not only get a reference to the database, but it will also try to open a session pool for it. As the database at that moment does not yet exist, filling the session pool with sessions will fail. That is not a problem as long as the default for `min=0`, but once it is changed to a non-zero value, it will cause errors.

skuruppu:
@olavloite thanks for looking into changing the min sessions to be non-zero. That makes sense. The pool is initialized in the constructor but as you said, the database may not exist at that point.

So it makes sense to not change that here then.

I think we can get this working by moving the pool creation logic (conditionally) but that's for another time.

skuruppu:
I'd be happy for you to merge this in when you're ready.

PR #967, opened 2020-05-21
fix!: return API objects in streams

AVaksman:

Convert objects returned by streams to API objects.


skuruppu:
@olavloite since you've been looking into streams recently, would you be able to help review this?

olavloite:
@AVaksman Do I understand this correctly that the main feature change here is that these streams will now emit the data as Spanner (rich) client classes (e.g. `class Session`) instead of the generated gRPC objects (e.g. `interface ISession`)? Won't that break user code that currently relies on these streams returning the gRPC objects?

src/database.ts
```diff
@@ -1348,12 +1349,29 @@ class Database extends GrpcServiceObject {
...
+    const self = this;
+    const transform = function (
+      this: Transform,
+      chunk: databaseAdmin.spanner.v1.ISession,
+      enc: string,
+      callback: Function
+    ) {
+      const session = self.session(chunk.name!);
+      session.metadata = chunk;
+      this.push(session);
```
stephenplusplus, line 1362: I believe `callback(null, session)` would be our usual pattern.

AVaksman: This seems neater.
Thanks!

AVaksman:
> Do I understand this correctly that the main feature change here is that these streams will now emit the data as Spanner (rich) client classes (e.g. `class Session`) instead of the generated gRPC objects (e.g. `interface ISession`)? 

Yes, as streaming methods should emit same data as their callback cousins.

> Won't that break user code that currently relies on these streams returning the gRPC objects?

I believe this was missed in the previous PR #925 and this PR fixes it.

The documentation also sights the emitted data as Spanner (rich) client classes
```ts 
   * @returns {ReadableStream} A readable stream that emits {@link Session}
   *     instances.
---------------
   * @returns {ReadableStream} A readable stream that emits {@link Instance}
   *     instances.
---------------
   * @returns {ReadableStream} A readable stream that emits {@link Backup}
   *     instances.
---------------
   * @returns {ReadableStream} A readable stream that emits {@link Database}
   *     instances.
```

stephenplusplus:
> I believe this was missed in the previous PR #925 and this PR fixes it.

I would still call this breaking :( User code that worked will not work after this release.

AVaksman:
> I would still call this breaking :( User code that worked will not work after this release.

Adjusted the PR header
Thanks!

src/database.ts
```diff
@@ -1348,12 +1349,28 @@ class Database extends GrpcServiceObject {
...
+      chunk: databaseAdmin.spanner.v1.ISession,
+      enc: string,
+      callback: Function
+    ) {
+      const session = self.session(chunk.name!);
+      session.metadata = chunk;
+      callback(null, session);
+    };
+
+    return new pumpify.obj([
```
olavloite, line 1365: Sorry for the possibly stupid question here, but do we need to introduce `pumpify` for this? Would it not be just as easy (and one dependency less) to do something like the following:

```typescript
    const req = this.requestStream({
      client: 'SpannerClient',
      method: 'listSessionsStream',
      reqOpts,
      gaxOpts,
    })
      .on('error', e => req.destroy(e))
      .pipe(new Transform({objectMode: true, transform}));
    return req;
```

AVaksman: `pumpify` is used throughout the different clients as well ([bigtable](https://github.com/googleapis/nodejs-bigtable/blob/6d79c48a07111f6def752b636def22d17f748956/src/table.ts#L813), [storage](https://github.com/googleapis/nodejs-storage/blob/621e6f4723d48511c94f11cf0aa7d2bc20516b65/src/file.ts#L1738), [storage](https://github.com/googleapis/nodejs-storage/blob/621e6f4723d48511c94f11cf0aa7d2bc20516b65/src/file.ts#L1304)).

In case of using `pipe()` as `source.pipe(dest)` the source stream will not be destroyed if dest stream breaks, emits close,  or an error. `pumpify` ensures that source stream is destroyed.

olavloite: Sounds reasonable.

olavloite (approved):

JustinBeckwith:
@AVaksman @stephenplusplus @skuruppu  this PR is looking a little moldy :) What's the plan here?

AVaksman:
Just waiting for the next major release.

skuruppu:
The next major release will come in around a month's time. We're just gathering up all the breaking changes :)

PR #1049, opened 2020-05-27
feat: change default min sessions to 10

olavloite:
Sets the default for `min` in the session pool to 10.


olavloite:
Not needed.

PR #1072, opened 2020-06-03
fix: update DELETE samples to match docs

### LINKED ISSUE #1071: Update the spanner_delete_data code sample to have examples for all the delete methods
The [Deleting rows in a table](https://cloud.google.com/spanner/docs/modify-mutation-api#python_2) section states:

Each client library provides multiple ways to delete rows:
- Delete all the rows in a table.
- Delete a single row by specifying the key column values for the row.
- Delete a group of rows by creating a key range.
- Delete rows in an interleaved table by deleting the parent rows, if the interleaved table includes ON DELETE CASCADE in its schema definition.

Node.js (and most languages) don't give examples of all of these methods.

skuruppu:
[Docs](https://cloud.google.com/spanner/docs/modify-mutation-api#deleting_rows_in_a_table) list a set of dot points on the multiple ways to perform deletes. The updated samples now contain these examples.

Fixes #1071 🦕

### DIFF
--- a/samples/crud.js
+++ b/samples/crud.js
@@ -156,7 +156,7 @@ async function deleteData(instanceId, databaseId, projectId) {
     }
     try {
       const [rowCount] = await transaction.runUpdate({
-        sql: 'DELETE FROM Singers WHERE SingerId >= 3',
+        sql: 'DELETE FROM Singers WHERE SingerId >= 3 AND SingerId < 5',
       });
       console.log(`${rowCount} records deleted from Singers.`);
       await transaction.commit();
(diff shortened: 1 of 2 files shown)


samples/crud.js
```diff
@@ -134,19 +134,57 @@ async function deleteData(instanceId, databaseId, projectId) {
...
+  // Deletes remaining rows from the Singers table and the Albums table,
+  // because Albums table is defined with ON DELETE CASCADE.
+  database.runTransaction(async (err, transaction) => {
+    if (err) {
+      console.error(err);
+      return;
+    }
+    try {
+      const [rowCount] = await transaction.runUpdate({
+        sql: 'DELETE FROM Singers WHERE true',
```
olavloite, line 177: Maybe we should add a comment here on why the `WHERE true` clause is added to the statement.

skuruppu: Good idea. Done.

PR #1073, opened 2020-06-03
feat: add code sample for creating an instance

skuruppu:

### DIFF
--- a/samples/instance.js
+++ b/samples/instance.js
@@ -1,0 +1,71 @@
+/**
+ * Copyright 2020 Google LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


samples/instance.js
```diff
@@ -0,0 +1,81 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+async function createInstance(instanceId, configuration, nodes, projectId) {
```
larkee, line 18: Did we want to have users specify all variables? It does make more sense from a command line tool perspective. In Python and PHP, generally only the instance ID and database ID are specified. Is this different from Node?

Java only lets users specify instance ID. Does Java support the samples as a command line tool?

hengfengli: I think this depends on the style guide for each language. For example, in Go, the old style is pass a pre-created client into the function. But the new style suggests to initialize clients inside the code sample.

skuruppu: Node.js samples are part of a command line tool. But I also want the samples to be as consistent as possible across the languages. So I'm happy to hard-code the region and nodes given that the Java and Ruby samples only accept the instance and project arguments.

Issue #1078, opened 2020-06-05
Errors from encodeValue_ are not caught and generate a UnhandledPromiseRejectionWarning


WaldoJeffers:
#### Environment details

  - OS: MacOSX 10.14.6
  - Node.js version: v10.19.0
  - npm version: 6.13.4
  - `@google-cloud/spanner` version: 5.1.0

#### Bug description
Errors thrown by [common-grpc/service/encodeValue_](https://github.com/googleapis/nodejs-spanner/blob/master/src/common-grpc/service.ts#L285) like [this one](https://github.com/googleapis/nodejs-spanner/blob/master/src/common-grpc/service.ts#L335) (passing `undefined` as value for a named parameter) are never caught by this function's caller, resulting in Unhandled Promise Rejections. **Also, for some reason it seems to crash the underlying Node.js process.**

The error thrown in `encodeValue_` goes all the way to `SessionPool.getReadSession` where it fails to be caught:
- [encodeValue_](https://github.com/googleapis/nodejs-spanner/blob/master/src/common-grpc/service.ts#L285)
- [GrpcService.encodeValue_](https://github.com/googleapis/nodejs-spanner/blob/master/src/common-grpc/service.ts#L786)
- [codec.encode](https://github.com/googleapis/nodejs-spanner/blob/master/src/codec.ts#L346)
- [Snapshot.encodeParams](https://github.com/googleapis/nodejs-spanner/blob/master/src/transaction.ts#L1034)
- [Snapshot.runStream](https://github.com/googleapis/nodejs-spanner/blob/master/src/transaction.ts#L889)
- [Database.runStream](https://github.com/googleapis/nodejs-spanner/blob/master/src/database.ts#L2109)
- [SessionPool.getReadSession](https://github.com/googleapis/nodejs-spanner/blob/master/src/session-pool.ts#L429) <-- Probably where the error should be caught

#### Steps to reproduce
There are probably different ways to reproduce the error; all you have to do is make the `encodeValue_` helper throw. One way to achieve this is to provide an `undefined` value to a named parameter in a query
  1. Run a query using named parameters, but don't pass a value to the parameter
```js
const q = {
  json: true,
  params: { },
  sql: 'SELECT * FROM tableId WHERE namedParameter = @namedParameter',
};
const [rows] = await db.run(q); // This will lead to an Unhandled Promise rejection
  ```

So any attempt to catch the error will fail
```js
[...]

PR #1079, opened 2020-06-05
fix: handle errors thrown by success callback in getReadSession/getWriteSession

WaldoJeffers:

Fixes https://github.com/googleapis/nodejs-spanner/issues/1078 🦕

**Description**
Errors thrown by the `callback` executed in the "success" branch of the `then()` method are never caught by its second parameter; and since there are no outside `catch` block; it results in Unhandled Promise rejections.

I don't know if the correct fix is to move the second parameter provided to `then()` to a different `catch` block, or if it should be left as is while providing a different error callback to a different `catch` block (it depends on whether you want errors from both ` this._acquire` and its success callback to be handled by the same handler or not)

### DIFF
--- a/src/session-pool.ts
+++ b/src/session-pool.ts
@@ -425,10 +425,9 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @param {GetReadSessionCallback} callback The callback function.
    */
   getReadSession(callback: GetReadSessionCallback): void {
-    this._acquire(types.ReadOnly).then(
-      session => callback(null, session),
-      callback
-    );
+    this._acquire(types.ReadOnly)
+      .then(session => callback(null, session))
+      .catch(callback);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/session-pool.ts
```diff
@@ -425,10 +425,9 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @param {GetReadSessionCallback} callback The callback function.
    */
   getReadSession(callback: GetReadSessionCallback): void {
-    this._acquire(types.ReadOnly).then(
-      session => callback(null, session),
```
WaldoJeffers, line 429: Errors generated by `callback(null, session)` are not caught by the second parameter provided to `then()`. If they are intended to be caught by `callback`,  this must be done in a separate `catch()` block

stephenplusplus: I'm not very familiar with how this code is intended to work, but it seems we would want to execute `callback` once. If executing `callback` in the success case throws an error, calling it again with its own error seems like a round trip that we could skip. From your investigation, would there be another way to return the error when it happens, or is this the best way to handle it?

Thanks for all of your help!

WaldoJeffers: @stephenplusplus I am not familiar either with how it's supposed to work; so it might not be the best way to solve this issue. 

I decided to re-use the `callback` parameter because it was a function with error handling capabilities and it was available where the error occurred. 
Re-using `callback` possibly adds an extra round trip; but most Node.js callbacks handle potential errors first; before doing anything else; so I assume it wouldn't lead to too much code being run for nothing.

However, it might make more sense to handle errors generated by `callback` inside `callback` itself; but I wasn't sure where it was defined; and what was the best way to handle errors.

Maybe we should add a `try/catch` block [around this line ?](https://github.com/googleapis/nodejs-spanner/blob/23958ae48f49306cf38755831db091fef16998fb/src/database.ts#L2109)

AVaksman:
#1078 issue is closed by PR #1208.
Closing this PR

PR #1149, opened 2020-06-15
chore: retry flaky tests for exceed administrative requests per minute quota error

### LINKED ISSUE #958: Retry requests in tests where we run out of admin request quota
We get flaky test failure notifications regularly due to the following error (#944 and issues referencing it):

```
8 RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Administrative requests' and limit 'Administrative requests per minute' of service 'spanner.googleapis.com' for consumer 'project_number:<project>'.
Error: 8 RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Administrative requests' and limit 'Administrative requests per minute' of service 'spanner.googleapis.com' for consumer 'project_number:<project>'.
```

I requested a quota increase for the project but the Spanner team recommended that we "implement exponential back-off with extended total timeout and number of re-tries for admin operations".

I think this is reasonable to do in the system and samples tests, starting with the create instance and database requests.

AVaksman:

Related to #1089, #1090, #1091 issues
Fixes #958

### DIFF
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -32,7 +32,7 @@ import {
 } from '../src/transaction';
 import {Row} from '../src/partial-result-stream';
 import {GetDatabaseConfig} from '../src/database';
-import {grpc} from 'google-gax';
+import {grpc, CallOptions} from 'google-gax';
 import {google} from '../protos/protos';
 import CreateDatabaseMetadata = google.spanner.admin.database.v1.CreateDatabaseMetadata;
 import CreateBackupMetadata = google.spanner.admin.database.v1.CreateBackupMetadata;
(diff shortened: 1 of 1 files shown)


system-test/spanner.ts
```diff
@@ -44,6 +44,24 @@ const spanner = new Spanner({
   projectId: process.env.GCLOUD_PROJECT,
   apiEndpoint: process.env.API_ENDPOINT,
 });
+const RETRYOPTIONS: CallOptions = {
```
stephenplusplus, line 47: This is pretty trivial, but could this be renamed to `GAX_OPTIONS`? I'm thinking it might be more clear where it's used to understand the type of object we're providing.

AVaksman: Sure, done!

system-test/spanner.ts
```diff
@@ -1383,15 +1406,12 @@ describe('Spanner', () => {
...
+            ) PRIMARY KEY(SingerId)`,
+          GAX_OPTIONS
         )
         .then(onPromiseOperationComplete);
     });
 
-    after(() => {
-      return table.delete().then(onPromiseOperationComplete);
-    });
-
```
skuruppu, line 1394: I'm not sure I understand why this is deleted.

system-test/spanner.ts
```diff
@@ -44,6 +44,24 @@ const spanner = new Spanner({
   projectId: process.env.GCLOUD_PROJECT,
   apiEndpoint: process.env.API_ENDPOINT,
 });
+const GAX_OPTIONS: CallOptions = {
```
skuruppu, line 47: This looks great. Is it possible to add the same logic to the samples tests? The quota issue can happen in both types of tests.

system-test/spanner.ts
AVaksman: The table will be deleted together with the database in after all clean up section.
One less administrative call.

PR #1154, opened 2020-06-18
fix: chained promises to guarantee deletion order

### LINKED ISSUE #1147: samples tests fail
They fail :)   (#1146 is an example)

skuruppu:
Previously, in the delete code sample, due to the async callbacks given
to `runTransaction`, it was possible for the last delete query to be
executed before the second last delete query. The deletion order is
important so now using `runTransactionAsync` with the promises chained
to guarantee the deletion order.

Fixes #1147, #1095


skuruppu:
Closing in favor of #1156.

PR #1159, opened 2020-06-19
fix: race condition in "should transfer value from one record to another using DML statements within a transaction test"

AVaksman:

Currently `should transfer value from one record to another using DML statements within a transaction` test throws an error ([link](https://source.cloud.google.com/results/invocations/81f776c0-1e79-4c92-a3d5-59094a889c93/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-spanner%2Fnode10%2Fsamples-test/log)) and does not "transfer value from one record to another ..." as it supposed to.

### DIFF
--- a/samples/dml.js
+++ b/samples/dml.js
@@ -474,32 +474,46 @@ function writeWithTransactionUsingDml(instanceId, databaseId, projectId) {
         // Note: Cloud Spanner interprets Node.js numbers as FLOAT64s, so they
         // must be converted (back) to strings before being inserted as INT64s.
 
-        transaction.runUpdate({
-          sql: `UPDATE Albums SET MarketingBudget = @Budget
-            WHERE SingerId = 1 and AlbumId = 1`,
-          params: {
-            Budget: firstBudget.toString(),
-          },
-        });
-        transaction.runUpdate({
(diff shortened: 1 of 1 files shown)


samples/dml.js
```diff
@@ -474,32 +474,46 @@ function writeWithTransactionUsingDml(instanceId, databaseId, projectId) {
...
+        // Verify the accuracy of the new budgets.
+        const rows = results[0].map(row => row.toJSON());
+        if (
+          rows[0].MarketingBudget === firstBudget &&
+          rows[1].MarketingBudget === secondBudget
+        ) {
+          console.log(
+            `Successfully executed read-write transaction using DML to transfer ${transferAmount} from Album 2 to Album 1.`
+          );
+        }
```
skuruppu, line 512: We can't really have this part because this would make the Node.js sample inconsistent with that of the other languages. Is it ok if we remove it?

skuruppu: https://cloud.google.com/spanner/docs/getting-started/go#update-data for example.

AVaksman: Sure.

It was added to catch/verify that the current issue would have been caught.

Done!

PR #1163, opened 2020-06-25
feat: spanner NUMERIC support

skuruppu:


skuruppu:
I get this error when running the system tests:

```
node_modules/@google-cloud/spanner/build/src/codec.d.ts:1:21 - error TS7016: Could not find a declaration file for module 'big.js'. '/tmp/tmp-167842-GXX9y8Rl79kn/node_modules/big.js/big.js' implicitly has an 'any' type.
  Try `npm install @types/big.js` if it exists or add a new declaration (.d.ts) file containing `declare module 'big.js';`

1 import { Big } from 'big.js';
                      ~~~~~~~~


Found 1 error.
```

I don't yet know how to resolve this.

package.json
```diff
@@ -55,6 +55,7 @@
     "@google-cloud/promisify": "^2.0.0",
     "@types/stack-trace": "0.0.29",
```
olavloite, line 56: Adding `"@types/big.js": "^4.0.5",` here should fix the problem where the system tests are complaining about missing types for `big.js`.

protos/google/spanner/v1/type.proto
```diff
@@ -74,6 +74,17 @@ enum TypeCode {
   // Encoded as `list`, where list element `i` is represented according
   // to [struct_type.fields[i]][google.spanner.v1.StructType.fields].
   STRUCT = 9;
+
+  // Encoded as `string`, in decimal format or scientific notation format.
+  // <br>Decimal format:
+  // <br>`[+-]Digits[.[Digits]]` or
```
olavloite, line 80: One minor problem with using the `big.js` library for this, is that it does not seem to support exactly the same formats as Cloud Spanner. Most notably, Cloud Spanner allows a leading + sign for positive numbers, while `big.js` seems not to accept that. As long as Cloud Spanner will return positive values without a leading + sign, it should not really be a big problem, but it will mean that a user cannot in all cases use the same notation in the Node client as what could have been used directly with Cloud Spanner.

skuruppu: @larkee FYI.

Ah ok, thanks @olavloite for pointing this out. I did not think about this. AFAIK we chose this to be consistent with BigQuery since the backend representation is the same in both Spanner and BigQuery. I don't think Spanner will actually return values with a leading + sign but I will check with the backend team.

skuruppu: I verified with the backend team and they said the backend will not prefix positive values with a '+'.

package.json
skuruppu: Thanks heaps @olavloite. That did indeed fix the issue.

skuruppu:
PTAL @olavloite. I've fixed most of the issues and now also includes ITs. They do still fail against the prod backend though.

system-test/spanner.ts
```diff
@@ -527,6 +557,69 @@ describe('Spanner', () => {
...
+        insert({NumericArray: [null]}, (err, row) => {
+          assert.ifError(err);
+          assert.deepStrictEqual(row.toJSON().NumericArray, [null]);
+          done();
+        });
+      });
+
+      it('should write numeric array values', done => {
+        const values = [
+          Spanner.numeric('-99999999999999999999999999999.999999999'),
```
olavloite, line 610: nit: could we also include a null value in the array to ensure that those are also encoded and decoded correctly?

olavloite (approved):
LGTM, with a small question on one of the system tests.

system-test/spanner.ts
skuruppu: I'm happy to do so, but just curious if you think the 'should write null numeric array values' test just above this one is not sufficient. For some reason all the other 'should write <type> array values' tests are like this where they only include non-null values in the test.

olavloite: Sorry, you're right. I misread the above test case as 'write a null array', and not as 'write an array containing one null element'.

olavloite:
> LGTM, with a small question on one of the system tests.

Never mind my question, i misread the test before the one I was asking a question about.

skuruppu:
@olavloite PTAL since I added the code samples based on your work for Java (https://github.com/GoogleCloudPlatform/java-docs-samples/pull/3482).

samples/datatypes.js
```diff
@@ -607,6 +611,64 @@ async function queryWithTimestamp(instanceId, databaseId, projectId) {
...
+  const exampleNumeric = Spanner.numeric('300000');
+
+  const query = {
+    sql: `SELECT VenueId, VenueName, Revenue FROM Venues
+            WHERE Revenue >= @revenue`,
+    params: {
+      revenue: exampleNumeric,
+    },
+    types: {
+      venueName: fieldType,
```
olavloite, line 648: Shouldn't this be `revenue: fieldType`?

olavloite (approved):
Looks generally good to me, but a small question on the parameter type that is being set in the query.

samples/datatypes.js
skuruppu: Good catch. Fixed now.

PR #1167, opened 2020-07-06
perf: increase default min sessions to 25

### LINKED ISSUE #1160: Allow a database client to be created with min > 0 sessions
In https://github.com/googleapis/nodejs-spanner/pull/963#issuecomment-634860373, @olavloite said that if a client is created with min sessions > 0, the client creation could fail. This is because we create the session pool before creating the database connection, and the database may not exist yet.

This means that we default to min=0 sessions so the client starts up with no sessions prepared and the first create session request is sent when the first request is received. In #963, a change was made to create sessions in batches of 10 (using BatchCreateSessions). So at least, the extra latency will only be experienced in fewer requests.

Ideally we should be able to set min > 0 sessions and have them be initialized before the first request.

olavloite:
Increase the default setting of `SessionPoolOptions.min` to 25.

Fixes #1160


src/session-pool.ts
```diff
@@ -217,6 +217,36 @@ export function isSessionNotFoundError(
   );
 }
 
+/**
+ * Checks whether the given error is a 'Database not found' error.
+ * @param error the error to check
+ * @return true if the error is a 'Database not found' error, and otherwise false.
```
skuruppu, line 223:
```suggestion
 * @param {Error} error The error to check.
 * @return {boolean} True if the error is a 'Database not found' error, and otherwise false.
```

src/session-pool.ts
```diff
@@ -217,6 +217,36 @@ export function isSessionNotFoundError(
...
+    error !== undefined &&
+    error.code === grpc.status.NOT_FOUND &&
+    error.message.includes('Database not found')
+  );
+}
+
+/**
+ * Checks whether the given error is an 'Instance not found' error.
+ * @param error the error to check
+ * @return true if the error is an 'Instance not found' error, and otherwise false.
```
skuruppu, line 238:
```suggestion
 * @param {Error} error The error to check.
 * @return {boolean} True if the error is an 'Instance not found' error, and otherwise false.
```

skuruppu (approved):
Thanks so much for this change.

PR #1171, opened 2020-07-07
test: change system-test setup against emulator

skuruppu:
Use Docker image instead of manually downloading the emulator binary and running it.


.github/workflows/system-tests-against-emulator.yaml
```diff
@@ -7,11 +7,22 @@ name: system-tests-against-emulator
 jobs:
   test:
     runs-on: ubuntu-latest
+
+    services:
+      emulator:
+        image: gcr.io/cloud-spanner-emulator/emulator:0.8.0
```
JustinBeckwith, line 13: Cool!  Curious more than anything - do you need to pin to a specific version of the emulator?  How are you planning to keep it up to date?

JustinBeckwith (approved):
This is so much better than a shell script and it makes me so happy 🤗

.github/workflows/system-tests-against-emulator.yaml
skuruppu: That's a good question. For a while I think we had to explicitly run against v0.8.0 so that we can pick up some new feature implementations. But now it seems that v0.8.0 and 'latest' are the same thing so I think I can change this. Let me try.

skuruppu: \o/ it works!

PR #1175, opened 2020-07-07
chore: update samples cleanup

AVaksman:


sofisl (approved):

skuruppu (approved):

samples/system-test/spanner.test.js
```diff
@@ -69,7 +69,7 @@ const GAX_OPTIONS = {
 async function deleteStaleInstances() {
   const [instances] = await spanner.getInstances({
     filter:
-      '(labels.gcloud-sample-tests:true) OR (labels.cloud_spanner_samples:true)',
+      '(labels.node-sample-tests:true) OR (labels.cloud_spanner_samples:true)',
```
stephenplusplus, line 72: Can we combine the variables with the string here to make sure they’re consistent?

stephenplusplus (changes requested):

samples/system-test/spanner.test.js
AVaksman: done with `'node-sample-tests'`, the `'cloud_spanner_samples'` label is from the [sample](https://github.com/googleapis/nodejs-spanner/blob/e40abaffadce6bd92f5ecbac7d054d4a1ca1c928/samples/instance.js#L44) itself.

stephenplusplus (approved):

samples/system-test/spanner.test.js
```diff
@@ -67,9 +67,9 @@ const GAX_OPTIONS = {
 };
 
 async function deleteStaleInstances() {
+  console.log(`(labels.${LABEL}:true) OR (labels.cloud_spanner_samples:true)`);
```
stephenplusplus, line 70: Can we get rid of this?

stephenplusplus (changes requested):

samples/system-test/spanner.test.js
AVaksman: Yes, of course.
Done!

stephenplusplus (approved):

PR #1187, opened 2020-07-17
refactor(test): use exponential backoff and retry a sample test

AVaksman:
Retry a failing sample test, that causes other dependent tests to fail as well.

Should fix failing samples tests #1182 #1183 #1184 #1185 #1191

### DIFF
--- a/samples/system-test/spanner.test.js
+++ b/samples/system-test/spanner.test.js
@@ -68,9 +68,10 @@ const GAX_OPTIONS = {
 
 const delay = async test => {
   const retries = test.currentRetry();
-  if (retries === 0) return; // no retry on the first failure.
-  // see: https://cloud.google.com/storage/docs/exponential-backoff:
-  const ms = Math.pow(2, retries) * 250 + Math.random() * 1000;
+  // No retry on the first failure.
+  if (retries === 0) return;
+  // See: https://cloud.google.com/storage/docs/exponential-backoff
+  const ms = Math.pow(2, retries) + Math.random() * 1000;
   return new Promise(done => {
     console.info(`retrying "${test.title}" in ${ms}ms`);
     setTimeout(done, ms);
@@ -306,7 +307,8 @@ describe('Spanner', () => {
   // create_storing_index
   it('should create a storing index in an example table', async function () {
     this.retries(5);
-    await delay(this.test); // delay the start of the test, if this is a retry.
+    // delay the start of the test, if this is a retry.
+    await delay(this.test);
 
     const output = execSync(
       `${indexingCmd} createStoringIndex ${INSTANCE_ID} ${DATABASE_ID} ${PROJECT_ID}`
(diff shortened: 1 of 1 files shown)


samples/system-test/spanner.test.js
```diff
@@ -66,6 +66,17 @@ const GAX_OPTIONS = {
   },
 };
 
+const delay = async test => {
+  const retries = test.currentRetry();
+  if (retries === 0) return; // no retry on the first failure.
```
skuruppu, line 71:
```suggestion
  // No retry on the first failure.
  if (retries === 0) return;
```

samples/system-test/spanner.test.js
```diff
@@ -66,6 +66,17 @@ const GAX_OPTIONS = {
...
+  // see: https://cloud.google.com/storage/docs/exponential-backoff:
```
skuruppu, line 72:
```suggestion
  // See: https://cloud.google.com/storage/docs/exponential-backoff
```

samples/system-test/spanner.test.js
```diff
@@ -66,6 +66,17 @@ const GAX_OPTIONS = {
...
+  const ms = Math.pow(2, retries) * 250 + Math.random() * 1000;
```
skuruppu, line 73: Out of curiosity, why is it not?
```
const ms = (Math.pow(2, retries) + Math.random()) * 1000;
```

I guess I'm wondering about the 250msec increments of time.

samples/system-test/spanner.test.js
```diff
@@ -293,7 +304,10 @@ describe('Spanner', () => {
   });
 
   // create_storing_index
-  it('should create a storing index in an example table', async () => {
+  it('should create a storing index in an example table', async function () {
+    this.retries(5);
+    await delay(this.test); // delay the start of the test, if this is a retry.
```
skuruppu, line 309:
```suggestion
    // Delay the start of the test, if this is a retry.
    await delay(this.test);
```

skuruppu (approved):
Just a few minor nits but LGTM otherwise. Thanks heaps for fixing the test issue.

samples/system-test/spanner.test.js
AVaksman: Done, thanks!

samples/system-test/spanner.test.js
AVaksman: Done

samples/system-test/spanner.test.js
AVaksman: Done

samples/system-test/spanner.test.js
AVaksman: Done

samples/system-test/spanner.test.js
```diff
@@ -293,7 +305,11 @@ describe('Spanner', () => {
   });
 
   // create_storing_index
-  it('should create a storing index in an example table', async () => {
+  it('should create a storing index in an example table', async function () {
+    this.retries(5);
+    // delay the start of the test, if this is a retry.
```
skuruppu, line 310:
```suggestion
    // Delay the start of the test, if this is a retry.
```

PR #1196, opened 2020-07-28
feat: add resource header prefix

### LINKED ISSUE #1186: Set the google-cloud-resource-prefix header in clients
The `google-cloud-resource-prefix` header is expected to be set for the data and two admin clients of Cloud Spanner. I found that the header is set in [Firestore](https://github.com/googleapis/nodejs-firestore/blob/3686816d4a98cabaee4bac089c34a335aa4ddbf8/dev/src/index.ts#L1217) and [Datastore](https://github.com/googleapis/nodejs-datastore/blob/f12091d3536509be677d04b38094b9d5df820d4e/src/request.ts#L1404).

So we can follow that example to set it for the Spanner clients.

AVaksman:

Fixes #1186

| Method | Header|
|--------------------|-------------------------| 
createInstance | project
listInstances | project
listInstancesStream | project
listInstanceConfigs | project
listInstanceConfigsStream | project
deleteInstance | instance
getInstance | instance
updateInstance | instance
createBackup | instance
getBackup | instance
updateBackup | instance
deleteBackup | instance
restoreDatabase | instance
listBackups | instance
listBackupOperations | instance
listDatabaseOperations | instance
createDatabase | instance
listDatabases | instance
dropDatabase | database
getDatabase | database
getDatabaseDdl | database
updateDatabaseDdl | database
partitionQuery | database
partitionRead | database
batchCreateSessions | database
createSession | database
listSessions | database
listSessionsStream | database
deleteSession | database
getSession | database
executeSql | database
beginTransaction | database
streamingRead | database
executeStreamingSql | database
executeBatchDml | database
commit | database
rollback | database


AVaksman:
@skuruppu please verify the correctness of header to method

src/index.ts
```diff
@@ -504,28 +505,25 @@ class Spanner extends GrpcService {
...
-    const gaxOpts = extend(true, {}, options.gaxOptions);
-
-    let reqOpts = extend({}, options, {
-      parent: 'projects/' + this.projectId,
-    });
-
-    delete reqOpts.gaxOptions;
+    const gaxOpts = addResourcePrefixHeader(
+      options.gaxOptions!,
+      'projects/' + this.projectId
```
skuruppu, line 510: Should we just add a this.formattedName_ field to this class?

test/batch-transaction.ts
```diff
@@ -201,6 +210,20 @@ describe('BatchTransaction', () => {
...
+      batchTransaction.createPartitions_(CONFIG, assert.ifError);
+      const {gaxOpts} = REQUEST.lastCall.args[0];
+      assert.deepStrictEqual(
+        gaxOpts,
+        addResourcePrefixHeader(
+          CONFIG.gaxOpts,
+          (batchTransaction.session.parent as Database).formattedName_
+        )
+      );
+      console.log(gaxOpts);
```
skuruppu, line 223: nit: remove

test/batch-transaction.ts
```diff
@@ -201,6 +210,20 @@ describe('BatchTransaction', () => {
       assert.deepStrictEqual(reqOpts.transaction, {id: ID});
     });
 
+    it('should add resource prefix to `gaxOpt`', done => {
```
skuruppu, line 213:
```suggestion
    it('should add resource prefix to `gaxOpts`', done => {
```

skuruppu (approved):
Thanks very much for making this change, especially with having to switch between the various values that could be set for this header.

stephenplusplus:
Would it work to do this all in the `request` function, somewhere around here: https://github.com/googleapis/nodejs-spanner/blob/d835ba21489885f1fc092011d955c9ca5ef162ca/src/index.ts#L856?

To handle the different values that are used for the header, the value could be passed to the function call, e.g.:

```js
this.request({
  client: 'InstanceAdminClient',
  method: 'listInstanceConfigs',
  reqOpts,
  gaxOpts,
  resourcePrefixHeader: 'projects/' + this.projectId,
});
```

Or even:

```js
this.request({
  client: 'InstanceAdminClient',
  method: 'listInstanceConfigs',
  reqOpts,
  gaxOpts,
  headers: {
    'google-cloud-resource-prefix': 'projects/' + this.projectId,
  },
});
```

Reducing it to this format should save a lot of code, and be a bit more future proof should some other requirement come along to set a property on the gaxOptions, like:

```js
gaxOpts: addTrackingIdHeader(addResourcePrefixHeader({}, this.formattedName_)),
```

AVaksman:
Different methods require a different level of granularity, project, instance, database.
The desired data can come in the `reqOpts` object under different properties, `parent`, `name`, `database`, `backup.name`, `instance.name`, `session` (in case of `session`, for header purposes the path would have to be shortened to `database`).

stephenplusplus:
I'm sorry if I'm not catching on quickly enough, but how does that invalidate my suggestion for code simplification? I'm just looking at replacing all of the calls to `addResourcePrefixHeader(theValue)` with something more generic and adaptable in the future.

```
export function addResourcePrefixHeader(
  gaxOpts: CallOptions,
  headerValue: string
): CallOptions {
  return extend(true, {}, gaxOpts, {
    otherArgs: {
      headers: {[CLOUD_RESOURCE_HEADER]: headerValue},
    },
  });
}
```

AVaksman:
@stephenplusplus PTAL
Is it closer to what you had in mind?

src/index.ts
```diff
@@ -365,10 +374,13 @@ class Spanner extends GrpcService {
         ['A configuration object is required to create an instance.'].join('')
       );
     }
-    const formattedName = Instance.formatName_(this.projectId, name);
+    const formattedName = Instance.formatName_(
```
stephenplusplus, line 377: Can we keep the `formatName_` functions written the same as before, and still just provide "this.projectId" as the first argument, instead of changing it to expect the "projects/" prefix?

src/index.ts
```diff
@@ -856,7 +854,7 @@ class Spanner extends GrpcService {
       const requestFn = gaxClient[config.method].bind(
         gaxClient,
         reqOpts,
-        config.gaxOpts
+        addResourcePrefixHeader(config.gaxOpts, config.headers)
```
stephenplusplus, line 857: I think we should just do the extend within this function, rather than have the very specific, single-use external method. It will be more clear for someone to see how it works.

stephenplusplus (commented):
:+1: Thank you! I had some small tweaks to suggest in this review, but I appreciate you taking the time to make the changes.

stephenplusplus:
@AVaksman thank you for all of the changes! I think the resource header is set very nicely. There are a lot of seemingly unrelated changes to how reqOpts combines with gaxOpts. I wouldn't mind a separate PR for those, unless they are actually tied to this change?

AVaksman:
Not, tied to this change.
Will move it in a separate PR.

src/index.ts
```diff
@@ -254,6 +258,10 @@ class Spanner extends GrpcService {
     this.auth = new GoogleAuth(this.options);
     this.clients_ = new Map();
     this.instances_ = new Map();
+    this.projectFormattedName_ = 'projects/' + this.projectId;
```
stephenplusplus, line 261: We should check that this is set on the instance as well.

test/backup.ts
```diff
@@ -164,6 +167,7 @@ describe('Backup', () => {
 
         assert.notStrictEqual(config.reqOpts, QUERY);
         assert.deepStrictEqual(QUERY, ORIGINAL_QUERY);
+        assert.deepStrictEqual(config.headers, resourceHeader);
```
stephenplusplus, line 170: To test that we're sending the right header, we should compare to "instance.resourceHeader_".

stephenplusplus (changes requested):

src/database.ts
```diff
@@ -849,6 +856,7 @@ class Database extends GrpcServiceObject {
           client: 'DatabaseAdminClient',
           method: 'dropDatabase',
           reqOpts,
+          gaxOpts: addResourcePrefixHeader({}, this.formattedName_),
```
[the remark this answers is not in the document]
AVaksman: Resolved with https://github.com/googleapis/nodejs-spanner/pull/1206

AVaksman:
@stephenplusplus PTAL

test/database.ts
```diff
@@ -993,7 +1006,7 @@ describe('Database', () => {
     it('should accept gaxOptions', done => {
       const gaxOptions = {};
       database.request = config => {
-        assert.strictEqual(config.gaxOpts, gaxOptions);
+        assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 1009: This is a test that would be a false positive, unless we put in fake properties or revert to comparing that the input object is strictly equal.

test/database.ts
```diff
@@ -1648,7 +1665,7 @@ describe('Database', () => {
...
```
stephenplusplus, line 1668: This is a test that would be a false positive, unless we put in fake properties or revert to comparing that the input object is strictly equal.

stephenplusplus (changes requested):
Thank you! Just these few gaxOptions lingering issues.

AVaksman:
Done, thank you!

test/batch-transaction.ts
```diff
@@ -181,7 +190,7 @@ describe('BatchTransaction', () => {
     const RESPONSE = {partitions: PARTITIONS};
 
     const QUERY = {a: 'b'};
-    const CONFIG = {reqOpts: QUERY};
+    const CONFIG = {reqOpts: QUERY, gaxOpts: {timeout: 1000}};
```
stephenplusplus, line 193:
```suggestion
    const CONFIG = {reqOpts: QUERY};
```

It doesn't look like this function is doing anything with `gaxOpts`, or that we're testing it anywhere, so having this defined here could be confusing.

test/database.ts
```diff
@@ -1017,7 +1032,7 @@ describe('Database', () => {
...
```
stephenplusplus, line 1035:
```suggestion
        assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/instance.ts
```diff
@@ -258,7 +269,7 @@ describe('Instance', () => {
     it('should accept gaxOptions', done => {
       const options = Object.assign({}, OPTIONS, {gaxOptions: {}});
       instance.request = config => {
-        assert.strictEqual(config.gaxOpts, options.gaxOptions);
+        assert.deepStrictEqual(config.gaxOpts, options.gaxOptions);
```
stephenplusplus, line 272:
```suggestion
        assert.strictEqual(config.gaxOpts, options.gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `options.gaxOptions`.

test/instance.ts
```diff
@@ -555,7 +568,7 @@ describe('Instance', () => {
       const gaxOptions = {};
 
       instance.request = (config, callback: Function) => {
-        assert.strictEqual(config.gaxOpts, gaxOptions);
+        assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 571:
```suggestion
        assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/instance.ts
```diff
@@ -1228,7 +1247,7 @@ describe('Instance', () => {
     it('should accept gaxOptions', done => {
       const gaxOptions = {};
       instance.request = config => {
-        assert.strictEqual(config.gaxOpts, gaxOptions);
+        assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 1250:
```suggestion
        assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/instance.ts
```diff
@@ -1273,7 +1294,7 @@ describe('Instance', () => {
...
```
stephenplusplus, line 1297:
```suggestion
        assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/session.ts
```diff
@@ -281,7 +297,7 @@ describe('Session', () => {
     it('should accept and pass gaxOptions to getMetadata', done => {
       const gaxOptions = {};
       session.request = config => {
-        assert.strictEqual(config.gaxOpts, gaxOptions);
+        assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 300:
```suggestion
        assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/session.ts
```diff
@@ -312,7 +330,7 @@ describe('Session', () => {
     it('should accept gaxOptions', done => {
       const gaxOptions = {};
       session.request = config => {
-        assert.strictEqual(config.gaxOpts, gaxOptions);
+        assert.deepStrictEqual(config.gaxOpts, {});
```
stephenplusplus, line 333:
```suggestion
        assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/transaction.ts
```diff
@@ -278,7 +294,7 @@ describe('Transaction', () => {
 
         const {gaxOpts, reqOpts} = REQUEST_STREAM.lastCall.args[0];
 
-        assert.strictEqual(gaxOpts, fakeOptions);
+        assert.deepStrictEqual(gaxOpts, fakeOptions);
```
stephenplusplus, line 297:
```suggestion
        assert.strictEqual(config.gaxOpts, fakeOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `fakeOptions`.

test/transaction.ts
```diff
@@ -558,7 +575,7 @@ describe('Transaction', () => {
         const {gaxOpts, reqOpts} = REQUEST_STREAM.lastCall.args[0];
 
         assert.strictEqual(reqOpts.gaxOptions, undefined);
-        assert.strictEqual(gaxOpts, fakeQuery.gaxOptions);
+        assert.deepStrictEqual(gaxOpts, fakeQuery.gaxOptions);
```
stephenplusplus, line 578:
```suggestion
        assert.strictEqual(gaxOpts, fakeQuery.gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `fakeQuery.gaxOptions`.

test/transaction.ts
```diff
@@ -1022,7 +1039,7 @@ describe('Transaction', () => {
       it('should accept gaxOptions', done => {
         const gaxOptions = {};
         transaction.request = config => {
-          assert.strictEqual(config.gaxOpts, gaxOptions);
+          assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 1042:
```suggestion
          assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/transaction.ts
```diff
@@ -1190,15 +1208,18 @@ describe('Transaction', () => {
...
+        assert.strictEqual(method, 'beginTransaction');
         assert.deepStrictEqual(reqOpts.options, expectedOptions);
+        assert.deepStrictEqual(headers, transaction.resourceHeader_);
       });
 
       it('should accept gaxOptions', done => {
         const gaxOptions = {};
         transaction.request = config => {
-          assert.strictEqual(config.gaxOpts, gaxOptions);
+          assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 1222:
```suggestion
          assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/transaction.ts
```diff
@@ -1211,18 +1232,19 @@ describe('Transaction', () => {
...
         assert.strictEqual(reqOpts.session, SESSION_NAME);
         assert.deepStrictEqual(reqOpts.mutations, []);
+        assert.deepStrictEqual(headers, transaction.resourceHeader_);
       });
 
       it('should accept gaxOptions', done => {
         const gaxOptions = {};
         transaction.request = config => {
-          assert.strictEqual(config.gaxOpts, gaxOptions);
+          assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 1247:
```suggestion
          assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

test/transaction.ts
```diff
@@ -1411,17 +1433,18 @@ describe('Transaction', () => {
...
         assert.strictEqual(method, 'rollback');
         assert.deepStrictEqual(reqOpts, expectedReqOpts);
+        assert.deepStrictEqual(headers, transaction.resourceHeader_);
       });
 
       it('should accept gaxOptions', done => {
         const gaxOptions = {};
         transaction.request = config => {
-          assert.strictEqual(config.gaxOpts, gaxOptions);
+          assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
stephenplusplus, line 1447:
```suggestion
          assert.strictEqual(config.gaxOpts, gaxOptions);
```

If after making that change, the test fails, then we would want to keep `deepStrictEqual`, but also set a property on `gaxOptions`.

stephenplusplus (changes requested):

src/session.ts
```diff
@@ -36,7 +36,9 @@ import {
   CreateSessionOptions,
 } from './database';
 import {ServiceObjectConfig} from '@google-cloud/common';
-import {NormalCallback} from './common';
+import {NormalCallback, CLOUD_RESOURCE_HEADER} from './common';
+// import {ServiceObjectConfig} from '@google-cloud/common';
```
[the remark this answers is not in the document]
stephenplusplus: I think these comments can be removed. (This and the next line)

stephenplusplus (changes requested):

stephenplusplus (approved):

PR #1198, opened 2020-07-29
feat: set metadata during getBackups and getMetadata calls

AVaksman:


test/session.ts
```diff
@@ -264,28 +264,49 @@ describe('Session', () => {
...
       };
       session.getMetadata(gaxOptions, assert.ifError);
     });
+
+    it('should update metadata', done => {
+      const metadata = {};
+      session.request = (config, callback) => {
+        callback(null, metadata);
+      };
+      session.getMetadata(() => {
```
stephenplusplus, line 294:
```suggestion
      session.getMetadata((err) => {
        assert.ifError(err);
```

src/instance.ts
```diff
@@ -355,7 +355,9 @@ class Instance extends common.GrpcServiceObject {
         let backupInstances: Backup[] | null = null;
         if (backups) {
           backupInstances = backups.map(backup => {
-            return this.backup(backup.name!);
+            const backupInstance = this.backup(backup.name!);
+            backupInstance.metadata = backup;
```
stephenplusplus, line 359: Nice catch. Please add a test to see that this is set.

stephenplusplus (changes requested):

src/instance.ts
AVaksman: I believe I added that verification in an existing [test](https://github.com/googleapis/nodejs-spanner/pull/1198/files#diff-189d5720e2e29b973f957a72855ced1cR1459). Is it sufficient?

test/session.ts
AVaksman: Done, thanks!

src/instance.ts
stephenplusplus: 🤦 sorry. LG!

stephenplusplus (approved):

PR #1199, opened 2020-07-29
feat: return full nextQuery object in paginated calls

AVaksman:

Combine a `nextPageRequest` from server with user provided custom options, this way user can just pass it to the next call as it is documented in the examples

https://github.com/googleapis/nodejs-spanner/blob/d835ba21489885f1fc092011d955c9ca5ef162ca/src/instance.ts#L1159-L1171

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -1336,7 +1336,7 @@ class Database extends GrpcServiceObject {
         reqOpts,
         gaxOpts,
       },
-      (err, sessions, nexPageRequest, ...args) => {
+      (err, sessions, nextPageRequest, ...args) => {
         let sessionInstances: Session[] | null = null;
         if (sessions) {
           sessionInstances = sessions.map(metadata => {
@@ -1345,8 +1345,8 @@ class Database extends GrpcServiceObject {
             return session;
           });
         }
-        const nextQuery = nexPageRequest!
-          ? extend({}, options, nexPageRequest!)
+        const nextQuery = nextPageRequest!
+          ? extend({}, options, nextPageRequest!)
           : null;
         callback!(err, sessionInstances!, nextQuery, ...args);
       }
--- a/src/database.ts
+++ b/src/database.ts
@@ -1336,7 +1336,7 @@ class Database extends GrpcServiceObject {
         reqOpts,
         gaxOpts,
       },
-      (err, sessions, ...args) => {
+      (err, sessions, nexPageRequest, ...args) => {
         let sessionInstances: Session[] | null = null;
         if (sessions) {
           sessionInstances = sessions.map(metadata => {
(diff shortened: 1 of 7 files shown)


src/database.ts
```diff
@@ -1336,7 +1336,7 @@ class Database extends GrpcServiceObject {
         reqOpts,
         gaxOpts,
       },
-      (err, sessions, ...args) => {
+      (err, sessions, nexPageRequest, ...args) => {
```
stephenplusplus, line 1339: `nextPageRequest`

stephenplusplus, line 1339: What is the format/properties of `nextPageRequest`?

src/index.ts
```diff
@@ -534,7 +534,7 @@ class Spanner extends GrpcService {
         reqOpts,
         gaxOpts,
       },
-      (err, instances, ...args) => {
+      (err, instances, nexPageRequest, ...args) => {
```
stephenplusplus, line 537: `nextPageRequest`*

src/index.ts
```diff
@@ -737,7 +740,12 @@ class Spanner extends GrpcService {
         reqOpts,
         gaxOpts,
       },
-      callback
+      (err, instanceConfigs, nexPageRequest, ...args) => {
```
stephenplusplus, line 743: `nextPageRequest`*

src/instance.ts
```diff
@@ -351,15 +351,18 @@ class Instance extends common.GrpcServiceObject {
         reqOpts,
         gaxOpts,
       },
-      (err, backups, ...args) => {
+      (err, backups, nexPageRequest, ...args) => {
```
stephenplusplus, line 354: `nextPageRequest`*

src/instance.ts
```diff
@@ -545,7 +548,13 @@ class Instance extends common.GrpcServiceObject {
         reqOpts,
         gaxOpts,
       },
-      callback
+      (err, operations, nexPageRequest, ...args) => {
```
stephenplusplus, line 551: `nextPageRequest`*

src/instance.ts
```diff
@@ -663,7 +672,13 @@ class Instance extends common.GrpcServiceObject {
...
```
stephenplusplus, line 675: `nextPageRequest`*

src/instance.ts
```diff
@@ -1221,7 +1236,7 @@ class Instance extends common.GrpcServiceObject {
         reqOpts,
         gaxOpts,
       },
-      (err, rowDatabases, ...args) => {
+      (err, rowDatabases, nexPageRequest, ...args) => {
```
stephenplusplus, line 1239: `nextPageRequest`*

stephenplusplus (changes requested):

src/database.ts
AVaksman: Thanks,
Fixed here and everywhere below.

AVaksman: `nextPageRequest` comes in as `proto.~.IList<resource>Request` which is usually
```ts
parent/database;
pageSize;
pageToken;
//some request objects also have 
filter
```

stephenplusplus (approved):

stephenplusplus:
@AVaksman should we merge, or do we need to wait for another review?

stephenplusplus:
It seemed mergeable. Hopefully that wasn't a mistake!

Issue #1177, opened 2020-07-09
Improvements for retrying PDML queries


skuruppu:
I looked through the [`runPartitionedUpdate` ](https://github.com/googleapis/nodejs-spanner/blob/1953725aa3d2d971b0cd868d8376a8280004509a/src/database.ts#L2014) implementation and we correctly use the `ExecuteStreamingSql` RPC to run the update. I think based on our current understanding of how the backend deals with retries, it will also be good to add the following logic on the client library:

- If the transaction fails with a transient error like UNAVAILABLE or DEADLINE_EXCEEDED, and if a `PartialResultSet` was received with a resume token, we should retry the transaction with the same transaction ID.
- If the transaction returns an ABORTED error or if we encounter a retriable error and didn't receive a resume token, then we should retry with a new transaction ID.

skuruppu:
Looks like @olavloite has a fix in #1205 and @thiagotnunes will try it with the test data next week.

PR #1205, opened 2020-08-04
fix: retry PDML on Aborted and Internal errors

### LINKED ISSUE #1197: spanner-client: Retry PDML on "Received unexpected EOS on DATA frame from server"
This bug is related to the Spanner client library.

For long lived transactions (>= 30 minutes), in the case of large PDML changes, it is possible that the gRPC connection is terminated with an error "Received unexpected EOS on DATA frame from server".

In this case, we need to retry the transaction either with the received resume token obtained on reading the stream or from scratch. This will ensure that the PDML transaction continues to execute until it is successful or a hard timeout is reached.

We have already implemented such change in the Java client library, for more information see this PR: https://github.com/googleapis/java-spanner/pull/360.

In order to test the fix, we can use a large spanner database. Please speak to @thiagotnunes for more details.

olavloite:
PDML statements should be retried on the following errors:
* ABORTED: Retry using a new transaction
* INTERNAL(Received unexpected EOS on DATA frame from server):
    Retry using the same transaction and any resume token that was received.

Fixes #1197, #1177

### DIFF
--- a/test/database.ts
+++ b/test/database.ts
@@ -2235,7 +2235,6 @@ describe('Database', () => {
         'begin'
       ) as sinon.SinonStub).callsFake(callback => callback(null));
 
-      // runUpdateStub = sandbox.stub(fakePartitionedDml, 'runUpdate');
       runUpdateStub = (sandbox.stub(
         fakePartitionedDml,
         'runUpdate'
--- a/test/database.ts
+++ b/test/database.ts
@@ -2235,7 +2235,11 @@ describe('Database', () => {
         'begin'
       ) as sinon.SinonStub).callsFake(callback => callback(null));
 
-      runUpdateStub = sandbox.stub(fakePartitionedDml, 'runUpdate');
+      // runUpdateStub = sandbox.stub(fakePartitionedDml, 'runUpdate');
+      runUpdateStub = (sandbox.stub(
+        fakePartitionedDml,
+        'runUpdate'
+      ) as sinon.SinonStub).callsFake((_, callback) => callback(null));
     });
 
     it('should get a read only session from the pool', () => {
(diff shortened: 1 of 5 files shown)


thiagotnunes (commented):
LGTM, but we will need to internally test it to make sure. I will try and do that next week. Thanks!

test/database.ts
```diff
@@ -2235,7 +2235,11 @@ describe('Database', () => {
         'begin'
       ) as sinon.SinonStub).callsFake(callback => callback(null));
 
-      runUpdateStub = sandbox.stub(fakePartitionedDml, 'runUpdate');
+      // runUpdateStub = sandbox.stub(fakePartitionedDml, 'runUpdate');
```
skuruppu, line 2238: nit: remove commented out line.

skuruppu (approved):

skuruppu:
Adding a "do not merge" until @thiagotnunes tests the implementation.

thiagotnunes:
@olavloite Still getting an error from the backend (a different one though): `ERROR: Error: 13 INTERNAL: Received RST_STREAM with code 2 (Internal server error)`

If we change the `isRetryable` function to something like, it should be enough:
```js
function isRetryableInternalError(err: grpc.ServiceError): boolean {
  return (
    err.code === grpc.status.INTERNAL &&
    (err.message.includes('Received unexpected EOS on DATA frame from server') || err.message.includes('Received RST_STREAM'))
  );
}
```

olavloite:
@thiagotnunes Thanks a lot for testing it. I've updated the check to include this error as well.

I also had a quick look at a couple of the other client libraries:
1. In Go we also retry this error: https://github.com/googleapis/google-cloud-go/blob/14d43f9dd504466f50d8764d7dd346dce29ed98e/spanner/retry.go#L63
1. In Java we do not: https://github.com/googleapis/java-spanner/blob/6f47b8a759643f772230df0c2e153338d44f70ce/google-cloud-spanner/src/main/java/com/google/cloud/spanner/IsRetryableInternalError.java#L35

Could it be that this is not necessary in Java, because we set a keep-alive on the gRPC transport channel in Java: https://github.com/googleapis/java-spanner/blob/6f47b8a759643f772230df0c2e153338d44f70ce/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L316

thiagotnunes:
@olavloite thanks for the fix!

I don't know why this is not happening in Java, but you might have a good lead there. By searching internally, it seems that for other clients have decided to also add this error as retryable in any case. I think we should add that in the java client to be on the safe side.

thiagotnunes (approved):

PR #1207, opened 2020-08-05
fix: batch transaction should use a session from the pool

### LINKED ISSUE #1200: Update batch transactions to take a session from the pool
Currently [Database.createBatchTransaction](https://github.com/googleapis/nodejs-spanner/blob/41b935ae8742d319640bbff8584ea02885460572/src/database.ts#L547) and [Database.batchTransaction](https://github.com/googleapis/nodejs-spanner/blob/41b935ae8742d319640bbff8584ea02885460572/src/database.ts#L468) both create sessions instead of taking one from the pool.

In https://github.com/googleapis/google-cloud-go/issues/1951 we decided that we should just be using sessions from the pool. So we should change the implementation of these two methods accordingly.

olavloite:
Batch transactions created a separate session that was deleted when the batch transaction was closed. Instead, batch transactions should also use sessions from the pool, and return the session to the pool when the transaction is done.

Fixes #1200


skuruppu (approved):

skuruppu:
Adding a "do not merge" while we confirm from the Spanner team on if there are reasons why we shouldn't do this.

skuruppu:
I haven't heard anything from the backend team so I'm assuming there wasn't any concerns. @olavloite, please merge after resolving the conflicts.

PR #1208, opened 2020-08-05
fix: handle potential errors when creating stream

### LINKED ISSUE #1078: Errors from encodeValue_ are not caught and generate a UnhandledPromiseRejectionWarning

olavloite:
Errors in transaction.runStream(..) were not handled by database.runStream(..), which could cause Unhandled Promise Rejections.

Fixes #1078


skuruppu (approved):

skuruppu:
It would also be great to get feedback from @AVaksman since he was looking at this issue a few weeks ago.

stephenplusplus:
I'm not sure exactly where the thrown error is originating from, but would it be possible to move the try/catch to what I believe is the parent method?: https://github.com/googleapis/nodejs-spanner/blob/a257e8c5924d5040e1c1dfe6570b3d3b001eb05b/src/transaction.ts

olavloite:
> I'm not sure exactly where the thrown error is originating from, but would it be possible to move the try/catch to what I believe is the parent method?: https://github.com/googleapis/nodejs-spanner/blob/a257e8c5924d5040e1c1dfe6570b3d3b001eb05b/src/transaction.ts

I'm afraid the reference does not really point to a method, but only to a file. But assuming you mean the [runStream](https://github.com/googleapis/nodejs-spanner/blob/6e1f300ebe31027eb1f2922facb70bb0ba392b11/src/transaction.ts#L889) method in transaction.ts, then that is not a parent method of `database.runStream`. Conceptually what is going on here is the following:
1. The user application wants to execute a query or update on a database (`database.runStream(..)`)
1. The database [checks out a session from the session pool](https://github.com/googleapis/nodejs-spanner/blob/a257e8c5924d5040e1c1dfe6570b3d3b001eb05b/src/database.ts#L2167).
1. The database [starts a transaction on the session](https://github.com/googleapis/nodejs-spanner/blob/a257e8c5924d5040e1c1dfe6570b3d3b001eb05b/src/database.ts#L2173). The type of transaction is given in the options.
1. The database [executes the query / update on the transaction](https://github.com/googleapis/nodejs-spanner/blob/a257e8c5924d5040e1c1dfe6570b3d3b001eb05b/src/database.ts#L2180) that was created on the session. This is the `runStream(..)` method in transaction.ts.

The error occurs while encoding the query parameters that should be sent with the query to the server. If this fails, the query will never be sent to the server and there will never be any stream returned from the method. We need to handle that here, as we need to [return the session to the session pool](https://github.com/googleapis/nodejs-spanner/blob/a257e8c5924d5040e1c1dfe6570b3d3b001eb05b/src/database.ts#L2183).

stephenplusplus:
Woops, sorry for the bad link and thanks for the explanation.

The method you assumed I meant is correct. That is defined on the Snapshot class, which looks to be the method being called in this PR that is now wrapped in a try/catch (`snapshot.runStream()`)? https://github.com/googleapis/nodejs-spanner/blob/a257e8c5924d5040e1c1dfe6570b3d3b001eb05b/src/database.ts#L2180

olavloite:
@stephenplusplus What we could do is move the error handling to `snapshot.runStream` and, in case of an error, let that method return a `PartialResultStream` that will emit the error. That way, the normal error handling of the data stream can be used. WDYT?

stephenplusplus:
I think in this case, we should make a slight adjustment to suit users who do this:

```js
const queryStream = database.runStream(...);
// later...
queryStream.on('data', () => {});
```

Users would receive an immediate, uncaught error unless they always assign an error handler right after the "runStream()" call.

Our code in the PRS file will actually wait for the stream to be reading before it attempts the network request. I think we should align our sanitization step with that cadence, something like:

```js
const {gaxOptions, json, jsonOptions, maxResumeRetries} = query;
let reqOpts;

const sanitizeRequest = () => {
  const {params, paramTypes} = Snapshot.encodeParams(query);
  const transaction: spannerClient.spanner.v1.ITransactionSelector = {};
  if (this.id) {
    transaction.id = this.id as Uint8Array;
  } else {
    transaction.singleUse = this._options;
  }
  delete query.gaxOptions;
  delete query.json;
  delete query.jsonOptions;
  delete query.maxResumeRetries;
  delete query.types;
  reqOpts = Object.assign(query, {
    session: this.session.formattedName_!,
    seqno: this._seqno++,
    transaction,
    params,
    paramTypes,
  });
}

const makeRequest = (resumeToken?: ResumeToken): Readable => {
  if (!reqOpts) {
    try {
      sanitizeRequest();
    } catch (e) {
      const errorStream = new stream.PassThrough();
      setImmediate(() => errorStream.destroy(e));
      return errorStream;
    }
  }

  return this.requestStream({
    client: 'SpannerClient',
    method: 'executeStreamingSql',
    reqOpts: Object.assign({}, reqOpts, {resumeToken}),
    gaxOpts: gaxOptions,
  });
};

return partialResultStream(makeRequest, {
  json,
  jsonOptions,
  maxResumeRetries,
});
```

olavloite:
@stephenplusplus I updated the PR based on your suggestion. PTAL.

stephenplusplus:
Thank you! The only thing left is to fit the tests into the appropriate `describe` in "test/transaction.ts", like:

- "should sanitize input"
- "should destroy stream with validation error"

olavloite:
> Thank you! The only thing left is to fit the tests into the appropriate `describe` in "test/transaction.ts", like:
> 
> * "should sanitize input"
> * "should destroy stream with validation error"

Done

src/transaction.ts
```diff
@@ -898,30 +899,42 @@ export class Snapshot extends EventEmitter {
...
-      session: this.session.formattedName_!,
-      seqno: this._seqno++,
-      transaction,
-      params,
-      paramTypes,
-    });
+    let reqOpts;
+
+    const sanitizeRequest = () => {
+      const q = query as ExecuteSqlRequest;
```
AVaksman, line 905: nit: Is this step needed? It seems `query` is anyways used below line [918](https://github.com/googleapis/nodejs-spanner/pull/1208/files#diff-f5daf90433414776e2e49f79b8c93adaR918)

test/transaction.ts
```diff
@@ -629,6 +629,40 @@ describe('Transaction', () => {
...
+      });
+
+      it('should return an error stream for invalid parameters', done => {
+        REQUEST_STREAM.resetHistory();
+
+        const fakeQuery = Object.assign({}, QUERY, {
+          params: {a: undefined},
+        });
+
+        snapshot.runStream(fakeQuery).on('error', () => done());
```
AVaksman, line 662: nit: Should the error message be verified as well?
```ts
        const stream = snapshot.runStream(fakeQuery);
        stream.on('error', error => {
          assert.strictEqual(
            error.message,
            'Value of type undefined not recognized.'
          );
          done();
        });
```
Or that is excessive?

AVaksman (approved):

PR #1210, opened 2020-08-07
docs: sample on setting per-request gaxOptions

### LINKED ISSUE #1203: Add code sample on setting per-request gaxOptions
The request here is to add a code sample that shows how to set per-request retry and timeout settings. The example we can use is the equivalent one done for the Go client library.

https://github.com/GoogleCloudPlatform/golang-samples/pull/1594

laljikanjareeya:
Fixes googleapis/google-cloud-node-core#339


samples/dml.js
```diff
@@ -630,6 +630,74 @@ async function updateUsingBatchDml(instanceId, databaseId, projectId) {
...
+      backoffSettings: {
+        // Configure retry delay settings.
+        initialRetryDelayMillis: 500,
+        maxRetryDelayMillis: 64000,
+        retryDelayMultiplier: 1.5,
+        // Configure RPC and total timeout settings.
+        initialRpcTimeoutMillis: 60000,
+        rpcTimeoutMultiplier: 1.0,
+        maxRpcTimeoutMillis: 60000,
+        totalTimeoutMillis: 600000,
```
hengfengli, line 667: This should be `60s` as well.

hengfengli (approved):
LGTM, 👍 . Only need a small change.

samples/dml.js
```diff
@@ -630,6 +630,74 @@ async function updateUsingBatchDml(instanceId, databaseId, projectId) {
...
+   */
+  // const projectId = 'my-project-id';
+  // const instanceId = 'my-instance';
+  // const databaseId = 'my-database';
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+  const DEADLINE_EXCEEDED_STATUS_CODE = 4;
```
hengfengli, line 653: Why we don't use `grpc.status.DEADLINE_EXCEEDED`? Instead, we use hardcoded values here.

samples/dml.js
```diff
@@ -630,6 +630,74 @@ async function updateUsingBatchDml(instanceId, databaseId, projectId) {
   // [END spanner_dml_batch_update]
 }
 
+async function executeSqlWithCustomTimeoutAndRetrySettings(
```
hengfengli, line 633: Does the Samples Test fail due to the `async` prefix here? I don't see what the difference is, comparing to https://github.com/googleapis/nodejs-spanner/blob/0a50621ec24082ba0e4ccf932d5cb7402941092c/samples/dml.js#L20-L69.

laljikanjareeya: It's not about `async`, currently I am investigating on it.

hengfengli:
Hi @laljikanjareeya, is there any news for this?

samples/dml.js
skuruppu: This issue turns up purely when you set the gaxOptions. If you don't, the sample runs fine.

Also, you can't set a subset of the fields. You have to set all of them.
 - first I set only the retry codes then it said `ERROR: TypeError: Cannot read property 'retryDelayMultiplier' of undefined`.
- then I set `retryDelayMultiplier` then it said `Error: 2 UNKNOWN: Getting metadata from plugin failed with error: Deadline is too far in the future`
- until you set each field, it keeps producing the same error.
- once you set all of them, it errors with `ERROR: TypeError: stream.on is not a function`.

So it has something to do with the fact that this is a streaming RPC. We set gaxOptions in the admin RPC calls in the system tests and this doesn't happen.

samples/dml.js
skuruppu: You can do:
```
const {grpc} = require('google-gax');
...
retryCodes: [
  grpc.status.DEADLINE_EXCEEDED,
  grpc.status.UNAVAILABLE,
],
```

samples/dml.js
laljikanjareeya: @skuruppu @hengfengli  something with streaming in `google-gax` , I did check this in `bigtable` and found same error with stream.

laljikanjareeya: I see, we do not support retry in stream, [Here](https://github.com/googleapis/gax-nodejs/issues/729) is the issue in `google-gax`.

samples/dml.js
laljikanjareeya: It gives below error as we don't have direct dependency of `google-gax` in sample
`##[error]  17:24  error  "google-gax" is extraneous  node/no-extraneous-require`

skuruppu: Oh yeah I remember that happening for a separate issue. I don't know of a workaround. It's ok to leave this as is for now. Thanks for trying.

skuruppu (approved):

skuruppu:
Please also wait for approval by @hengfengli since the sample now looks different to the one in Go.

hengfengli (approved):
LGTM. 👍 Thanks for doing this.

PR #1220, opened 2020-08-14
fix: `Database not found` could be returned after create()

### LINKED ISSUE #1219: Spanner with mock server session-pool: should not propagate instance and database not found errors for SessionPoolOptions.min > 0 failed
This test failed!

To configure my behavior, see [the Build Cop Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop).

If I'm commenting on this issue too often, add the `buildcop: quiet` label and
I will stop commenting.

---

commit: 0708baab276def6cad0f0f3c4b7589084380c2d8
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/1505f14d-06a0-468c-b15e-4f438d5d652e), [Sponge](http://sponge2/1505f14d-06a0-468c-b15e-4f438d5d652e)
status: failed
<details><summary>Test output</summary><br><pre>5 NOT_FOUND: Database not found
Error: 5 NOT_FOUND: Database not found
    at Object.callErrorFromStatus (node_modules/@grpc/grpc-js/build/src/call.js:31:26)
    at Object.onReceiveStatus (node_modules/@grpc/grpc-js/build/src/client.js:176:52)
    at listener.onReceiveStatus (node_modules/@grpc/grpc-js/build/src/call-stream.js:65:35)
    at Object.onReceiveStatus (node_modules/grpc-gcp/build/src/index.js:73:29)
    at InterceptingListenerImpl.onReceiveStatus (node_modules/@grpc/grpc-js/build/src/call-stream.js:60:23)
    at Object.onReceiveStatus (node_modules/@grpc/grpc-js/build/src/client-interceptors.js:342:141)
    at Object.onReceiveStatus (node_modules/@grpc/grpc-js/build/src/client-interceptors.js:305:181)
    at Http2CallStream.outputStatus (node_modules/@grpc/grpc-js/build/src/call-stream.js:117:74)
    at Http2CallStream.maybeOutputStatus (node_modules/@grpc/grpc-js/build/src/call-stream.js:156:22)
    at Http2CallStream.endCall (node_modules/@grpc/grpc-js/build/src/call-stream.js:142:18)
    at Http2CallStream.handleTrailers (node_modules/@grpc/grpc-js/build/src/call-stream.js:274:14)
    at emit (internal/http2/core.js:265:8)
    at process._tickCallback (internal/process/next_tick.js:63:19)</pre></details>

olavloite:
The session pool could return a Database not found after calling database.create(). This would happen in the following case:
1. A database that does not yet exist is opened with a session pool config with Min > 0.
2. The fill() method of the session pool will then initiate one or more BatchCreateSessions RPCs.
3. The client application calls database.create().
4. The client application executes a query on the newly created database. While doing so, the client
   will request a session from the pool. If one of the requests from step 2 had still not finished at
   this time, its result (Database not found) could be propagated to this request.

Fixes #1219

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -316,21 +316,30 @@ class Database extends common.GrpcServiceObject {
           // wait until these have finished before we try to create the database.
           // Otherwise the results of these requests might be propagated to
           // client requests that are submitted after the database has been
-          // created.
-          const p = new Promise<void>((resolve, reject) => {
-            pool
-              .on('available', () => {
-                if (pool._pending === 0) {
-                  resolve();
-                }
-              })
-              .on('createError', () => {
-                if (pool._pending === 0) {
-                  resolve();
-                }
-              });
-          });
-          p.then(() =>
+          // created. If the pending requests have not finished within 10 seconds,
+          // they will be ignored and the database creation will proceed.
+          let timeout;
+          const promises = [
+            new Promise<void>(
+              resolve => (timeout = setTimeout(resolve, 10000))
+            ),
+            new Promise<void>(resolve => {
+              pool
+                .on('available', () => {
+                  if (pool._pending === 0) {
+                    clearTimeout(timeout);
+                    resolve();
+                  }
+                })
+                .on('createError', () => {
+                  if (pool._pending === 0) {
+                    clearTimeout(timeout);
+                    resolve();
+                  }
+                });
+            }),
+          ];
+          Promise.race(promises).then(() =>
             instance.createDatabase(formattedName_, options, callback)
           );
         } else {
--- a/src/database.ts
+++ b/src/database.ts
@@ -310,7 +310,32 @@ class Database extends common.GrpcServiceObject {
         options: CreateDatabaseOptions,
         callback: CreateDatabaseCallback
       ) => {
-        return instance.createDatabase(formattedName_, options, callback);
+        const pool = this.pool_ as SessionPool;
+        if (pool._pending > 0) {
+          // If there are BatchCreateSessions requests pending, then we should
+          // wait until these have finished before we try to create the database.
+          // Otherwise the results of these requests might be propagated to
+          // client requests that are submitted after the database has been
+          // created.
+          const p = new Promise<void>((resolve, reject) => {
+            pool
+              .on('available', () => {
+                if (pool._pending === 0) {
+                  resolve();
+                }
+              })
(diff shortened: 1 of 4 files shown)


src/database.ts
```diff
@@ -310,7 +310,32 @@ class Database extends common.GrpcServiceObject {
         options: CreateDatabaseOptions,
         callback: CreateDatabaseCallback
       ) => {
-        return instance.createDatabase(formattedName_, options, callback);
+        const pool = this.pool_ as SessionPool;
+        if (pool._pending > 0) {
+          // If there are BatchCreateSessions requests pending, then we should
```
stephenplusplus, line 315: I don't know if this would make sense here, but would a user-provided timeout setting be useful if this is something that could exceed expectations?

olavloite: A user-provided timeout setting is not really useful in this case, as we would require the users to have a very thorough understanding of the internals of the client library for them to be able to set a sensible value. But a system default timeout value does make sense for this. A timeout should however not cause an error, but cause the pending requests to be ignored and then continue to create the database.

src/database.ts
```diff
@@ -310,7 +310,41 @@ class Database extends common.GrpcServiceObject {
...
+                .on('createError', () => {
+                  if (pool._pending === 0) {
+                    clearTimeout(timeout);
+                    resolve();
+                  }
+                });
+            }),
+          ];
+          Promise.race(promises).then(() =>
+            instance.createDatabase(formattedName_, options, callback)
```
skuruppu, line 343: At this point, I guess the fill() call has failed because we haven't created the database yet. Then we create the database. When will the fill() call be retried?

I just want to make sure that we don't get into a situation where the database is created but the session pool is never initialized.

olavloite: Good point, that never happened. The pool would be gradually filled as the user application is requesting sessions, and if the user application was using the default session pool settings this would also happen in one `BatchCreateSessions` RPC (because of the default `incStep: 25, min: 25`. But with custom settings such as for example:

```typescript
min: 25,
max: 400,
incStep: 10,
```

The minimum configured value would not be created directly.

I've added that logic to the `_getSession` function where the dynamic session creation happens, so it also works for any other scenario where the initial filling of the session pool failed.

skuruppu (approved):
Thanks for noticing this issue and fixing it.

PR #1221, opened 2020-08-18
fix: adds chai dev dependency

thiagotnunes:
Adds chai in dev dependencies. This missing dependency is necessary to run the sample tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "@types/uuid": "^8.0.0",
     "binary-search-bounds": "^2.0.4",
     "c8": "^7.0.0",
+    "chai": "^4.2.0",
     "codecov": "^3.0.2",
     "concat-stream": "^2.0.0",
     "dedent": "^0.7.0",


skuruppu (approved):

package.json
```diff
@@ -90,6 +90,7 @@
     "@types/uuid": "^8.0.0",
     "binary-search-bounds": "^2.0.4",
     "c8": "^7.0.0",
+    "chai": "^4.2.0",
```
JustinBeckwith, line 93: If this is only needed for the sample tests, it shouldn't be here, and instead should be in `devDependencies` of `package.json` in `samples/package.json`

JustinBeckwith (changes requested):

package.json
thiagotnunes: @JustinBeckwith good point! I see that `chai` is already in the dev dependencies of the samples `package.json`. I am, however, getting an error on running the sample libraries as per the `CONTRIBUTING.md` file:
```
npm run samples-test
```
Maybe we are missing a step on the `.md` instead?

JustinBeckwith: Yeah, that's the gap here 🤦   You can run the samples tests in 1 of 2 ways:
```
npm install
cd samples
npm install
cd ..
npm run samples-test
```

Or

```
npm install
cd samples
npm install
npm test
```

The `CONTRIBUTING.md` is centrally managed, so we'll need to cut a PR out to all 94 repos for this one.  Apologies for the confusion!

thiagotnunes: No worries, glad we identified the problem! I will close this PR. Let me know if you need any help with the PR in the centralized repo.

Cheers,

PR #1228, opened 2020-08-28
fix: do not create sessions after getDatabases call

### LINKED ISSUE #1227: ListDatabases unintentionally initializes a session pool for each returned database
@larkee pointed out an interesting observation in the Python library if we initialize the session pool with > 0 sessions in the context of `ListDatabses` and I think this might apply in Node.js as well.

1. When we call [`getDatabases`](https://github.com/googleapis/nodejs-spanner/blob/7046d0b22fb1bdf8996887422ec9cf833d38ac37/src/instance.ts#L1135), the [`GetDatabaseResponse`](https://github.com/googleapis/nodejs-spanner/blob/7046d0b22fb1bdf8996887422ec9cf833d38ac37/src/instance.ts#L1153) translates the Database protos into Database objects.
2. Each Database object contains a SessionPool which is initialized in the [constructor](https://github.com/googleapis/nodejs-spanner/blob/7046d0b22fb1bdf8996887422ec9cf833d38ac37/src/database.ts#L366).

Is it an issue that we potentially create lots of sessions unnecessarily for all results of a ListDatabases call?

@olavloite assigning to you since you've looked at the pool implementation more recently.

olavloite:
Calling `instance.getDatabases()` would return a list of `Database` instances. Each of these databases would be initialized with a default session pool containing a minimum of 25 sessions, potentially creating a lot of sessions unintentionally.

Fixes #1227


test/spanner.ts
```diff
@@ -2579,6 +2579,11 @@ describe('Spanner with mock server', () => {
     it('should list databases', async () => {
       const [databases] = await instance.getDatabases();
       assert.strictEqual(databases.length, 2);
+      // Assert that listing the databases does not cause a session pool to be
+      // initialized for the databases.
+      for (const db of databases) {
+        assert.strictEqual((db.pool_ as SessionPool).size, 0);
```
stephenplusplus, line 2585: Can we relocate an adaptation of this test to `test/instance.ts`, to confirm when the Database instance is created, the `{min: 0}` object is passed?

stephenplusplus (changes requested):

skuruppu (approved):

test/spanner.ts
olavloite: I've added the assertion to one of the existing tests here: https://github.com/googleapis/nodejs-spanner/blob/7821ac5a5e1d02f9f64cef2747f97a6ce8fed4c1/test/instance.ts#L1038

AVaksman (approved):

olavloite:
@stephenplusplus Friendly ping on this PR. Would you mind taking another look? Your request for change is currently blocking merging this.

skuruppu:
@stephenplusplus friendly ping again. If you could take another look, that would be awesome.

stephenplusplus (approved):

PR #1238, opened 2020-09-15
docs(samples): updated NUMERIC samples

skuruppu:
Added specific samples for NUMERIC instead of modifying the existing datatypes samples.

### DIFF
--- a/samples/datatypes.js
+++ b/samples/datatypes.js
@@ -45,7 +45,6 @@ async function createVenuesTable(instanceId, databaseId, projectId) {
         LastContactDate        Date,
         OutdoorVenue           BOOL,
         PopularityScore        FLOAT64,
-        Revenue                NUMERIC,
         LastUpdateTime TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true)
       ) PRIMARY KEY (VenueId)`,
   ];
@@ -103,7 +102,6 @@ async function insertData(instanceId, databaseId, projectId) {
       LastContactDate: '2018-09-02',
       OutdoorVenue: false,
       PopularityScore: 0.85543,
-      Revenue: Spanner.numeric('215100.10'),
       LastUpdateTime: 'spanner.commit_timestamp()',
     },
     {
@@ -115,7 +113,6 @@ async function insertData(instanceId, databaseId, projectId) {
       LastContactDate: '2019-01-15',
       OutdoorVenue: true,
       PopularityScore: 0.98716,
-      Revenue: Spanner.numeric('1200100.00'),
       LastUpdateTime: 'spanner.commit_timestamp()',
     },
     {
@@ -127,7 +124,6 @@ async function insertData(instanceId, databaseId, projectId) {
       LastContactDate: '2018-10-01',
       OutdoorVenue: false,
       PopularityScore: 0.72598,
-      Revenue: Spanner.numeric('390650.99'),
       LastUpdateTime: 'spanner.commit_timestamp()',
     },
   ];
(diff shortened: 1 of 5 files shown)


samples/datatypes.js
```diff
@@ -611,6 +607,97 @@ async function queryWithTimestamp(instanceId, databaseId, projectId) {
...
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const projectId = 'my-project-id';
+  // const instanceId = 'my-instance';
+  // const databaseId = 'my-database';
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
```
JustinBeckwith, line 624: So typically, we don't bother passing `projectId` in like this, and rather rely on the library figuring it out on it's own.  There are places we need it, but I kind of think we should keep the constructor parameterless where we can.

samples/datatypes.js
```diff
@@ -611,6 +607,97 @@ async function queryWithTimestamp(instanceId, databaseId, projectId) {
   // [END spanner_query_with_timestamp_parameter]
 }
 
+async function addNumericColumn(instanceId, databaseId, projectId) {
```
JustinBeckwith, line 610: For all new samples, we're really trying to keep to one snippet per file:
https://g3doc.corp.google.com/company/teams/cloud-devrel/dpe/samples/nodejs.md?cl=head

samples/system-test/spanner.test.js
```diff
@@ -774,16 +774,38 @@ describe('Spanner', () => {
...
-  // query_with_numeric_parameter
-  it('should use a NUMERIC query parameter to query record from the Venues example table', async () => {
+  // add_numeric_column
+  it('should add a Revenue column to Venues example table', async () => {
     const output = execSync(
-      `${datatypesCmd} queryWithNumeric ${INSTANCE_ID} ${DATABASE_ID} ${PROJECT_ID}`
+      `${datatypesCmd} addNumericColumn "${INSTANCE_ID}" "${DATABASE_ID}" ${PROJECT_ID}`
+    );
+
+    assert.match(
```
JustinBeckwith, line 783: If you don't need a regex here, you can use `assert.include`:
```js
assert.include(output, `Waiting for ...`);
```

JustinBeckwith (commented):
The feedback here is mostly `nit` in nature, and maybe stuff we want to try applying more broadly, and not just in this PR.  

Other issue I see here - when you use the `fix:` conventional commit prefix, that signals that this change should trigger a release of the library.  But really - there are no material changes to the shipped code, just the samples.  For that, I would suggest using a `docs(samples): ` prefix :)

olavloite (approved):
From a purely Spanner - Numeric sample point of view this LGTM (but note the comments from @JustinBeckwith )

samples/datatypes.js
skuruppu: Ah yes, that makes sense. I've now moved the new samples into their own files.

samples/datatypes.js
skuruppu: Oh interesting. Yeah I suppose we have the GCLOUD_PROJECT env var that could be read. I understood this as, customers may not know how to set that (or maybe they're not comfortable using env vars) so we explicitly set them.

But if it is against the style guide, we can do a repo-wide cleanup of this.

I'll leave it as it is for now, just for consistency with how existing samples are set up.

samples/system-test/spanner.test.js
skuruppu: Thanks, that's much neater.

skuruppu:
Thanks so much for the review and feedback @olavloite and @JustinBeckwith.

JustinBeckwith (approved):

PR #1250, opened 2020-10-07
feat: Point In Time Recovery (PITR)

skuruppu:


samples/database-create-with-version-retention-period.js
```diff
@@ -0,0 +1,68 @@
...
+    projectId: projectId,
+  });
+
+  const instance = spanner.instance(instanceId);
+  const database = instance.database(databaseId);
+
+  try {
+    // Creates a new database.
+    console.log(`Creating database ${instance.formattedName_}.`);
+    const versionRetentionStatement = `
```
olavloite, line 44: nit: I think we should add a short comment here on the fact that the `database.create()` method does not support specifying a retention period, but that we execute it as an extra `ALTER DATABASE` statement after the database has been created. (I had to read the code a couple of times before I really saw that, as the sample and function is called `createDatabaseWithVersionRetentionPeriod`).

olavloite (approved):

samples/database-create-with-version-retention-period.js
skuruppu: Yes, completely agree that this is not obvious. I added a comment to describe what's happening. Thanks for the suggestion.

thiagotnunes (approved):
LGTM

skuruppu:
PTAL @thiagotnunes, I updated the backup/restore samples to include `version_time`.

thiagotnunes (approved):
LGTM

src/backup.ts
```diff
@@ -178,6 +181,10 @@ class Backup {
         name: this.formattedName_,
       },
     };
+    if ('versionTime' in options) {
+      reqOpts.backup!.versionTime =
+        Spanner.timestamp(options.versionTime).toStruct();
+    }
```
skuruppu, line 187: PTAL this file since I had to add hand-written code to support `versionTime` when creating a backup.

thiagotnunes (approved):

src/backup.ts
thiagotnunes: LGTM. Should we add `versionTime` to the `@example` in the comments?

skuruppu: Good idea. Done.

PR #1253, opened 2020-10-12
feat!: inline BeginTransaction with first statement

olavloite:
Enables `BeginTransaction` to be inlined with the first request of a transaction, which means that the session pool does not need to keep track of both read and read/write sessions, and does not need to prepare a certain fraction of the sessions with a read/write transaction.

This PR only adds the inline option and makes it the default. It does not remove the code for session preparing in the pool, but the feature is disabled by default.


JustinBeckwith:
@skuruppu this PR is about to have it's first birthday 🎂  Can we close this out?

skuruppu:
@JustinBeckwith good point, sorry that I haven't got to this. I will try to review this within the next couple of weeks.

test/spanner.ts
```diff
@@ -2424,6 +2477,674 @@ describe('Spanner with mock server', () => {
...
+        const response2 = tx.run(selectSql);
+        await Promise.all([response1, response2]);
+        await tx.commit();
+      });
+      assertInlinedBegin(spannerMock);
+      assertSecondRequestUsesTransactionId(spannerMock);
+      await database.close();
+    });
+
+    it('should retry and use explicit BeginTransaction if the first statement fails', async () => {
```
olavloite, line 2621: @asthamohta This test

surbhigarg92:
Closing the PR as the new implementation for ILB has been merged.

PR #1274, opened 2020-11-27
feat: customer-managed encryption keys

skuruppu:
Supports creating databases and backups, and restoring databases with a customer-managed encryption key.

### DIFF
--- a/samples/database-create-with-encryption-key.js
+++ b/samples/database-create-with-encryption-key.js
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2021 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/samples/database-create-with-encryption-key.js
+++ b/samples/database-create-with-encryption-key.js
@@ -41,8 +41,6 @@ async function createDatabaseWithEncryptionKey(
   // Gets a reference to a Cloud Spanner instance
   const instance = spanner.instance(instanceId);
 
-  // Note: Cloud Spanner interprets Node.js numbers as FLOAT64s, so they
-  // must be converted to strings before being inserted as INT64s
   const request = {
     encryptionConfig: {
       kmsKeyName: keyName,
--- a/samples/database-create-with-encryption-key.js
+++ b/samples/database-create-with-encryption-key.js
@@ -1,0 +1,72 @@
+// Copyright 2017 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+async function createDatabaseWithEncryptionKey(
+  instanceId,
+  databaseId,
+  projectId,
+  keyName
+) {
+  // [START spanner_create_database_with_encryption_key]
+  // Imports the Google Cloud client library
+  const {Spanner} = require('@google-cloud/spanner');
+
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const projectId = 'my-project-id';
+  // const instanceId = 'my-instance';
+  // const databaseId = 'my-database';
+  // const keyName =
+  //   'projects/my-project-id/eur5/keyRings/my-key-ring/cryptoKeys/my-key';
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  // Gets a reference to a Cloud Spanner instance
+  const instance = spanner.instance(instanceId);
+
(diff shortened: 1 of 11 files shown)


samples/database-create-with-encryption-key.js
```diff
@@ -0,0 +1,72 @@
...
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  // Gets a reference to a Cloud Spanner instance
+  const instance = spanner.instance(instanceId);
+
+  // Note: Cloud Spanner interprets Node.js numbers as FLOAT64s, so they
```
olavloite, line 44: nit: It seems that this comment is a left-over from another sample

olavloite (approved):

src/database.ts
```diff
@@ -1879,28 +1895,34 @@ class Database extends common.GrpcServiceObject {
    */
   restore(
     backupName: string,
-    optionsOrCallback?: CallOptions | RestoreDatabaseCallback,
+    optionsOrCallback?: RestoreOptions | RestoreDatabaseCallback,
```
stephenplusplus, line 1898: Just calling this out to make sure the PR is merged as a breaking change.

```diff
-database.restore({ timeout: 0 });
+database.restore({ gaxOptions: { timeout: 0 } });
```

stephenplusplus (dismissed):

src/database.ts
skuruppu: Oh my bad. I was meant to send out a separate PR for this to group all the breaking changes together (e.g. #1254). I will then rebase this PR so that this explicitly doesn't need to be marked as a breaking change.

skuruppu: Sent out #1278. Will rebase once we're able to merge that in together with all the other breaking changes (likely in early January).

samples/database-create-with-encryption-key.js
skuruppu: Good spotting. Fixed now.

src/database.ts
skuruppu: The breaking change is no longer needed.

skuruppu:
PTAL @olavloite

src/database.ts
```diff
@@ -1874,34 +1890,53 @@ class Database extends common.GrpcServiceObject {
    * const instance = spanner.instance('my-instance');
    * const database = instance.database('my-database');
    * const backupName = 'projects/my-project/instances/my-instance/backups/my-backup';
-   * const [, restoreOperation] = await database.restore(backupName);
+   * const [, restoreOperation] = await database.restore(
+   *   backupName,
+   *   {
+   *     encryptionConfig: {
```
olavloite, line 1896: Depending on how often we think that users will use encrypted backups, I would consider leaving the encryption information out of this example. As it is now, it at first hand seems to be something that is required for all restore operations, while it is only required if you actually have an encrypted backup.

samples/database-create-with-encryption-key.js
```diff
@@ -0,0 +1,70 @@
...
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  // Gets a reference to a Cloud Spanner instance
+  const instance = spanner.instance(instanceId);
+
+  const request = {
+    encryptionConfig: {
+      kmsKeyName: keyName,
```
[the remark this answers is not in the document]
skuruppu: Yeah this is confusing. But `CreateDatabaseRequest` has an `EncryptionConfig` field which only lets you specify the [`kms_key_name`](https://github.com/googleapis/googleapis/blob/8d17d8fafbb87ac64bb3179b99ac34ed41375a51/google/spanner/admin/database/v1/common.proto#L56).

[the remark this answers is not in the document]
skuruppu: I think this might be because by default, it uses `GOOGLE_DEFAULT_ENCRYPTION`. But if the customer specifies an encryption key, then it has to be `CUSTOMER_MANAGED_ENCRYPTION`.

samples/system-test/spanner.test.js
```diff
@@ -893,6 +1013,29 @@ describe('Spanner', () => {
     );
   });
 
+  // restore_backup_with_encryption_key
+  it('should restore database from a backup using an encryption key', async function () {
```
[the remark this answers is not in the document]
skuruppu: I wish we could do that but I guess the samples were designed such that we take a separate backup and do a separate restore for CMEK. That means I have to test them separately here :( But if it becomes too flaky over time, then I may have to end up skipping these.

src/database.ts
skuruppu: Yeah, I think it's reasonable to remove this as it's not something you have to specify by default. Sometimes we have multiple examples so I may consider adding a separate one.

BTW, for restoring with an encryption key, it's not a requirement to specify the encryption key if the backup was encrypted. If the backup was encrypted, then the restored database is automatically encrypted with the same key as the backup. You only have to specify the key here if you want to encrypt it with a different key.

src/database.ts
```diff
@@ -1874,34 +1890,53 @@ class Database extends common.GrpcServiceObject {
...
+      reqOpts.encryptionConfig = (options as RestoreOptions).encryptionConfig;
+    }
+
     return this.request(
       {
         client: 'DatabaseAdminClient',
         method: 'restoreDatabase',
         reqOpts,
-        gaxOpts,
+        gaxOpts: gaxOpts,
```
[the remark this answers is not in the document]
skuruppu: Good catch.

PR #1279, opened 2020-12-02
feat: create instances with processing units

skuruppu:
Only one of processing units or nodes should be specified.


olavloite (approved):

skuruppu:
@AVaksman thanks for noticing the discrepancy in the tests. It's fixed now.

PR #1282, opened 2020-12-04
feat: support RPC priority

olavloite:
Adds support for setting RPC priority.
This PR also uses some of the changes that are introduced in googleapis/nodejs-spanner#1254.


skuruppu:
Also adding @AVaksman who is working on tagging which may touch similar code paths.

samples/rpc-priority.js
```diff
@@ -0,0 +1,82 @@
...
+  // Gets a reference to a Cloud Spanner instance and database.
+  const instance = spanner.instance(instanceId);
+  const database = instance.database(databaseId);
+
+  const sql = `SELECT AlbumId, AlbumTitle, MarketingBudget
+          FROM Albums
+          ORDER BY AlbumTitle`;
+
+  try {
+    const [rows] = await database.run({
```
skuruppu, line 42: It would be great to add a comment to say that we're explicitly lowering the priority of this query by setting this option.

src/transaction.ts
```diff
@@ -1253,15 +1261,15 @@ export class Transaction extends Dml {
 
   batchUpdate(
     queries: Array<string | Statement>,
-    gaxOptions?: CallOptions
+    options?: BatchUpdateOptions
```
skuruppu, line 1264: Because of changes like this, we may actually have to make the type changes separately in time for the breaking change release we'll do in early January. Do you think that would be ok?

skuruppu (approved):

src/transaction.ts
olavloite: Yeah, I'll extract this change to a separate PR.

olavloite: Following @JustinBeckwith's suggestion in https://github.com/googleapis/nodejs-spanner/pull/1278#discussion_r540352588, we could also change this parameter (and others) to `BatchUpdateOptions | CallOptions`. The definition of this and the other methods would then look something like this:

```typescript
  batchUpdate(
    queries: Array<string | Statement>,
    optionsOrCallback?: BatchUpdateOptions | CallOptions | BatchUpdateCallback,
    cb?: BatchUpdateCallback
  ): Promise<BatchUpdateResponse> | void {
    const options =
      typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
    const callback =
      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb!;
    const gaxOpts =
      'gaxOptions' in options
        ? (options as BatchUpdateOptions).gaxOptions
        : options;

  ...

  }
```

(@JustinBeckwith Please feel free to correct me if I misunderstood your comment)

olavloite: @JustinBeckwith Friendly ping on this. Would you mind giving us your input/opinion on the above?

skuruppu: I tried this out for CommitStats. It works to avoid the breaking change. But there is a caveat that @AVaksman pointed at which is that if options doesn't have `gaxOptions` then we pass the `options` as a whole even if it is not a `CallOptions`.

Is this ok?

I tried to use a [Discriminating Union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) but then we need the user to specify the descriminator field, which would end up being a breaking change anyway.

skuruppu: I assume there is no other way to do this than the idea you suggested @olavloite . I'll be going ahead with this implementation for CommitStats. So if you could do the same in this PR, then that works.

JustinBeckwith: Yeah - we really want to avoid breaking changes in these library unless it's unreasonable to avoid.  We're starting to evaluate the possibility of backporting fixes to prior major releases, so the cost of these is going to go waaaaay up.  I would recommend avoiding the break if you can.

samples/rpc-priority.js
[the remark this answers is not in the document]
skuruppu: Ack @JustinBeckwith, I didn't realize. We'll do this as a separate cleanup effort. Tracked in #1295.

samples/rpc-priority.js
```diff
@@ -0,0 +1,82 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+async function queryWithRpcPriority(instanceId, databaseId, projectId) {
```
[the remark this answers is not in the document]
olavloite: Yes, eventually, but they are not known yet.

samples/rpc-priority.js
olavloite: Done.

src/transaction.ts
olavloite: I've changed the different options arguments to `BatchUpdateOptions | CallOptions` (and similar) and added type checking where necessary. That should make this a non-breaking change.

samples/rpc-priority.js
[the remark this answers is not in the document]
olavloite: Thanks for the information. I've updated this specific sample based on the example you provided. I unfortunately am not allowed to access the documentation that you linked.

samples/rpc-priority.js
```diff
@@ -0,0 +1,82 @@
...
+    const [rows] = await database.run({
+      sql,
+      requestOptions: {
+        priority: spanner.v1.RequestOptions.Priority.PRIORITY_LOW,
+      },
+    });
+
+    rows.forEach(row => {
+      const json = row.toJSON();
+      const marketingBudget = json.MarketingBudget
```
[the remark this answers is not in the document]
olavloite: Good point, it's not (lazy copy-paste from a different sample). I removed that specific part.

samples/rpc-priority.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+  } catch (err) {
+    console.error('ERROR:', err);
+  } finally {
+    // Close the database when finished.
+    await database.close();
+  }
+}
+
+require('yargs')
```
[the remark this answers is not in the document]
olavloite: Changed to positional parameters.

src/transaction.ts
```diff
@@ -60,6 +61,7 @@ export interface RequestOptions {
 }
 
 export interface CommitOptions {
+  requestOptions?: IRequestOptions;
```
AVaksman, line 64: The `IRequestOptions` interface will contain RPC priority options and the tagging options. However `CommitOptions` should not support tagging options as a parameter.
I am suggesting either to define a separate interface with just a `priority` option or another way would be to use some ts magic

```ts
requestOptions?: Pick<IRequestOptions, 'priority'>;
```

src/transaction.ts
```diff
@@ -123,6 +126,7 @@ export interface BatchUpdateCallback {
     response?: spannerClient.spanner.v1.ExecuteBatchDmlResponse
   ): void;
 }
+export type BatchUpdateOptions = CommitOptions;
```
AVaksman, line 129: Does `BatchUpdateOptions` suppose to accept `returnCommitStats`?

src/transaction.ts
AVaksman: I think we also need to think of a different name, since we have a same name parameter in `ExecuteSqlRequest` and `BatchUpdateOptions` interfaces, in which "requestOptions" will support `priority` and a `requestTag`.

AVaksman: perhaps "commitRequestOptions" and "transactionRequetOptions", any suggestions are most welcome.

src/transaction.ts
AVaksman: Also to have in mind that "requestOptions" (or "transactionRequetOptions" or better name) of `BatchUpdateOptions` will support `requestTag` in the future.

AVaksman: Once proto changes for tagging is merged something like this could be used
```ts
requestOptions?: Omit<IRequestOptions, 'transactionTag'>
```

src/transaction.ts
```diff
@@ -76,6 +78,7 @@ export interface ExecuteSqlRequest extends Statement, RequestOptions {
   partitionToken?: Uint8Array | string;
   seqno?: number;
   queryOptions?: IQueryOptions;
+  requestOptions?: IRequestOptions;
```
AVaksman, line 81: Probably should not use the same property name as in `CommitOptions` interface since `CommitOptions.requestOptions` will only support `priority` parameter and in both `ExecuteSqlRequest` and `BatchUpdateOptions` same property name will support `requestTag` parameter

skuruppu:
I merged the codegen PR so once we rebase, we can continue the discussion.

src/transaction.ts
skuruppu: I guess that means we have to explicitly create a `BatchUpdateOptions` interface as it's not the same as `CommitOptions`.

skuruppu:
@olavloite if you could make the changes such that this PR can be merged before #1286, that would be great.

olavloite:
> I merged the codegen PR so once we rebase, we can continue the discussion.

@skuruppu 

I think that there's something strange going on with that codegen. The `RequestOptions` message has been added to the [proto files](https://github.com/googleapis/nodejs-spanner/blob/d8518581b5b9ef7f001b7ede22d680f9c888e592/protos/google/spanner/v1/spanner.proto#L429), but it has not actually regenerated the proto [Typescript](https://github.com/googleapis/nodejs-spanner/blob/master/protos/protos.d.ts) and [Javascript](https://github.com/googleapis/nodejs-spanner/blob/master/protos/protos.js) files. So the build is currently failing because it cannot find the `RequestOptions` definition.

It seems that the regenerate step is missing the actual `protoc` step. Any idea why?

AVaksman:
> > I merged the codegen PR so once we rebase, we can continue the discussion.
> 
> @skuruppu
> 
> I think that there's something strange going on with that codegen. The `RequestOptions` message has been added to the [proto files](https://github.com/googleapis/nodejs-spanner/blob/d8518581b5b9ef7f001b7ede22d680f9c888e592/protos/google/spanner/v1/spanner.proto#L429), but it has not actually regenerated the proto [Typescript](https://github.com/googleapis/nodejs-spanner/blob/master/protos/protos.d.ts) and [Javascript](https://github.com/googleapis/nodejs-spanner/blob/master/protos/protos.js) files. So the build is currently failing because it cannot find the `RequestOptions` definition.
> 
> It seems that the regenerate step is missing the actual `protoc` step. Any idea why?


I've tried to regenerate locally and this is what I get

```
DEBUG:synthtool:Compiling protos...
(node:14631) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open 'protos/google/rpc/error_details.proto'
    at Object.openSync (fs.js:462:3)
    at Object.readFileSync (fs.js:364:35)
    at fetch (/www/nodejs-spanner/node_modules/protobufjs/src/root.js:172:34)
    at Root.load (/www/nodejs-spanner/node_modules/protobufjs/src/root.js:206:13)
    at Root.loadSync (/www/nodejs-spanner/node_modules/protobufjs/src/root.js:247:17)
    at main (/www/nodejs-spanner/node_modules/protobufjs/cli/pbjs.js:235:18)
    at internal/util.js:297:30
    at new Promise (<anonymous>)
    at main (internal/util.js:296:12)
    at compileProtos (/www/nodejs-spanner/node_modules/google-gax/build/tools/compileProtos.js:212:11)
(node:14631) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:14631) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

skuruppu:
Hmmmm I'm not sure what's going on. I asked the Node folks for help.

skuruppu:
The issue should be fixed now, please rebase again.

olavloite:
@skuruppu Thanks for getting the proto changes fixed and merged. It works now.

@AVaksman I've rebased and updated the PR based on our latest discussions. Would you mind taking a look and see if this would work for you as well for tagging?

src/transaction.ts
```diff
@@ -61,7 +61,7 @@ export interface RequestOptions {
 }
 
 export interface CommitOptions {
-  requestOptions?: IRequestOptions;
+  requestOptions?: Omit<IRequestOptions, 'requestTag'>;
```
AVaksman, line 64: I believe it should be 
```ts
requestOptions?: Pick<IRequestOptions, 'priority'>;
```
here

olavloite: No, I don't think that's correct. `CommitOptions` are used for blind writes such as `database.table('some-table').insert(..)`, and those should support setting a transaction tag.

olavloite: (After our conversation earlier today): I've removed `transaction-tag` from `CommitOptions`.

samples/rpc-priority.js
```diff
@@ -0,0 +1,73 @@
...
+  // const projectId = 'my-project-id';
+  // const instanceId = 'my-instance';
+  // const databaseId = 'my-database';
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  async function queryWithRpcPriority(instanceId, databaseId) {
```
skuruppu, line 35: I forgot to mention this earlier. There are no code samples for this feature that will get pulled in from docs. So there's a question of whether we should remove the sample and just add an integration test. There's no way to test if the priority is being accepted by the backend but at least you're able to set it.

What do you think?

samples/system-test/spanner.test.js
```diff
@@ -489,6 +490,18 @@ describe('Spanner', () => {
     );
   });
 
+  // query with RPC priority
+  // TODO: Enable when RPC Priority has been released.
+  it.skip('should use request options', async () => {
```
skuruppu, line 495: If we choose to keep the sample then you can unskip this now.

skuruppu (approved):

samples/system-test/spanner.test.js
olavloite: Argh. I missed this one and went ahead and merged without it. I've opened a new PR for enabling it here: https://github.com/googleapis/nodejs-spanner/pull/1338

PR #1286, opened 2020-12-10
feat: add tagging support

AVaksman:
- [ ] Ensure the tests and linter pass


src/session-pool.ts
```diff
@@ -95,7 +95,12 @@ export interface SessionPoolInterface extends EventEmitter {
    * @name SessionPoolInterface#getWriteSession
    * @param {GetWriteSessionCallback} callback The callback function.
    */
+  getWriteSession(
+    transactionRequestOptions: Pick<TxnRequestOptions, 'transactionTag'>,
```
skuruppu, line 99: I like the use of `Pick` here to get the value we want.

src/transaction.ts
```diff
@@ -65,12 +65,18 @@ export interface Statement {
   types?: Type | {[param: string]: Value};
 }
 
+export interface TxnRequestOptions {
```
skuruppu, line 68: Yeah finding a good name for this is hard. I was half wondering if we could just add these new fields to the existing `RequestOptions`. But I don't know how that affects the types that extend from `RequestOptions`.

skuruppu (commented):
I need to have a look to verify the cases for using transaction tag vs request tag. The approach is good. Please also look at the comment that Justin made in https://github.com/googleapis/nodejs-spanner/pull/1278#discussion_r540352588.

src/table.ts
```diff
@@ -45,23 +46,28 @@ export type DropTableCallback = UpdateSchemaCallback;
...
 export type UpdateRowsResponse = CommitResponse;
-export type UpdateRowsOptions = CommitOptions;
+export type UpdateRowsOptions = DeleteRowsOptions;
 
 export type UpsertRowsCallback = CommitCallback;
 export type UpsertRowsResponse = CommitResponse;
-export type UpsertRowsOptions = CommitOptions;
+export type UpsertRowsOptions = DeleteRowsOptions;
+
+type MutateRowsOptions = DeleteRowsOptions;
```
olavloite, line 69: Wouldn't it be more natural to define `MutateRowsOptions` as the extension of `CommitOptions` and then declare `DeleteRowsOptions = MutateRowsOptions` etc.?

So like this:

```typescript

interface MutateRowsOptions extends CommitOptions {
  transactionRequestOptions?: Pick<TxnRequestOptions, 'transactionTag'>;
}

export type DeleteRowsCallback = CommitCallback;
export type DeleteRowsResponse = CommitResponse;
export type DeleteRowsOptions = MutateRowsOptions;

export type InsertRowsCallback = CommitCallback;
export type InsertRowsResponse = CommitResponse;
export type InsertRowsOptions = MutateRowsOptions;

export type ReplaceRowsCallback = CommitCallback;
export type ReplaceRowsResponse = CommitResponse;
export type ReplaceRowsOptions = MutateRowsOptions;

export type UpdateRowsCallback = CommitCallback;
export type UpdateRowsResponse = CommitResponse;
export type UpdateRowsOptions = MutateRowsOptions;

export type UpsertRowsCallback = CommitCallback;
export type UpsertRowsResponse = CommitResponse;
export type UpsertRowsOptions = MutateRowsOptions;
```

src/transaction.ts
```diff
@@ -144,6 +155,10 @@ export interface RunUpdateCallback {
 }
 
 export type CommitCallback = NormalCallback<spannerClient.spanner.v1.ICommitResponse>;
+export interface BeginTransactionOptions {
+  gaxOptions?: CallOptions;
+  transactionRequestOptions?: Pick<TxnRequestOptions, 'transactionTag'>;
```
olavloite, line 160: `BeginTransaction` should not be tagged. From the spec:

> All requests sent using the transaction (excluding BeginTransaction) should include the
transaction tag regardless of whether a request tag has been set.

That would probably also remove the requirement to pass in any transaction options to `getWriteSession`.

src/transaction.ts
```diff
@@ -70,12 +70,18 @@ export interface Statement {
...
+  transactionTag?: string;
+}
+
 export interface ExecuteSqlRequest extends Statement, RequestOptions {
   resumeToken?: ResumeToken;
   queryMode?: spannerClient.spanner.v1.ExecuteSqlRequest.QueryMode;
   partitionToken?: Uint8Array | string;
   seqno?: number;
   queryOptions?: IQueryOptions;
+  txnRequestOptions?: Pick<TxnRequestOptions, 'requestTag'>;
```
olavloite, line 84: I worry about this in combination with #1282 
For RPC priority we add `IRequestOptions` to `ExecuteSqlRequest`, `CommitOptions` and `BatchUpdateOptions`. `IRequestOptions` will eventually (when the proto changes for RPC Prio and tagging have been merged) contain both the tagging options and the RPC priority options. If we keep these two PRs as they are now, we will introduce a very confusing situation, because:
1. A user can specify tags in `txnRequestOptions`
2. A user can specify tags in `IRequestOptions`

I don't think it will be clear to users which one they should use.

So either we need to change #1282 to also introduce a separate options interface for RPC priority, or this PR should be changed into also using `IRequestOptions` on these interfaces. My preference would be the latter to avoid cluttering up the options interfaces with two many different types.

src/transaction.ts
```diff
@@ -969,6 +1029,21 @@ export class Snapshot extends EventEmitter {
...
+   */
+  configureTagOptions(
+    singleUse?: boolean,
+    transactionTag?: string,
+    requestTag?: string
+  ): spannerClient.spanner.v1.IRequestOptions | null {
+    if (!transactionTag && !requestTag) {
+      return null;
+    }
+    return singleUse ? {requestTag} : {transactionTag, requestTag};
```
olavloite, line 1044: I don't think this check is correct:
1. A `singleUse` transaction could in theory be a read/write transaction in the form of a single `Commit` call containing mutations.
2. A non-`singleUse` transaction is not by definition taggable, as it could also be a read-only transaction, which should not include transaction tags.

test/transaction.ts
```diff
@@ -255,6 +311,17 @@ describe('Transaction', () => {
         assert.deepStrictEqual(reqOpts.transaction, expectedTransaction);
       });
 
+      it('shoould set request tag if `singleUse', () => {
```
olavloite, line 314: typo: `should`

test/transaction.ts
```diff
@@ -523,6 +591,19 @@ describe('Transaction', () => {
...
```
olavloite, line 594: typo: `should`

test/transaction.ts
```diff
@@ -141,6 +146,57 @@ describe('Transaction', () => {
...
+        snapshot.request = config => {
+          assert.strictEqual(config.gaxOpts, gaxOptions);
+          done();
+        };
+
+        snapshot.begin(gaxOptions);
+      });
+
+      it('should accept tag options', done => {
+        const options: BeginTransactionOptions = {
```
olavloite, line 163: See also above: `BeginTransaction` should not include a transaction tag.

test/transaction.ts
```diff
@@ -1335,6 +1464,22 @@ describe('Transaction', () => {
         assert.strictEqual(reqOpts.transactionId, id);
       });
 
+      it('should set transactionTag when not single use transaction', done => {
```
olavloite, line 1467: It should only do so if it is a read/write transaction. Read-only transactions should not contain a transaction tag.

src/transaction.ts
AVaksman: > If the library supports hand-crafted transactions, creating the transaction should support setting
> a transaction tag so that it may be used in the BeginTransaction request.

src/transaction.ts
AVaksman: As per documentation
> Single use transactions should not support setting a transaction tag.

In terms of read-only vs read-write transactions
Only `class Transaction` object constructor accepts `transactionRequestOptions` param and passes it to all its methods regardless whether `request tag` is set on per method basis.
`class Snaphot` (a read-only transaction) object does not accept any tags, this way only `request tag` could be set on per method basis.

src/transaction.ts
AVaksman: Otherwise transactions are created at the time `database` is created without transaction tags, so in that case no transaction tag will be passed to `Transaction#begin()`

olavloite: > > If the library supports hand-crafted transactions, creating the transaction should support setting
> > a transaction tag so that it may be used in the BeginTransaction request.

Sorry, I missed that one. So in case of a hand-crafted read/write transaction setting a transaction tag on the `BeginTransaction` request should be supported. This still means that we should not allow this for read-only transactions, which also use this method (and most probably is the most common reason for applications to use this method).


> Otherwise transactions are created at the time `database` is created without transaction tags, so in that case no transaction tag will be passed to `Transaction#begin()`

Correct, which is also one of the reasons why I would say that `getWriteSession()` should never support setting a transaction tag:
1. There is no guarantee that `getWriteSession()` will invoke `BeginTransaction`. If it finds a session that already has a read/write transaction in the pool, it will return that session. The transaction on that session has not been tagged.
2. There is the possibility that `getWriteSession()` does invoke `BeginTransaction` and returns the transaction that is created by that call. That transaction will include the tag.

So the result of calling `getWriteSession()` with a transaction tag is non-deterministic, which will be confusing.

src/transaction.ts
olavloite: > Single use transactions should not support setting a transaction tag.

Hmm... I think the specs are a little ambiguous regarding this (although the above sentence is not ambiguous...). It also states the following:

> __Batch Methods__
Methods that batch Mutations together should support setting a transaction tag. They should
not support setting a request tag. This is because they are sent together as a single
CommitRequest.

The above can apply to both a multi-use and single-use read/write transaction. Single-use read/write transactions are not very well supported in the NodeJS client library (which is not necessarily a bad thing, as they are for advanced use only), but the following code snippet would execute a single-use read/write transaction:

```typescript
const database = instance.database('some-database');
const session = database.session();
await session.create();
const transaction = new Transaction(session);
transaction.upsert('Singers', {SingerId: 1, Name: 'Some Name'});
await transaction.commit();
```

The above transaction will in the current setup not include the transaction tag.


Regarding the transaction tag you're completely right that `Snapshot` does not accept `transactionRequestOptions`. Sorry for missing that one. That does indeed make it impossible to create a read-only transaction that would include a transaction tag on subsequent requests.

It is however possible to call `begin` on a `Snapshot` and include a transaction tag there. That transaction tag will be included in the `BeginTransaction` request, but it will not be used for subsequent requests on the read-only transaction. I think that will be confusing for users. I'm also not sure whether the backend would accept a `BeginTransaction` request for a read-only transaction that includes a transaction tag.

skuruppu:
I merged the codegen PR so once we rebase, we can continue the discussion.

AVaksman:
Rebased

src/transaction.ts
skuruppu: I think we'll go with using `IRequestOptions` with Pick/Omit so we won't need this anymore.

src/transaction.ts
skuruppu: In https://github.com/googleapis/nodejs-spanner/pull/1282#discussion_r605310538, I was saying that I don't think we want to introduce different names for different combinations of options. We want to have a single name that mirror the proto field that it maps to. We'll try to be helpful to the user on which fields can be set by using Pick/Omit. Hope you don't mind.

src/table.ts
```diff
@@ -43,25 +44,30 @@ export type CreateTableCallback = LongRunningCallback<Table>;
...
 export type UpdateRowsCallback = CommitCallback;
 export type UpdateRowsResponse = CommitResponse;
-export type UpdateRowsOptions = CommitOptions;
+export type UpdateRowsOptions = MutateRowsOptions;
 
 export type UpsertRowsCallback = CommitCallback;
 export type UpsertRowsResponse = CommitResponse;
-export type UpsertRowsOptions = CommitOptions;
+export type UpsertRowsOptions = MutateRowsOptions;
+
```
skuruppu, line 70: I think `table.createReadStream()` and `table.read()` below should support setting a request tag.

src/transaction.ts
```diff
@@ -1568,6 +1620,12 @@ export class Transaction extends Dml {
...
 
+    reqOpts.requestOptions = this.configureTagOptions(
+      typeof reqOpts.singleUseTransaction !== 'undefined',
+      this.requestOptions?.transactionTag!,
+      requestOptions
+    );
+
     this.request(
       {
         client: 'SpannerClient',
```
skuruppu, line 1631: The `PartitionedDml` class at the end of this file should support setting a `requestTag`.

skuruppu (changes requested):
I think [createBatchTransaction](https://github.com/googleapis/nodejs-spanner/blob/7055aa17033f26700ef4f4a5e7f2294678485b35/src/database.ts#L609) should support setting a `transactionTag`.

test/transaction.ts
AVaksman: This test is in the Transaction sub-module, Snapshot sub-module does not have such a test

src/session.ts
```diff
@@ -450,9 +451,10 @@ export class Session extends common.GrpcServiceObject {
    * const transaction = session.transaction();
    */
   transaction(
```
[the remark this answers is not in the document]
AVaksman: `requestTag` is supported at 'per-function' level, `createReadStream`,  `read`, `run` and other

src/table.ts
AVaksman: I believe it does. `requestTag` is included in the `ReadRequest` interface that is used with both `table.createReadStream()` and `table.read()`

src/transaction.ts
AVaksman: `requestTag` should be able to be set at function level, `runUpdate` accepts `requestTag` as part of the `ExecuteSqlRequest.requestOptions`

protos/protos.d.ts
```diff
@@ -16178,8 +16178,7 @@ export namespace google {
             REQUIRED = 2,
             OUTPUT_ONLY = 3,
             INPUT_ONLY = 4,
-            IMMUTABLE = 5,
-            UNORDERED_LIST = 6
+            IMMUTABLE = 5
```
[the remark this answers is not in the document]
AVaksman: Reverted the changes

src/transaction.ts
```diff
@@ -998,6 +1026,24 @@ export class Snapshot extends EventEmitter {
...
+    transactionTag?: string,
+    requestOptions = {}
+  ): IRequestOptions | null {
+    if (!singleUse && transactionTag) {
+      (requestOptions as IRequestOptions).transactionTag = transactionTag;
+    }
+
+    return requestOptions!;
+
+    // return singleUse ? {requestTag: requestOptions} : {transactionTag, requestTag: requestOptions};
```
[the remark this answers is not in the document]
AVaksman: Removed

test/transaction.ts
```diff
@@ -1707,6 +1801,66 @@ describe('Transaction', () => {
...
+        transaction.runStream(QUERY);
+      });
+    });
+
+    describe('createReadStream', () => {
+      before(() => {
+        PARTIAL_RESULT_STREAM.callsFake(makeRequest => makeRequest());
+      });
+
+      it('should set transaction tag if not `signleUse`', () => {
```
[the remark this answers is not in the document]
AVaksman: fixed

test/transaction.ts
```diff
@@ -1707,6 +1801,66 @@ describe('Transaction', () => {
...
+        transaction.createReadStream(TABLE);
+
+        const {reqOpts} = REQUEST_STREAM.lastCall.args[0];
+
+        assert.deepStrictEqual(reqOpts.requestOptions, {
+          transactionTag,
+        });
+      });
+
+      it('should not set transaction tag if `signleUse`', () => {
```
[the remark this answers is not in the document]
AVaksman: Fixed typo
Thank you

src/transaction.ts
```diff
@@ -1292,12 +1336,14 @@ export class Transaction extends Dml {
   constructor(
     session: Session,
     options = {} as spannerClient.spanner.v1.TransactionOptions.ReadWrite,
-    queryOptions?: IQueryOptions
+    queryOptions?: IQueryOptions,
+    requestOptions?: Pick<IRequestOptions, 'transactionTag'>
```
olavloite, line 1340: In most cases, this constructor is called by the `_prepareTransaction` method in the session pool here: https://github.com/googleapis/nodejs-spanner/blob/d3a63c839a902ccb7de8c0270921202f18b93370/src/session-pool.ts#L1146

That means that the `requestOptions` will be undefined and not used. This means that any request options with a transaction tag that is passed in to one of the [run transaction methods](https://github.com/googleapis/nodejs-spanner/blob/d3a63c839a902ccb7de8c0270921202f18b93370/src/database.ts#L2603) will not be used, and no transaction tag will actually be used for the requests on the transaction.

skuruppu (commented):
Other than Knut's last comment, the only other concern that's not addressed is [createBatchTransaction](https://github.com/googleapis/nodejs-spanner/blob/7055aa17033f26700ef4f4a5e7f2294678485b35/src/database.ts#L609) not supporting a `transactionTag`. But this might also be tricky given the existing `TimestampBounds` argument.

olavloite:
Closing in favor of #1419

PR #1310, opened 2021-02-22
docs: use earliest version time for backup in sample

olavloite:
Use earliest version time for backup to prevent problems with the version time being in the future if there is a (slight) clock skew
between the backend and the local client.


samples/backups-create.js
```diff
@@ -46,12 +47,12 @@ async function createBackup(instanceId, databaseId, backupId, projectId) {
     const databasePath = database.formattedName_;
     // Expire backup 14 days in the future
     const expireTime = Date.now() + 1000 * 60 * 60 * 24 * 14;
-    // Create a backup of the state of the database now.
-    const versionTime = Date.now();
+    // Create a backup of the state of the database at it earliest possible
```
skuruppu, line 50:
```suggestion
    // Create a backup of the state of the database at the earliest possible
```

skuruppu (approved):

PR #1312, opened 2021-03-01
samples: PITR samples backup fix

thiagotnunes:
Instead of using the earliest version time of the database, uses the current time (from spanner). If we used the earliest version time of the database instead we would be creating an empty backup, since the database we are backing up is not 1 hour old (default version retention period).

### DIFF
--- a/samples/backups-create.js
+++ b/samples/backups-create.js
@@ -15,7 +15,13 @@
 
 'use strict';
 
-async function createBackup(instanceId, databaseId, backupId, projectId, versionTime) {
+async function createBackup(
+  instanceId,
+  databaseId,
+  backupId,
+  projectId,
+  versionTime
+) {
   // [START spanner_create_backup]
   // Imports the Google Cloud client library and precise date library
   const {Spanner} = require('@google-cloud/spanner');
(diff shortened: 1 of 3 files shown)


samples/backups-create.js
```diff
@@ -47,12 +46,17 @@ async function createBackup(instanceId, databaseId, backupId, projectId) {
     const databasePath = database.formattedName_;
     // Expire backup 14 days in the future
     const expireTime = Date.now() + 1000 * 60 * 60 * 24 * 14;
-    // Create a backup of the state of the database at the earliest possible
-    // version time.
+    // Version time is the server's current time
+    const query = {
```
olavloite, line 50: Same as for the Go sample: Would it be better to move this out of the sample code?

Also, you need to [change this back](https://github.com/olavloite/nodejs-spanner/blob/84045a145e4dfabb4d5c5fc1b0393dff11aae92a/samples/system-test/spanner.test.js#L852) to 7.
(See the changes in https://github.com/googleapis/nodejs-spanner/pull/1310/files)

skuruppu: Agree with @olavloite as for the Go sample.

thiagotnunes: Parameterised the version time in the samples

skuruppu (approved):

PR #1318, opened 2021-03-11
fix(sample): run "spanner_batch_client" independently

### LINKED ISSUE #1316: Read data in parallel is not working
Hi,
I was trying to run the example for reading data in parallel: https://cloud.google.com/spanner/docs/reads#read_data_in_parallel

But the nodejs version of code is failing up with an error 
`TypeError: Cannot read property 'request' of undefined
    at new Snapshot (/Users/sudhanshupatel/Documents/Dream11/node_modules/@google-cloud/spanner/build/src/transaction.js:133:32)
    at new BatchTransaction (/Users/sudhanshupatel/Documents/Dream11/node_modules/@google-cloud/spanner/build/src/batch-transaction.js:34:1)
    at Database.batchTransaction (/Users/sudhanshupatel/Documents/Dream11/node_modules/@google-cloud/spanner/build/src/database.js:271:29)`
    
  Failed Statement is : `const transaction = database.batchTransaction(identifier)`

#### Environment details

  - OS: MacOs 11.2.1
  - Node.js : v12.18.4
  - npm version:6.14.6
  - `@google-cloud/spanner` version: 5.5.0

#### Steps to reproduce

  1. Copy the code for creating the partition  https://cloud.google.com/spanner/docs/reads#read_data_in_parallel
  2. execute it.

Thanks!

skuruppu:
To run the sample previously, we had to separately create a batch transaction, then pass in the session/transaction identifier. Now, the sample creates the partitions and executes paritions all in one sample. This also makes the sample consistent with that of other [languages](https://cloud.google.com/spanner/docs/reads#read_data_in_parallel).

Fixes #1316 🦕

### DIFF
--- a/samples/batch.js
+++ b/samples/batch.js
@@ -14,10 +14,9 @@
 
 'use strict';
 
-async function createQueryPartitions(
+async function createAndExecuteQueryPartitions(
   instanceId,
   databaseId,
-  identifier,
   projectId
 ) {
   // [START spanner_batch_client]
@@ -30,7 +29,6 @@ async function createQueryPartitions(
   // const projectId = 'my-project-id';
   // const instanceId = 'my-instance';
   // const databaseId = 'my-database';
-  // const identifier = {};
 
   // Creates a client
   const spanner = new Spanner({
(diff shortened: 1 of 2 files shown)


samples/batch.js
```diff
@@ -40,12 +38,21 @@ async function createQueryPartitions(
...
-  const transaction = database.batchTransaction(identifier);
+  const [transaction] = await database.createBatchTransaction();
 
   const query = 'SELECT * FROM Singers';
 
   const [partitions] = await transaction.createQueryPartitions(query);
   console.log(`Successfully created ${partitions.length} query partitions.`);
+
+  let row_count = 0;
+  for (const partition of partitions) {
```
olavloite, line 49: Would it make sense to let this sample read the different partitions in parallel instead of sequentially, as that is the intention behind batch transactions? The read actions could then be added to an array of promises, and the sample would then wait for all the read operations on the different partitions to finish.

skuruppu: Thanks @olavloite, I think that's a good idea. I modified it to use promises. PTAL.

olavloite (approved):
LGTM

PR #1340, opened 2021-04-09
fix: prevent unhandled promise rejection while projectId or credential not found

### LINKED ISSUE #1333: prevent unhandled promise rejection while default credentials and projectId not set
In case of Project Id or default credential not set it throws the unhandled promise rejection.

laljikanjareeya:
Fixes #1333

### DIFF
--- a/src/session-pool.ts
+++ b/src/session-pool.ts
@@ -277,9 +277,9 @@ export function isDefaultCredentialsNotSetError(
 }
 
 /**
- * Checks whether the given error is a 'Project Id not set in Environment' error.
+ * Checks whether the given error is an 'Unable to detect a Project Id in the current environment' error.
  * @param {Error} error The error to check.
- * @return {boolean} True if the error is a 'Project Id not set in Environment' error, and otherwise false.
+ * @return {boolean} True if the error is an 'Unable to detect a Project Id in the current environment' error, and otherwise false.
  */
 export function isProjectIdNotSetInEnvironmentError(
   error: grpc.ServiceError | undefined
--- a/src/session-pool.ts
+++ b/src/session-pool.ts
@@ -248,9 +248,9 @@ export function isInstanceNotFoundError(
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/session-pool.ts
```diff
@@ -262,6 +262,36 @@ export function isCreateSessionPermissionError(
...
+  error: grpc.ServiceError | undefined
+): boolean {
+  return (
+    error !== undefined &&
+    error.message.includes('Could not load the default credentials')
+  );
+}
+
+/**
+ * Checks whether the given error is a 'Project Id not set in Environment' error.
```
olavloite, line 280: nit: could we make the error description in the comment correspond with the one that is actually being tested in the method?

olavloite (approved):
This seems good to me, but it seems that the CI environment is broken at the moment. There's a lot of (I think) unrelated build errors at the moment. I assume that the tests are green locally(?)

AVaksman (approved):

laljikanjareeya:
> This seems good to me, but it seems that the CI environment is broken at the moment. There's a lot of (I think) unrelated build errors at the moment. I assume that the tests are green locally(?)

Yes All the tests are green locally.

src/session-pool.ts
laljikanjareeya: `Project Id not set in Environment` is actual message getting while Project Id is not set in environment.

olavloite: Then the check below is wrong, because that one reads:

```typescript
    error.message.includes(
      'Unable to detect a Project Id in the current environment'
    )
```

laljikanjareeya: @olavloite  my bad below is the actual message getting while project Id or credential not set:
1.  `Unable to detect a Project Id in the current environment`
2.  `Could not load the default credentials`

I will make changes in method description accordingly.

laljikanjareeya: Done!!

olavloite (approved):

PR #1347, opened 2021-04-20
samples: indexing samples conform to rubric

laljikanjareeya:
This PR Contains Indexing Samples that covers the new Canonical format.
It does not remove the old samples.
Towards googleapis/google-cloud-node#7377


AVaksman (approved):

samples/index-create-storing.js
```diff
@@ -0,0 +1,73 @@
...
+
+      console.log('Added the AlbumsByAlbumTitle2 index.');
+    } catch (err) {
+      console.error('ERROR:', err);
+    } finally {
+      // Close the database when finished.
+      database.close();
+    }
+  }
+  createStoringIndex().catch(console.error);
```
skuruppu, line 66: The examples given don't do this `catch()` call. I would like to understand why it's here.

samples/index-read-data-with-storing.js
```diff
@@ -0,0 +1,84 @@
...
+
+  // Imports the Google Cloud Spanner client library
+  const {Spanner} = require('@google-cloud/spanner');
+
+  // Instantiates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  async function readDataWithStoringIndex() {
```
skuruppu, line 42: Should we add the comment from the original sample?
```
  // "Storing" indexes store copies of the columns they index
  // This speeds up queries, but takes more space compared to normal indexes
  // See the link below for more information:
  // https://cloud.google.com/spanner/docs/secondary-indexes#storing_clause
```

samples/system-test/spanner.test.js
```diff
@@ -440,11 +439,7 @@ describe('Spanner', () => {
 
   it('should respect query boundaries when querying an example table with an index', async () => {
     const output = execSync(
-      `${indexingCmd} queryIndex ${INSTANCE_ID} ${DATABASE_ID} ${PROJECT_ID} -s Ardvark -e Zoo`
-    );
-    assert.match(
-      output,
-      /AlbumId: 1, AlbumTitle: Total Junk, MarketingBudget:/
```
skuruppu, line 447: Why is this check removed?

skuruppu (commented):
Sorry for the huge delay in reviewing this. The change looks good. I have a couple of questions in the comments.

samples/index-create-storing.js
laljikanjareeya: No, we don't need the `catch` block as we are handing `unhandledRejection` on the process.
Removed!

samples/index-read-data-with-storing.js
laljikanjareeya: Added!

samples/system-test/spanner.test.js
laljikanjareeya: Reverted.

skuruppu (approved):

laljikanjareeya:
@skuruppu any idea on `OwlBot Post Processor` Build step. it's taking longer than expected.

skuruppu:
@laljikanjareeya would you be able check the docs failure?

laljikanjareeya:
@skuruppu `docs` test failing because the newly added sample in this PR will not be available in the master branch while docs test checking link on a master branch instead of on the same branch.
Ref:
image

skuruppu:
Ah yeah that problem. I'm following up with the Node team to see how this can be fixed. I'll follow up.

PR #1355, opened 2021-04-29
samples: added comments for timeout settings

skuruppu:
Also updated the timeout values to highlight exponential backoff.

### DIFF
--- a/samples/dml.js
+++ b/samples/dml.js
@@ -654,16 +654,29 @@ async function insertWithCustomTimeoutAndRetrySettings(
   const UNAVAILABLE_STATUS_CODE = 14;
   const retryAndTimeoutSettings = {
     retry: {
+      // The set of error codes that will be retried.
       retryCodes: [DEADLINE_EXCEEDED_STATUS_CODE, UNAVAILABLE_STATUS_CODE],
       backoffSettings: {
         // Configure retry delay settings.
+        // The initial amount of time to wait before retrying the request.
         initialRetryDelayMillis: 500,
-        maxRetryDelayMillis: 64000,
+        // The maximum amount of time to wait before retrying. I.e. after this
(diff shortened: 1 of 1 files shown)


hengfengli (changes requested):
Can we also add some comments for retry delay settings?

skuruppu:
> Can we also add some comments for retry delay settings?

I added the comments for retry delay as well.

@hengfengli can you check if the values make sense? I have a hard time trying to figure out how `initialRetryDelayMillis` relates to `initialRpcTimeoutMillis`, and for max retry delay and timeout, and for the multipliers.

samples/dml.js
```diff
@@ -654,16 +654,27 @@ async function insertWithCustomTimeoutAndRetrySettings(
   const UNAVAILABLE_STATUS_CODE = 14;
   const retryAndTimeoutSettings = {
     retry: {
+      // The set of error codes that will be retried.
       retryCodes: [DEADLINE_EXCEEDED_STATUS_CODE, UNAVAILABLE_STATUS_CODE],
       backoffSettings: {
         // Configure retry delay settings.
+        // The initial amount of time to wait before retrying the request.
         initialRetryDelayMillis: 500,
+        // The request won't be retried after this duration has passed.
```
hengfengli, line 663: `maxRetryDelayMillis` is the max waiting time before retrying. E.g., 

```
500 // waiting time after the first timeout
750 // =500 * 1.5, waiting time after the second timeout
...
64000 // max waiting time after the Nth timeout
64000 // unchanged waiting time after the (N+1)th timeout
```

skuruppu: Ah thanks a lot. That makes a lot more sense now. I updated the comment.

hengfengli (approved):
LGTM. Thanks!

olavloite (approved):

PR #1358, opened 2021-05-05
test: enable numeric integration tests on emulator

olavloite:
The emulator now supports the `NUMERIC` data type.


skuruppu (approved):
Nice catch :)

PR #1397, opened 2021-06-14
fix: reset buffered chunked value before retry

### LINKED ISSUE #1392: TypeError: decoded.map is not a function
#### Environment details

  - OS: Google Container-Optimized OS
  - Node.js version: 14.15 (also experienced on 12.18)
  - npm version: 6.14
  - `@google-cloud/spanner` version: 5.7.0

#### Steps to reproduce
1. ?
2. ?

We're seeing an issue more-or-less identical to this one opened by a coworker of mine a ways back: https://github.com/googleapis/nodejs-spanner/issues/198.  It happens pretty inconsistently, so I'm not sure of how exactly to reproduce it.  The error message looks like:
```
TypeError: decoded.map is not a function at Object.decode (/opt/app/node_modules/@google-cloud/spanner/build/src/codec.js:299:31)
```

Possibly related, we're seeing a separate error message, though much less frequently:
```
TypeError: tail.shift is not a function at Function.mergeLists (/opt/app/node_modules/@google-cloud/spanner/build/src/partial-result-stream.js:239:77)
```

The code in our application looks something like this:
```
// query returns ~90k rows -- each `arrayOfStrings` value consists of 0-4 short strings
const stream = spannerDb.runStream({
  sql: '
    SELECT string1, string2, bool1, arrayOfStrings
    FROM <table>
  ',
  json: true,
});
```

Please let me know what other information may be helpful

olavloite:
Retrying a stream of PartialResultSets must correctly reset to the last state that it had when it received a PartialResultSet with a resume token. That means:
1. The pending value that should be merged with the next result should be reset to the pending value of the last PartialResultSet with a resume token.
2. The stream must correctly check for the existence of a resume token.
3. The stream must ensure that all values that are already in the pipeline must be handled before resetting the stream to the last resume token to ensure that no PartialResultSet is added twice to the stream.

Fixes #1392


olavloite:
@skuruppu I'm not really sure why the lint check is failing on CI. When running the same command locally (`npm run lint`) it says that everything is OK. Do you have any idea why the result on CI is different?

skuruppu:
I pretty much have the same problem. I think the `gts` settings are different locally to what it is on the public repo. I haven't yet figured out exactly what is different and I've just been applying the changes manually based on the lint results on GHActions. But that's really tedious.

@bcoe do you have any advice on how we're meant to setup `gts fix` locally?

skuruppu:
This looks good to me based on the debugging that @olavloite did but I really don't know much about how the streaming results are handled in Node.js. @stephenplusplus I would feel much more comfortable if you could take a quick look to see if the fix makes sense.

AVaksman:
In regards to the lint issue I may suggest running the following commands. It worked for me and displayed same 53 problems message as on CI.
```sh
rm -rf package-lock.json build/ node_modules/
npm install
npm run lint
```

Hope it helps

olavloite:
> In regards to the lint issue I may suggest running the following commands. It worked for me and displayed same 53 problems message as on CI.
> 
> ```shell
> rm -rf package-lock.json build/ node_modules/
> npm install
> npm run lint
> ```
> 
> Hope it helps

That did the trick, thanks :+1:

skuruppu:
@alexander-fenster will you or someone in your team be able to review this PR for us? I'm not sure I understand the streaming APIs well enough to review this properly.

stephenplusplus (approved):

PR #1407, opened 2021-06-25
fix: replace projectId placeholder in formatted names

### LINKED ISSUE #1375: Project ID placeholder in `google_cloud_resource_prefix` header not always replaced
The project ID placeholder used in the `google_cloud_resource_prefix` header is sometimes not being replaced for data client methods leading to `google_cloud_resource_prefix` headers in the following form: 
```
projects/{{projectId}}/instances/<instanceId>/databases/<databaseId>
```

This placeholder should always be replaced with the project ID.

Note: This does not seem to be tied to a particular version of this client, `gax`, `gapic` or `gccl`.

olavloite:
The `{{projectId}}` placeholder was only replaced in the names of the objects that were included in the requests, and not in the headers and formatted names. This would cause the `CLOUD_RESOURCE_HEADER` to be filled with the `{{projectId}}` placeholder instead of the actual value.

As the actual project ID is gotten through an async method, and the constructor and `instance` methods are synchronous, we can only replace the actual values once it is retrieved for the first request.

Fixes #1375


test/spanner.ts
```diff
@@ -135,8 +136,9 @@ describe('Spanner with mock server', () => {
     // TODO(loite): Enable when SPANNER_EMULATOR_HOST is supported.
     // Set environment variable for SPANNER_EMULATOR_HOST to the mock server.
     // process.env.SPANNER_EMULATOR_HOST = `localhost:${port}`;
+    process.env.GOOGLE_CLOUD_PROJECT = 'test-project';
     spanner = new Spanner({
-      projectId: 'fake-project-id',
+      // projectId: '{{projectId}}',
```
skuruppu, line 141: nit: we should uncomment this line?

skuruppu (approved):
Thanks for getting to the bottom of this issue.

test/spanner.ts
olavloite: Done

olavloite:
@skuruppu Can you override the OwlBot post processor check? It does not seem to ever report a result.

skuruppu:
Yeah this owlbot gets stuck all the time. I just added the "owlbot:run" label and it usually passes the second time around. Let's see what it does.

PR #1427, opened 2021-07-12
docs: add samples for tagging

olavloite:
Adds samples for request tags and transaction tags.

### DIFF
--- a/samples/query-tag.js
+++ b/samples/query-tag.js
@@ -1,0 +1,60 @@
+/**
+ * Copyright 2021 Google LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+async function main(instanceId, databaseId, projectId) {
+  // [START spanner_query_tags]
+  // Imports the Google Cloud client library
+  const {Spanner} = require('@google-cloud/spanner');
+
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const projectId = 'my-project-id';
+  // const instanceId = 'my-instance';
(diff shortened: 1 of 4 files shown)


samples/query-tag.js
```diff
@@ -0,0 +1,60 @@
...
+    "Query stats last hour for request tag 'app=cart,env=dev,action=update':"
+  );
+  stats.forEach(row => {
+    console.log(
+      `${row.REQUEST_TAG} ${row.AVG_LATENCY_SECONDS} ${row.AVG_CPU_SECONDS}`
+    );
+  });
+
+  await database.close();
+  // [END spanner_query_tags]
```
skuruppu, line 56: According to the new style, this block has to go in an async function. You can use https://github.com/googleapis/nodejs-spanner/blob/master/samples/index-create-storing.js as a reference.

samples/query-tag.js
```diff
@@ -0,0 +1,60 @@
...
+}
```
skuruppu, line 57: We also need the following here:
```
process.on('unhandledRejection', err => {
  console.error(err.message);
  process.exitCode = 1;
});
```

samples/transaction-tag.js
```diff
@@ -0,0 +1,73 @@
...
+  // const projectId = 'my-project-id';
+  // const instanceId = 'my-instance';
+  // const databaseId = 'my-database';
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  // Gets a reference to a Cloud Spanner instance and database.
```
skuruppu, line 35: Same as above in this file.

samples/query-tag.js
olavloite: I've updated the sample accordingly.

One small question: The https://github.com/googleapis/nodejs-spanner/blob/master/samples/index-create-storing.js sample has a `usage: node createStoringIndex <INSTANCE_ID> <DATABASE_ID> <PROJECT_ID>` entry. Executing that command won't work (as far as I can tell), as `createStoringIndex` is not a file. Shouldn't that be `node index-create-storing.js ...`?

samples/query-tag.js
olavloite: Done.

samples/transaction-tag.js
olavloite: Done.

samples/query-tag.js
skuruppu: Sorry for the late reply. You're absolutely right that it has to say `<file>.js` for it to work.

olavloite:
@skuruppu I've updated the samples according to the most recent specs.

skuruppu (approved):

PR #1438, opened 2021-07-29
feat: add GetInstanceConfig function

olavloite:
Adds a function for getting the details of a specific `InstanceConfig`. This is a simple wrapper around the `GetInstanceConfig` RPC.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -33,7 +33,7 @@ import {
   CreateInstanceResponse,
 } from './instance';
 import {grpc, GrpcClientOptions, CallOptions} from 'google-gax';
-import {google, google as instanceAdmin} from '../protos/protos';
+import {google as instanceAdmin} from '../protos/protos';
 import {
   PagedOptions,
   PagedResponse,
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -857,6 +867,93 @@ class Spanner extends GrpcService {
...
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * spanner.getInstanceConfig().then(function(data) {
+   *   const instanceConfig = data[0];
+   * });
+   */
+  getInstanceConfig(
+    name: string,
+    optionsOrCallback?: CallOptions | GetInstanceConfigCallback,
```
skuruppu, line 926: I might be overthinking it here but should we future-proof the options by making a new interface for `GetInstanceConfigOptions` that includes `gaxOptions`? I feel like recently, we've introduced lots of new options into existing RPCs that made it very tricky to support in Node.js

system-test/spanner.ts
```diff
@@ -933,6 +933,25 @@ describe('Spanner', () => {
           })
         );
     });
+
```
skuruppu, line 936: Sorry if you're still working on this, but it would be good to have some unit tests as it would be checked by the code cov tool.

src/index.ts
olavloite: That's true. I've updated the function definition to use a specific `GetInstanceConfigOptions` that includes gax options.

system-test/spanner.ts
olavloite: I've added a couple of unit tests.

skuruppu (approved):

PR #1445, opened 2021-08-04
fix: adding option to skip back up tests

asthamohta:
set environment variable as following to skip backup tests:
export SKIP_BACKUPS=true


system-test/spanner.ts
```diff
@@ -37,6 +37,7 @@ import {google} from '../protos/protos';
 import CreateDatabaseMetadata = google.spanner.admin.database.v1.CreateDatabaseMetadata;
 import CreateBackupMetadata = google.spanner.admin.database.v1.CreateBackupMetadata;
 
+const SKIP_BACKUPS = process.env.SKIP_BACKUPS;
```
skuruppu, line 40: Minor nit on the name.

```suggestion
const SKIP_BACKUP_TESTS = process.env.SKIP_BACKUP_TESTS;
```

skuruppu (approved):

PR #1368, opened 2021-05-12
feat: add support for JSON data type

olavloite:
Adds support for `JSON` data type.

PR #1456, opened 2021-08-23
feat: add support for JSON data type

olavloite:


olavloite:
(Closing this in favor of #1368. This PR was only intended to test whether the same `UNAUTHENTICATED` error would happen when the PR was created directly in the repo, as opposed to in a fork. As the same error happens in this case as well, there's no need for this PR.)

Issue #1462, opened 2021-08-31
The memory usage seems increasing when the table.insert() error occurs


junichi-tanaka:
I am the original reporter of issue #1313.
Unfortunately, I'm still facing with the issue.
The cause of the problem seemed to be somewhere else, and I was able to reproduce it, so I'd like you to check it out.

As olavloite mentioned in #1313, the memory usage is stable in his script. https://github.com/googleapis/nodejs-spanner/issues/1313#issuecomment-790019913
However, it seems to be leaking slightly when table.insert() throws an error.
I think this is the root cause of our problem.

The results below are with table.upsert()
![image](https://user-images.githubusercontent.com/678655/131481170-4372f314-87e2-4226-92ba-e94a76586bd6.png)

The results below are with table.insert() causes an error
![image](https://user-images.githubusercontent.com/678655/131481260-682b75a1-ab81-47c2-8020-e9fc5b1dcf46.png)

#### Environment details

  - OS: macOS
  - Node.js version: v14.17.1
  - npm version: 7.20.5
  - `@google-cloud/spanner` version: 5.13.1

#### Steps to reproduce

  1. Causing an error in table.insert()
  2. repeat, repeat and repeat

My test script:

```javascript
const {Spanner} = require('@google-cloud/spanner');

const spanner = new Spanner();
const instance = spanner.instance('my-instance');
const database = instance.database('my-database');
const table = database.table('Singers');

main().then(() => console.log('[INFO] Finished'));

async function main() {
    for (let row = 1; row <= 50000; row++) {
        const newRow = {
            SingerId: 1,
            FirstName: 'firstname',
            LastName: 'lastname',
        };
        //await table.upsert(newRow);
        try {
          await table.insert(newRow);
        } catch(err) {
          if (err.code != 6) {    // ignore grpc AlreadyExists
            throw err;
          }
        }
        if (row % 25 === 0) {
            console.log(`[INFO] Rows updated so far: ${row}`);
            if (global.gc) {
                global.gc();
                const used = process.memoryUsage().heapUsed / 1024 / 1024;
                console.log(`[INFO] Memory usage: ${used} MB`);
            }
        }
    }
}
```

Output(with table.upsert()):
```
[INFO] Rows updated so far: 25
[INFO] Memory usage: 17.850181579589844 MB
[INFO] Rows updated so far: 50
[INFO] Memory usage: 17.832420349121094 MB
[INFO] Rows updated so far: 75
[...]

PR #1463, opened 2021-09-02
feat: add option to disable grpc-gcp

olavloite:
Adds an option to disable grpc-gcp as it seems that it causes a higher memory footprint than
when working without.

Towards #1462

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -230,14 +230,18 @@ class Spanner extends GrpcService {
         libName: 'gccl',
         libVersion: require('../../package.json').version,
         scopes,
-        // Enable grpc-gcp support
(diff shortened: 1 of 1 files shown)


skuruppu:
@alexander-fenster @WeiranFang is this expected? Should we be disabling this or can we find another solution?

olavloite:
> @alexander-fenster @WeiranFang is this expected? Should we be disabling this or can we find another solution?

Just so we are all on one page: I would only consider this as a temporary workaround until we know whether it is actually causing the issue described in #1462 and/or until we have a different solution (if applicable).

olavloite:
Closing as this is no longer relevant.

PR #1487, opened 2021-09-30
chore: update proto and tests for benchmarking tests

zoercai:


protos/protos.d.ts
```diff
@@ -16192,7 +16192,8 @@ export namespace google {
             OUTPUT_ONLY = 3,
             INPUT_ONLY = 4,
             IMMUTABLE = 5,
-            UNORDERED_LIST = 6
+            UNORDERED_LIST = 6,
+            NON_EMPTY_DEFAULT = 7
```
skuruppu, line 16196: I'm curious about these changes to the protos files. I think these are auto-generated.

zoercai: Yep they disappeared after pulling the latest main

skuruppu (approved):

PR #1498, opened 2021-10-14
feat: draft of changes for cross region backup

asthamohta:
The files changed by hand are:
1. src/backup.ts
2. src/instance.ts 
3. src/spanner.ts


skuruppu:
@asthamohta I looked through the latest commits and they look good. We'll have to uncomment the commented out sections of the samples tests but I imagine you'd do that in the other PR.

I also ran the unit tests and they all passed for me. Was there something else that failed?

asthamohta:
@skuruppu I wanted to add a unit test similar to this which is already present for backups. I couldn't understand though how it works:
https://github.com/googleapis/nodejs-spanner/blob/44c3476feb04b4979906fec6f38e9196d2a1abc8/test/instance.ts#L1475-L1493

test/instance.ts
```diff
@@ -1644,12 +1645,31 @@ describe('Instance', () => {
...
+        instance.request = config => {
+        assert.strictEqual(config.client, 'DatabaseAdminClient');
+        assert.strictEqual(config.method, 'createBackup');
+        assert.deepStrictEqual(config.reqOpts, expectedReqOpts);
+        assert.notStrictEqual(config.reqOpts, OPTIONS);
+        assert.deepStrictEqual(config.gaxOpts, options.gaxOptions);
+        done();
+      };
+
+      instance.copyBackup(BACKUP_NAME, COPY_BACKUP_NAME, options, assert.ifError);
```
skuruppu, line 1672: I know what the problem is but I don't know enough about the fakes to recommend a solution. The issue is that all calls to backups get intercepted by the FakeBackup class. That class doesn't have a create() method which is why this error is thrown.

So ideally, you would get FakeBackup to return something meaningful when create() is called. You don't want it to actually call DatabaseAdminClient.createBackup RPC because then it's no longer a unit test.

asthamohta: That makes sense Shanika. Let me try to maybe review the PR where these fake backup's were as that may have the way to set it all up.

src/backup.ts
```diff
@@ -352,6 +464,32 @@ class Backup {
...
+  getMaxExpireTime(callback: GetExpireTimeCallback): void;
+  /**
+   * Retrieves the max expiry time of the backup.
+   *
+   * @see {@link #updateExpireTime}
+   * @see {@link #getMetadata}
+   *
+   * @method Backup#getMaxExpireTime
+   * @returns {Promise<external:PreciseDate>} When resolved, contains the
+   *     current expire time of the backup if it exists.
```
skuruppu, line 477:
```suggestion
   *     max expire time of the backup if it exists.
```

PR #1500, opened 2021-10-19
feat: adding support for Cross Region Backups

asthamohta:
Adding support and samples for Cloud Spanner Cross-Region Backups. This feature allows customers to copy the backup to instances in different regions after creating a backup of the database. The following PR adds the following functionalities:

1. Support for copy_backup in the Node client
2. Tests for copy_backup
3. Samples for copy_backup

### DIFF
--- a/src/backup.ts
+++ b/src/backup.ts
@@ -102,7 +102,7 @@ export type ExistsCallback = NormalCallback<boolean>;
 /**
  * The {@link Backup} class represents a Cloud Spanner backup.
  *
- * Create a `Backup` object to interact with or create a Cloud Spanner backup.
+ * Create a `Backup` object to interact with or create a Cloud Spanner backup or copy backup.
  *
  * @class
  *
@@ -111,6 +111,7 @@ export type ExistsCallback = NormalCallback<boolean>;
  * const spanner = new Spanner();
  * const instance = spanner.instance('my-instance');
  * const backup = instance.backup('my-backup');
+ * const sourceName = instance.backup('my-source-backup');
  */
 class Backup {
   id: string;
@@ -156,6 +157,20 @@ class Backup {
    *     for more details.
    */
   /**
+   * @typedef {object} CopyBackupOptions
+   * @property {string|null}
+   *     sourceBackup The source backup path to be copied
+   * @property {string|number|google.protobuf.Timestamp|external:PreciseDate}
+   *     expireTime The expire time of the backup.
+   * @property {google.spanner.admin.database.v1.ICopyBackupEncryptionConfig}
+   *     encryptionConfig An encryption configuration describing the
+   *     encryption type and key resources in Cloud KMS to be used to encrypt
+   *     the copy backup.
+   * @property {object} [gaxOptions] The request configuration options,
+   *     See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions}
+   *     for more details.
+   */
+  /**
    * @typedef {array} CreateBackupResponse
    * @property {Backup} 0 The new {@link Backup}.
    * @property {google.longrunning.Operation} 1 An {@link Operation} object that can be used to check
@@ -163,7 +178,22 @@ class Backup {
    * @property {object} 2 The full API response.
    */
   /**
+   * @typedef {array} CopyBackupResponse
+   * @property {Backup} 0 The new {@link Backup}.
+   * @property {google.longrunning.Operation} 1 An {@link Operation} object that can be used to check
+   *     the status of the request.
+   * @property {object} 2 The full API response.
+   */
+  /**
    * @callback CreateBackupCallback
+   * @param {?Error} err Request error, if any.
+   * @param {Backup} backup The new {@link Backup}.
+   * @param {google.longrunning.Operation} operation An {@link Operation} object that can be used to
+   *     check the status of the request.
+   * @param {object} apiResponse The full API response.
+   */
+  /**
+   * @callback CopyBackupCallback
    * @param {?Error} err Request error, if any.
    * @param {Backup} backup The new {@link Backup}.
    * @param {google.longrunning.Operation} operation An {@link Operation} object that can be used to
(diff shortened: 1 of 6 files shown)


src/backup.ts
```diff
@@ -97,26 +111,36 @@ export type ExistsCallback = NormalCallback<boolean>;
  * const spanner = new Spanner();
  * const instance = spanner.instance('my-instance');
  * const backup = instance.backup('my-backup');
+ * const sourceName = instance.backup('my-source-backup');
```
skuruppu, line 114: I would add a separate example on how to copy a backup. I think you can have multiple `@example` blocks.

test/instance.ts
```diff
@@ -1457,6 +1457,12 @@ describe('Instance', () => {
         },
       ];
 
+      const COPY_BACKUPS = [
+        {
+          expireTime: new PreciseDate(1000),
+        },
+      ];
+
```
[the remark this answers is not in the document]
skuruppu: I think we decided offline that this wasn't necessary so resolving comment.

src/backup.ts
```diff
@@ -179,13 +232,55 @@ class Backup {
...
+   * const sourceBackup = instance.backup('my-source-backup');
+   * const [, sourceBackupOperation] = await sourceBackup.create({
+   *   databasePath: 'projects/my-project/instances/my-instance/databases/my-database',
+   *   expireTime: expireTime,
+   *   versionTime: versionTime,
+   *   encryptionConfig: {
+   *     encryptionType: 'CUSTOMER_MANAGED_ENCRYPTION',
+   *     kmsKeyName: 'projects/my-project-id/my-region/keyRings/my-key-ring/cryptoKeys/my-key',
+   *   },
+   * });
```
skuruppu, line 263: To keep it simple, I would remove this. Assume the source backup already exists and give it some name.

skuruppu (commented):
Some minor nits but looks good.

src/backup.ts
```diff
@@ -56,6 +57,19 @@ export interface CreateBackupOptions {
...
+  metadata: Metadata &
+    databaseAdmin.spanner.admin.database.v1.ICopyBackupMetadata;
+}
+
+export type CopyBackupResponse = [Backup, CopyBackupGaxOperation, IOperation];
+
+export interface CopyBackupOptions
+  extends databaseAdmin.spanner.admin.database.v1.ICopyBackupRequest {
+  gaxOptions?: CallOptions;
+}
```
skuruppu, line 71: @olavloite what are your thoughts on our design choice? For BackupOptions, we created a new object (not extending IBackupRequest). But this ended up being a poor choice because every time a new field was added, we'd have to modify the interface. So I asked @asthamohta to do it this way. The trouble now is that create and copy have different inputs. E.g. expireTime needs to be specified as a proto in the latter.

olavloite: I agree that extending `IBackupRequest` is probably the better choice here. At least our experiences with other features where we used the proto interfaces have been good.

Would it maybe be possible to change the `createBackup` function so it accepts both the current `CreateBackupOptions` and a new one that extends `ICreateBackupRequest`? (And deprecate the current `CreateBackupOptions` definition.) That way this design choice might feel less as a deviation than now.

src/backup.ts
```diff
@@ -97,26 +111,36 @@ export type ExistsCallback = NormalCallback<boolean>;
...
   request: BackupRequest;
+  sourceName: string | undefined;
   metadata?: databaseAdmin.spanner.admin.database.v1.IBackup;
-  constructor(instance: Instance, name: string) {
+  constructor(instance: Instance, name: string, sourceName?: string) {
     this.request = instance.request;
     this.instanceFormattedName_ = instance.formattedName_;
     this.formattedName_ = Backup.formatName_(instance.formattedName_, name);
     this.id = this.formattedName_.split('/').pop() || '';
+    if (sourceName) {
```
olavloite, line 129: nit: I think this `if` statement is redundant. You can just unconditionally assign `sourceName` to `this.sourceName`.

src/backup.ts
```diff
@@ -132,13 +156,34 @@ class Backup {
    *     See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions}
    *     for more details.
    */
+  /**
+   * @typedef {object} CopyBackupOptions
+   * @property {string|null}
+   *     sourceBackup The source backup path to be copied
```
olavloite, line 162:
```suggestion
   *     sourceBackup The full path of the backup to be copied
```

src/backup.ts
```diff
@@ -179,13 +232,55 @@ class Backup {
...
+   *   expireTime: expireTime,
+   *   versionTime: versionTime,
+   *   encryptionConfig: {
+   *     encryptionType: 'CUSTOMER_MANAGED_ENCRYPTION',
+   *     kmsKeyName: 'projects/my-project-id/my-region/keyRings/my-key-ring/cryptoKeys/my-key',
+   *   },
+   * });
+   * // Await completion of the source backup operation.
+   * await sourceBackupOperation.promise();
+   * const copyBackup = instance.backup('my-copy-backup', sourceBackup.formattedName_);
```
olavloite, line 266: I think this should be:
```suggestion
   * const copyBackup = instance.copy_backup('my-copy-backup', sourceBackup.formattedName_);
```

But see also my comment on the `copy_backup` name below.

src/instance.ts
```diff
@@ -249,6 +249,16 @@ class Instance extends common.GrpcServiceObject {
     return new Backup(this, backupId);
   }
 
+  copy_backup(backupId: string, sourceBackupId: string): Backup {
```
olavloite, line 252: I feel that the name of this function in combination with the (order of) arguments is slightly confusing. `copy_backup` sounds like I am creating a copy of a backup, and normally copy methods take the source as the first argument and the destination as the second. Also, other methods in this class use camel casing, so the name should then be `copyBackup`.
But I'm also wondering whether this function should not return a reference to a backup that the user then needs to call `backup.create()` on, for two reasons:
1. Consistency with other methods: The `createDatabase(..)` method in this class does not return a reference to a database that can be created, but instead executes a `CreateDatabase` RPC. The name `copy_backup` sounded to me like this function would also execute an RPC.
2. Is there any other conceivable use case than calling `backup.create()` after calling `copy_backup(..)`?

src/backup.ts
skuruppu: I think we can get it to accept an `ICreateBackupRequest` but it may not work very well because that proto doesn't have a field for gaxOptions. So we'll have to introduce a new interface. Then it might be confusing for users.

As for deprecating `CreateBackupOptions`, I don't think we can do that because that would be a breaking change :(

PR #1593, opened 2022-03-25
feat: integration testing for postgres dialect

asthamohta:


system-test/spanner.ts
```diff
@@ -252,37 +302,60 @@ describe('Spanner', () => {
...
+          params: {
+            a: data[0].Key,
+            b: data[1].Key,
+          },
+        };
+        if (dialect === Spanner.POSTGRESQL) {
+          table = postgreSqlTable;
+          database = PG_DATABASE;
+          query = {
+            sql: `SELECT * FROM ${table.name.toLowerCase()} WHERE "Key" = $1 OR "Key" = $2`,
```
olavloite, line 318: nit: I don't think we need the `.toLowerCase()` here. The table name will be folded to lowercase automatically by the backend.

system-test/spanner.ts
```diff
@@ -1050,23 +1369,45 @@ describe('Spanner', () => {
...
     const TABLE_NAME = 'SingersTest';
-    it('should auto create a database', done => {
-      const database = instance.database(generateName('database'));
+    const autoCreateDatabase = (done, databaseId) => {
+      const database = instance.database(generateName(databaseId));
 
       database.get({autoCreate: true} as GetDatabaseConfig, err => {
         assert.ifError(err);
         RESOURCES_TO_CLEAN.push(database);
         database.getMetadata(done);
```
olavloite, line 1378: nit: Maybe add a check here to verify that the database was created with the correct dialect.

system-test/spanner.ts
```diff
@@ -1116,37 +1457,63 @@ describe('Spanner', () => {
...
         execAfterOperationComplete(err => {
           assert.ifError(err);
 
-          function replaceNewLinesAndSpacing(str) {
-            return str.replace(/\n\s*/g, '').replace(/\s+/g, ' ');
+          function replaceNewLinesAndSpacing(str, dialect) {
+            const schema = str.replace(/\n\s*/g, '').replace(/\s+/g, ' ');
+            if (dialect === Spanner.GOOGLE_STANDARD_SQL) {
+              return schema;
+            } else return schema.toLowerCase();
```
olavloite, line 1470: nit: add curly braces for the else statement

system-test/spanner.ts
```diff
@@ -1221,77 +1610,155 @@ describe('Spanner', () => {
...
+          databaseDialect: Spanner.POSTGRESQL,
+          gaxOptions: GAX_OPTIONS,
+        });
+      await postgreSqlDatabase2CreateOperation.promise();
+
+      const schema = [
+        `CREATE TABLE Albums (
+            AlbumId VARCHAR NOT NULL PRIMARY KEY,
+        AlbumTitle VARCHAR NOT NULL
+    );`,
```
olavloite, line 1653: nit: The formatting seems a little off here

system-test/spanner.ts
```diff
@@ -1317,9 +1784,28 @@ describe('Spanner', () => {
...
+    it('GOOGLE_STANDARD_SQL should have completed a backup', async () => {
+      await completedBackup(
+        googleSqlBackup1,
+        googleSqlBackup1Name,
+        googleSqlDatabase1
+      );
+    });
+
+    it('POSTGRESQL should have completed a backup', async function () {
+      if (IS_EMULATOR_ENABLED) {
```
olavloite, line 1798: nit: here and elsewhere, I think you can remove this, as backup tests should be disabled for both dialects on the emulator

olavloite (approved):
Changes LGTM (with a couple of nits)

This will add a lot of additional integration tests, which means that the build time will increase. Is there any indication on how much?

asthamohta:
> Changes LGTM (with a couple of nits)
> 
> This will add a lot of additional integration tests, which means that the build time will increase. Is there any indication on how much?

Haven't looked into that but for golang we are skipping backup tests. What do you think?

system-test/spanner.ts
asthamohta: Adding thanks

olavloite:
> > Changes LGTM (with a couple of nits)
> > This will add a lot of additional integration tests, which means that the build time will increase. Is there any indication on how much?
> 
> Haven't looked into that but for golang we are skipping backup tests. What do you think?

I think that it would be good to skip backup tests for PostgreSQL. I don't think they add enough value to justify the additional build time (the last successful build with all the tests in this PR was 52 minutes, I think)

PR #1554, opened 2022-01-19
refactor: return GoogleError instead of Error in handwritten library.

rajatbhatta:
- Replacing Error by [GoogleError](https://github.com/googleapis/gax-nodejs/blob/main/src/googleError.ts) for making the error consistent across all GAPIC libraries.

PR #1611, opened 2022-04-18
feat: support customer managed instance configurations

rajatbhatta:
This PR implements the functionality for CMMR Phase 2. Ir provides the following features:

- Create, update, delete user managed instance config.
- Ability to list InstanceConfig operations (currently CreateInstanceConfig operations).


src/index.ts
```diff
@@ -90,6 +96,16 @@ export interface GetInstanceConfigOptions {
 export type GetInstanceConfigResponse = [IInstanceConfig];
 export type GetInstanceConfigCallback = NormalCallback<IInstanceConfig>;
 
+export type GetInstanceConfigOperationsOptions = PagedOptionsWithFilter;
+export type GetInstanceConfigOperationsResponse = PagedResponse<
+  IOperation,
+  instanceAdmin.spanner.admin.instance.v1.IListInstanceConfigOperationsResponse
+>;
+export type GetInstanceConfigOperationsCallback = RequestCallback<
```
olavloite, line 104: I think this should be a `PagedCallback`, as `GetInstanceConfigOperations` returns a paged result

src/index.ts
```diff
@@ -690,6 +717,180 @@ class Spanner extends GrpcService {
     });
   }
 
+  /**
+   * Config for the new instance config.
+   *
+   * @typedef {object} CreateInstanceConfigRequest
+   * @property {string} [displayName] The name of this instance configuration as
+   *     it appears in UIs.
```
olavloite, line 725: nit:
```suggestion
   *     it appears in the user interface.
```

src/index.ts
```diff
@@ -690,6 +717,180 @@ class Spanner extends GrpcService {
...
+    config: CreateInstanceConfigRequest,
+    callback: CreateInstanceConfigCallback
+  ): void;
+  createInstanceConfig(
+    name: string,
+    config: CreateInstanceConfigRequest,
+    callback?: CreateInstanceConfigCallback
+  ): void | Promise<CreateInstanceConfigResponse> {
+    if (!name) {
+      throw new GoogleError('A name is required to create an instance config.');
```
olavloite, line 833: I don't remember the latest status on errors, so just to be sure: Is this consistent with the other errors that we return in this client library?

src/index.ts
```diff
@@ -999,6 +1200,120 @@ class Spanner extends GrpcService {
     );
   }
 
+  /**
+   * Query object for listing backup operations.
```
olavloite, line 1204:
```suggestion
   * Query object for listing instance config operations.
```

src/index.ts
```diff
@@ -999,6 +1200,120 @@ class Spanner extends GrpcService {
...
+   */
+  /**
+   * @callback GetInstanceConfigOperationsCallback
+   * @param {?Error} err Request error, if any.
+   * @param {google.longrunning.IOperation[]} 0 Array of {@link IOperation} instances.
+   * @param {object} nextQuery A query object to receive more results.
+   * @param {object} apiResponse The full API response.
+   */
+  /**
+   * List pending and completed Instance Config operations.
```
olavloite, line 1231:
```suggestion
   * List pending and completed instance config operations.
```

src/instance-config.ts
```diff
@@ -0,0 +1,496 @@
...
+    } as {} as ServiceObjectConfig);
+    this.formattedName_ = formattedName_;
+    this.request = spanner.request.bind(spanner);
+    this.resourceHeader_ = {
+      [CLOUD_RESOURCE_HEADER]: this.formattedName_,
+    };
+  }
+
+  /**
+   * Gets the instance configuration for this InstanceConfig.
```
olavloite, line 165:
```suggestion
   * Gets the metadata from the backend for this InstanceConfig.
```

src/index.ts
```diff
@@ -690,6 +717,180 @@ class Spanner extends GrpcService {
...
+   * @param {CreateInstanceConfigCallback} [callback] Callback function.
+   * @returns {Promise<CreateInstanceConfigResponse>}
+   *
+   * @example
+   * ```
+   * const {Spanner} = require('@google-cloud/spanner');
+   * const spanner = new Spanner();
+   *
+   * const [baseInstanceConfig] =
+   *     await spanner.getInstanceConfig(baseInstanceConfigId);
```
olavloite, line 784: I think that this only gets a reference to the instance config. It does not load the metadata from the backend, which means that the `baseInstanceConfig.replicas` collection will be empty or undefined.

src/instance-config.ts
```diff
@@ -0,0 +1,496 @@
...
+   * @param {string} instanceConfig.name The unique identifier for the instance
+   *     config.
+   * @param {string} instanceConfig.displayName The name of the instance config
+   *     as it appears in UIs.
+   * @param {google.spanner.admin.instance.v1.IReplicaInfo[]} instanceConfig.replicas The replicas used by
+   *     this instance config.
+   * @param {string[]} 0.leaderOptions The possible leader options for this instance config.
+   */
+  /**
+   * Get a specific instance config.
```
olavloite, line 190:
```suggestion
   * Gets the metadata from the backend for this InstanceConfig.
```

src/instance-config.ts
```diff
@@ -0,0 +1,496 @@
...
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb!;
+    const options =
+      typeof optionsOrCallback === 'object'
+        ? optionsOrCallback
+        : ({} as GetInstanceConfigOptions);
+    this.parent.getInstanceConfig(this.id, options, callback);
+  }
+
+  /**
+   * Update the metadata for this instance config. Note that this method follows
```
olavloite, line 239: Maybe add something like: This function can only be called for custom instance configs.

src/instance-config.ts
```diff
@@ -0,0 +1,496 @@
...
+   * @typedef {array} DeleteInstanceConfigResponse
+   * @property {object} 0 The full API response.
+   */
+  /**
+   * @callback DeleteInstanceConfigCallback
+   * @param {?Error} err Request error, if any.
+   * @param {object} apiResponse The full API response.
+   */
+  /**
+   * Delete the instance config.
```
olavloite, line 335: Maybe add something like: This function can only be called for custom instance configs.

olavloite (commented):
Can we add some end-to-end tests for this?

src/index.ts
rajatbhatta: getInstanceConfig is a wrapper over InstanceAdminClient#getInstanceConfig, and makes a call to the backend, which returns the metadata. I ran the sample and verified that `baseInstanceConfig.replicas` collection is non-empty.

src/index.ts
rajatbhatta: Yes, it is consistent with other errors that we return in this client library. [This](https://github.com/googleapis/nodejs-spanner/pull/1554) is the PR where we made it consistent across the handwritten layer.

rajatbhatta:
> Can we add some end-to-end tests for this?

@olavloite: Yes, I'm working on a separate PR(in draft phase) for adding tests.

src/instance-config.ts
rajatbhatta: Done.

src/instance-config.ts
rajatbhatta: Done.

src/index.ts
rajatbhatta: Changed it to PagedCallback to have a more specific callback here (`RequestCallback<T, R>` where R does not extend from void also transforms to `PagedCallback<T, R>`)

olavloite (approved):

rajatbhatta:
@olavloite : I've now added unit tests and system tests for the feature as well to this PR. Requesting review again.

system-test/spanner.ts
```diff
@@ -45,6 +46,10 @@ const spanner = new Spanner({
   projectId: process.env.GCLOUD_PROJECT,
   apiEndpoint: process.env.API_ENDPOINT,
 });
+const spanner1 = new Spanner({
+  projectId: 'span-cloud-testing',
+  apiEndpoint: 'staging-wrenchworks.sandbox.googleapis.com',
+});
```
olavloite, line 52: I assume that this is temporary for the system tests

system-test/spanner.ts
```diff
@@ -131,6 +142,17 @@ describe('Spanner', () => {
       );
       await postgreSqlOperationUpdateDDL.promise();
       RESOURCES_TO_CLEAN.push(PG_DATABASE);
+
```
olavloite, line 145: This probably needs to be skipped if we are testing using the emulator

system-test/spanner.ts
```diff
@@ -1311,15 +1342,27 @@ describe('Spanner', () => {
...
+    it('should list the available instanceConfigs', function (done) {
       spanner.getInstanceConfigs((err, instanceConfigs) => {
         assert.ifError(err);
         assert(instanceConfigs!.length > 0);
         done();
       });
     });
 
-    it('should list the instanceConfigs in promise mode', done => {
```
olavloite, line 1322: Is there a specific reason that we are changing this notation?

test/index.ts
```diff
@@ -30,13 +30,14 @@ import {grpc} from 'google-gax';
...
 import {Duplex} from 'stream';
-import {CreateInstanceRequest} from '../src/index';
+import {CreateInstanceRequest, CreateInstanceConfigRequest} from '../src/index';
 import {
   GetInstanceConfigOptions,
   GetInstanceConfigsOptions,
   GetInstancesOptions,
 } from '../src';
 import {CLOUD_RESOURCE_HEADER} from '../src/common';
+import * as inst from '../src/instance';
```
olavloite, line 40: (Just to verify): Do we need this entire import?

test/instance-config.ts
```diff
@@ -0,0 +1,363 @@
+/*!
+ * Copyright 2017 Google Inc. All Rights Reserved.
```
olavloite, line 2: nit:
```suggestion
 * Copyright 2022 Google LLC All Rights Reserved.
```

olavloite (commented):
Thanks for adding the tests. This looks generally good to me (with some minor nits). All tests are currently failing. Do you know why that is?

system-test/spanner.ts
rajatbhatta: I've changed this notation in order to make it consistent across all tests for instance configs.

Note that if we want to skip a test entirely when using an emulator, accessing `this` inside arrow function(this.skip()) throws an error (TS7041: The containing arrow function captures the global value of 'this'). We changed the notation to a regular function to handle this error.

test/index.ts
rajatbhatta: This is an unused import. Will remove this. Thanks for pointing this out.

system-test/spanner.ts
rajatbhatta: Yes, will remove this.

rajatbhatta:
> All tests are currently failing. Do you know why that is?

I can confirm that the tests are passing locally right now, as currently the auto-generated code from the modified protos are available only locally. The tests should pass once the protos are live and autogenerated code has been merged.

PR #1617, opened 2022-04-19
fix: pin version for nodejs gax-node

asthamohta:


bcoe (approved):

duncanfinney:
@bcoe @asthamohta Was the downgrade from `^2.29.3` to `^2.3.0` purposeful, or is it a typo that is supposed to be an upgrade to `^2.30.0`? 

We are seeing some memory leaks in `google-gax --> @grpc/grpc-js` that we have a hunch are fixed with the newer version from reading some other github tickets.

PR #1687, opened 2022-07-21
samples: Samples for Commands with RPC Priority

surbhigarg92:
This PR adds the sample for using RPC priority.


samples/rpc-priority-PartitionedDML.js
```diff
@@ -0,0 +1,67 @@
+// Copyright 2021 Google LLC
```
olavloite, line 1: nit (here and elsewhere): 2022

Also here and elsewhere: The file names should be all lower case.

samples/rpc-priority-queryPartitions.js
```diff
@@ -0,0 +1,86 @@
...
+    databaseId
+  ) {
+    // Gets a reference to a Cloud Spanner instance and database
+    const instance = spanner.instance(instanceId);
+    const database = instance.database(databaseId);
+    const [transaction] = await database.createBatchTransaction();
+
+    const [partitions] = await transaction.createQueryPartitions({
+      sql: 'SELECT * FROM Singers',
+      requestOptions: {
```
olavloite, line 50: I think it would be better for the sample to add the priority to the `transaction.execute(partition)` calls below (if those support an RPC priority configuration), as this call is not a call that is going to take a long time. This could also confuse users into thinking that setting a priority here would have an effect for all the partitions that are returned.

samples/rpc-priority-queryPartitions.js
```diff
@@ -0,0 +1,86 @@
...
+        priority: Priority.PRIORITY_LOW,
+      },
+    });
+    console.log(`Successfully created ${partitions.length} query partitions.`);
+
+    let row_count = 0;
+    const promises = [];
+    partitions.forEach(partition => {
+      promises.push(
+        transaction.execute(partition).then(results => {
```
olavloite, line 60: It would make more sense to set a priority for this RPC invocation, as this is something that could consume a lot of resources on the server.

samples/rpc-priority-transaction.js
```diff
@@ -0,0 +1,84 @@
...
+    projectId: projectId,
+  });
+
+  function transactionWithRpcPriority(instanceId, databaseId) {
+    // Gets a reference to a Cloud Spanner instance and database.
+    const instance = spanner.instance(instanceId);
+    const database = instance.database(databaseId);
+
+    database.runTransaction(
+      {
```
olavloite, line 45: I think we should add a comment to explain what this does. At least, I don't know what the effect of this will be. Does it mean that all RPCs in this transaction will have priority `LOW`? Or does it only mean that the `Commit` RPC will be invoked with priority `LOW`?

samples/rpc-priority-batch-dml.js
```diff
@@ -0,0 +1,81 @@
...
+    try {
+      await database.runTransactionAsync(async transaction => {
+        const [rowCounts] = await transaction.batchUpdate(dmlStatements, {
+          requestOptions: {
+            priority: Priority.PRIORITY_LOW,
+          },
+        });
+        await transaction.commit();
+        console.log(
+          `Successfully executed ${rowCounts.length} SQL statements using Batch DML with low.`
```
asthamohta, line 65: nit: with low priority*

samples/rpc-priority-batch-dml.js
```diff
@@ -0,0 +1,81 @@
...
+      throw err;
+    } finally {
+      // Close the database when finished.
+      database.close();
+    }
+  }
+  await updateUsingBatchDmlWithRpcPriority(instanceId, databaseId);
+  // [END spanner_rpc_priority_batch_dml]
+}
+main(...process.argv.slice(2)).then(() =>
```
asthamohta, line 79: Replace this with the following in all samples
```
process.on('unhandledRejection', err => {
  console.error(err.message);
  process.exitCode = 1;
});
main(...process.argv.slice(2));
```

samples/rpc-priority-partitioned-dml.js
```diff
@@ -0,0 +1,67 @@
...
+    const database = instance.database(databaseId);
+
+    try {
+      const [rowCount] = await database.runPartitionedUpdate({
+        sql: 'UPDATE Albums SET MarketingBudget = 100000 WHERE SingerId = 1',
+        requestOptions: {
+          priority: Priority.PRIORITY_LOW,
+        },
+      });
+      console.log(`Successfully updated ${rowCount} records.`);
```
asthamohta, line 54: nit: can you add using rpc priority somewhere here?

samples/rpc-priority-batch-dml.js
surbhigarg92: done

samples/rpc-priority-partitioned-dml.js
surbhigarg92: done

asthamohta (approved):

samples/rpc-priority-queryPartitions.js
surbhigarg92: Adding priority to ```transaction.execute(partition)``` is not supported. Each partition query will inherit the options set from the partitionQuery call.

samples/rpc-priority-queryPartitions.js
surbhigarg92: Adding priority to ```transaction.execute(partition)``` is not supported. Each partition query will inherit the options set from the partitionQuery call.

samples/rpc-priority-transaction.js
surbhigarg92: You are right, it does not make any sense to add priority in run transaction method. Got it confirmed from Thiago as well that passing a priority to a transaction applies to the commit call. Made the changes as required.

samples/rpc-priority-queryPartitions.js
olavloite: Hmm... I think that is only partly correct. It seems that using a `partition` directly for the same client will indeed inherit the priority from the `createQueryPartitions` call. But the purpose of the batch read API is to also distribute the read operation to multiple clients by sending the partition IDs to other clients, and those clients would need to specify the priority, as the priority is not something that is 'remembered' server side for a partition (or am I missing something here?).

But I think that you can include query options when you are calling `transaction.execute(partition)`. That is, from what I can see, is happening in the implementation of `createQueryPartitions`. It will just add the query options that were given for the `createQueryPartitions` to the resulting partitions that are being returned: https://github.com/googleapis/nodejs-spanner/blob/2839d2317ca7368d288ee9d7feb806f0ac2069c6/src/batch-transaction.ts#L173

samples/rpc-priority-queryPartitions.js
olavloite: See above. I think you're right that the `partition` will inherit the options set for the `createQueryPartitions` call, but that only applies to partitions that are used in the same client.

samples/rpc-priority-transaction.js
```diff
@@ -0,0 +1,81 @@
...
+        const [rowCount] = await transaction.runUpdate({
+          sql: 'INSERT Singers (SingerId, FirstName, LastName) VALUES (17, @firstName, @lastName)',
+          params: {
+            firstName: 'Virginia',
+            lastName: 'Watson',
+          },
+        });
+
+        console.log(
+          `Successfully inserted ${rowCount} record into the Singers table using low RPC priority.`
```
olavloite, line 59: nit: Either this comment needs to be updated, or the `update` call above should also include a `Priority.PRIORITY_LOW` option.

samples/rpc-priority-queryPartitions.js
surbhigarg92: Even if partition is assigned to a different client , the partition will still be having priority assigned with it. Right ?
Added a comment that priority can be changed for a partition.

samples/rpc-priority-transaction.js
surbhigarg92: Moved the comment to after commit call.

samples/rpc-priority-queryPartitions.js
surbhigarg92: Added the priority in execute partition call.

samples/rpc-priority-queryPartitions.js
surbhigarg92: Added the priority in transaction.execute(partition)

olavloite (approved):

PR #1688, opened 2022-07-27
test: Integration Tests for DML returning

surbhigarg92:


system-test/spanner.ts
```diff
@@ -6144,6 +6144,186 @@ describe('Spanner', () => {
       });
     });
 
+    describe('dml returning', () => {
```
asthamohta, line 6147: i feel dml returning should be a subsection under dml and not separate, you can have a describe inside a describe as well

system-test/spanner.ts
```diff
@@ -6144,6 +6144,186 @@ describe('Spanner', () => {
...
+              assertRowsAndRowCount(data);
+              return transaction!.commit();
+            })
+            .then(() => done(), done)
+            .catch(done);
+        });
+      };
+
+      it('GOOGLE_STANDARD_SQL should return rowCount and rows from run with dml returning', done => {
+        googleSqlInsertReturning.params.key = key1;
```
asthamohta, line 6270: This can be written in a single line as `googleSqlInsertReturning.params.key=googleSqlUpdateReturning.params.key=googleSqlDeleteReturning.params.key=key1`

system-test/spanner.ts
surbhigarg92: Done

system-test/spanner.ts
surbhigarg92: Removed different keys for different tests as it was working with same key.

asthamohta (approved):
LGTM, can you just lint the file once?

PR #1690, opened 2022-08-03
feat: add support for Optimistic Concurrency Control

ko3a4ok:
BEGIN_COMMIT_OVERRIDE
chore: add support for Optimistic Concurrency Control 
END_COMMIT_OVERRIDE


src/database.ts
```diff
@@ -2713,7 +2717,10 @@ class Database extends common.GrpcServiceObject {
         ? (optionsOrRunFn as RunTransactionOptions)
         : {};
 
-    const getWriteSession = this.pool_.getWriteSession.bind(this.pool_);
+    options.optimisticLock;
```
asthamohta, line 2720: Not sure what this line over here is supposed to do, did you mean to assign it to something?

ko3a4ok: I forgot to clean up this line. It's removed now. thanks!

src/session-pool.ts
```diff
@@ -669,7 +682,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     this._traces.delete(session.id);
     this._pendingPrepare++;
     session.type = types.ReadWrite;
-    this._prepareTransaction(session)
+    this._prepareTransaction(session, optimisticLock)
```
jcorwin, line 685: This doesn't seem right. After releasing the session back to the pool, if this transaction was RW, this is attempting to prepare the next RW transaction on this session. At this point, we don't know if the next transaction will be optimistic or not.

ko3a4ok: Fixed. Using one common session pool for RO, RW and OCC transactions.

test/spanner.ts
```diff
@@ -438,7 +441,13 @@ describe('Spanner with mock server', () => {
         request.requestOptions!.transactionTag,
         'transaction-tag'
       );
-      assert.ok(request.transaction!.begin!.readWrite, 'ReadWrite is not set');
+      const beginTxnRequest = spannerMock.getRequests().find(val => {
```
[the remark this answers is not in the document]
ko3a4ok: this is the existing test that checks the fallback on BeginTransactionRequest if the inline begin is failed.  
The same for OCC: it inline begin fails with occ, we should call BeginTransactionRequest with the readLockMode='OPTIMISTIC'.

surbhigarg92 (approved):

src/transaction.ts
```diff
@@ -2278,6 +2279,20 @@ export class Transaction extends Dml {
   useInRunner(): void {
     this._useInRunner = true;
   }
+
+  /**
+   * Use optimistic concurrency control for the transaction.
```
surbhigarg92, line 2284: Should we mention in comments that its a restricted feature ?

ko3a4ok: The feature is restricted on the server side. It means that when the feature is GA, users can still use the current library version to try OCC(including DML).

PR #1692, opened 2022-08-06
feat: inline BeginTransaction with first statement

ko3a4ok:
Add inline BeginTransaction for the first read or executeSql request in a transaction.

Each transaction doesn't call `beginTransanction` before its start. Instead, all creation transaction parameters are passed within the first read or executeSql request. 
In case of a transaction restart, `beginTransaction` is called explicitly.


src/transaction.ts
```diff
@@ -1034,6 +1029,8 @@ export class Snapshot extends EventEmitter {
       const transaction: spannerClient.spanner.v1.ITransactionSelector = {};
       if (this.id) {
```
olavloite, line 1030: This does not take into account that a transaction could in theory have multiple requests in flight at the same time. If a transaction starts out with sending two `SELECT` statements to the backend, it might very well be that the first that is sent has not yet returned a transaction id before the second is being sent. That will cause both requests to include a `BeginTransaction` option.

Consider the following test case (the `getRowCountFromStreamingSql` function is defined in [test/Spanner.ts](https://github.com/googleapis/nodejs-spanner/blob/d91a02fb9387408af8d868072df675e98143c022/test/spanner.ts#L3898)):

```typescript
    it('should only inline one begin transaction', async () => {
      const database = newTestDatabase();
        const rowCount1 = getRowCountFromStreamingSql(tx, {sql: selectSql});
        const rowCount2 = getRowCountFromStreamingSql(tx, {sql: selectSql});
        await Promise.all([rowCount1, rowCount2]);
        await tx.commit();
      });
      await database.close();

      let request = spannerMock.getRequests().find(val => {
        return (val as v1.ExecuteSqlRequest).sql;
      }) as v1.ExecuteSqlRequest;
      assert.ok(request, 'no ExecuteSqlRequest found');
      assert.deepStrictEqual(request.transaction!.begin!.readWrite, {});
      assert.strictEqual(request.sql, selectSql);

      request = spannerMock
        .getRequests()
        .slice()
        .reverse()
        .find(val => {
          return (val as v1.ExecuteSqlRequest).sql;
        }) as v1.ExecuteSqlRequest;
      assert.ok(request, 'no ExecuteSqlRequest found');
      assert.strictEqual(request.sql, selectSql);
      assert.ok(request.transaction!.id, 'TransactionID is not set.');
      const beginTxnRequest = spannerMock.getRequests().find(val => {
        return (val as v1.BeginTransactionRequest).options?.readWrite;
      }) as v1.BeginTransactionRequest;
      assert.ok(!beginTxnRequest, 'beginTransaction was called');
    });

```

src/transaction.ts
```diff
@@ -573,6 +563,8 @@ export class Snapshot extends EventEmitter {
 
     if (this.id) {
       transaction.id = this.id as Uint8Array;
+    } else if (typeof this._options.readWrite !== 'undefined') {
+      transaction.begin = this._options;
```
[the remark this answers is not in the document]
ko3a4ok: Added. thanks a lot for this comment: it helped to find a bug in the code

src/transaction.ts
ko3a4ok: You're absolutely right. Added a lock that prevents a multiple inline begin transaction at the same time. Thanks!

src/transaction.ts
```diff
@@ -612,10 +611,14 @@ export class Snapshot extends EventEmitter {
       });
     };
 
-    return partialResultStream(makeRequest, {
+    return partialResultStream(this._wrapWithIdWaiter(makeRequest), {
       json,
       jsonOptions,
       maxResumeRetries,
+    }).on('response', response => {
```
surbhigarg92, line 618: What would happen if the call fails?
As per the detailed design [doc](https://docs.google.com/document/d/11A3r8FG5bTmtccVhtO77xlTNKo_kX0MbdiVhGzEoWeQ/edit#heading=h.nxddek4w5dli) we should make an explicit begin transaction call if the call fails.

ko3a4ok: if the call fails, the transaction is created explicitly with "BeginTransaction" rpc call.
It's covered in the test: **"should use beginTransaction on retry"**.

src/transaction.ts
```diff
@@ -573,6 +570,8 @@ export class Snapshot extends EventEmitter {
 
     if (this.id) {
       transaction.id = this.id as Uint8Array;
+    } else if (typeof this._options.readWrite !== 'undefined') {
```
asthamohta, line 573: Should this rather be` else if(this._options.readWrite)`

src/transaction.ts
asthamohta: Is this done at the backend or do we explicitly call begin transaction from the client side?

surbhigarg92:
@ko3a4ok Can you please fix the failing test cases ?

ko3a4ok:
> @ko3a4ok We should also remove the transaction.begin call at [line](https://github.com/googleapis/nodejs-spanner/blob/588c1a2e0c449cfcb86cac73da32dd5794ee2baa/src/transaction-runner.ts#L197) and add the appropriate test cases for the same.

@surbhigarg92, unfortunately, if we just remove it, the explicit beginTransaction call won't be called on the retry. I updated to call it only after the first attempt. Please let me know if it makes sense, or is it better to move the call `transaction.begin()` somewhere [here](https://github.com/googleapis/nodejs-spanner/blob/588c1a2e0c449cfcb86cac73da32dd5794ee2baa/src/transaction-runner.ts#L234).

surbhigarg92:
> > @ko3a4ok We should also remove the transaction.begin call at [line](https://github.com/googleapis/nodejs-spanner/blob/588c1a2e0c449cfcb86cac73da32dd5794ee2baa/src/transaction-runner.ts#L197) and add the appropriate test cases for the same.
> 
> @surbhigarg92, unfortunately, if we just remove it, the explicit beginTransaction call won't be called on the retry. I updated to call it only after the first attempt. Please let me know if it makes sense, or is it better to move the call `transaction.begin()` somewhere [here](https://github.com/googleapis/nodejs-spanner/blob/588c1a2e0c449cfcb86cac73da32dd5794ee2baa/src/transaction-runner.ts#L234).

This looks fine. Only thing I am wondering is that with this implementation, we are calling `transaction.begin()` in all possible failure cases. 
@olavloite Do you think there will be any particular error case where we wan't to skip calling `transaction.begin()`

surbhigarg92:
@ko3a4ok I believe we are missing 2 use cases for Inline BeginTransaction implementation.

1. Mutation - Consider a transaction which contains only mutation operations. In that case, first rpc in the transaction will be commit transaction. To handle this scenario , we would need to make an explicit begin transaction rpc.
2. ExecuteBatchDML - Transaction needs to be inlined for ExecuteBatchDML operation as well. Special handling is required for this use case.  Please refer  [design document](https://docs.google.com/document/d/1ghXIE2r0WtepsSLH5sGr9TUu3iLUGaJEM255-NeINuY/edit#)

> ExecuteBatchDml is also slightly odd since either the call itself can fail, or it can return success with a non-OK Status in the ExecuteBatchDmlResponse. In the latter case, there may still be one or more ResultSets returned, with the first containing a transaction ID in its ResultSetMetadata. The upshot is that the client should check ExecuteBatchDmlResponse::result_set for a transaction prior to considering ExecuteBatchDmlResponse::status.

ko3a4ok:
> @ko3a4ok I believe we are missing 2 use cases for Inline BeginTransaction implementation.
> 
> 1. Mutation - Consider a transaction which contains only mutation operations. In that case, first rpc in the transaction will be commit transaction. To handle this scenario , we would need to make an explicit begin transaction rpc.
> 2. ExecuteBatchDML - Transaction needs to be inlined for ExecuteBatchDML operation as well. Special handling is required for this use case.
> 
> > ExecuteBatchDml is also slightly odd since either the call itself can fail, or it can return success with a non-OK Status in the ExecuteBatchDmlResponse. In the latter case, there may still be one or more ResultSets returned, with the first containing a transaction ID in its ResultSetMetadata. The upshot is that the client should check ExecuteBatchDmlResponse::result_set for a transaction prior to considering ExecuteBatchDmlResponse::status.

Regarding 1.: in this case, we should not create any transaction at all because `CommitRequest` will include singleUseTransaction parameter, and `transactionId` will be empty. This is how it works now, and inlined begin transaction shouldn't affect this implementation. 

Regarding 2.: I completely forgot about that, thanks for bringing this up.  Added in https://github.com/googleapis/nodejs-spanner/pull/1692/commits/2caaa0fd233314386627e946499a9822391eef84

olavloite:
> 
> Regarding 1.: in this case, we should not create any transaction at all because `CommitRequest` will include singleUseTransaction parameter, and `transactionId` will be empty. This is how it works now, and inlined begin transaction shouldn't affect this implementation.
> 

No, I don't think it's correct that the current implementation uses a single-use read/write transaction if you use `runTransaction(..)` to execute a transaction with only mutations. Currently, `runTransaction(..)` will get a write-prepared session from the pool and use that transaction to commit the mutations. You are right that `CommitRequest` will just use a single-use read/write transaction if there is no transaction ID, but that is not something that will happen when you use `runTransaction(..)`.
In the new implementation, `runTransaction(..)` will not have a transaction id when it is committed if you have not executed any  SQL statements, and will now use a single-use read/write transaction.

So there is a change in behavior specifically when using `runTransaction`, and that behavior change is relatively important, as it changes the semantics of `runTransaction(..)` from `execute once` to `execute at least once`.

ko3a4ok:
> > Regarding 1.: in this case, we should not create any transaction at all because `CommitRequest` will include singleUseTransaction parameter, and `transactionId` will be empty. This is how it works now, and inlined begin transaction shouldn't affect this implementation.
> 
> No, I don't think it's correct that the current implementation uses a single-use read/write transaction if you use `runTransaction(..)` to execute a transaction with only mutations. Currently, `runTransaction(..)` will get a write-prepared session from the pool and use that transaction to commit the mutations. You are right that `CommitRequest` will just use a single-use read/write transaction if there is no transaction ID, but that is not something that will happen when you use `runTransaction(..)`. In the new implementation, `runTransaction(..)` will not have a transaction id when it is committed if you have not executed any SQL statements, and will now use a single-use read/write transaction.
> 
> So there is a change in behavior specifically when using `runTransaction`, and that behavior change is relatively important, as it changes the semantics of `runTransaction(..)` from _execute once_ to _execute **at least** once_.

Thanks @olavloite, for explaining that. Fixed.

src/transaction.ts
```diff
@@ -612,11 +611,19 @@ export class Snapshot extends EventEmitter {
...
       maxResumeRetries,
-    });
+    })
+      ?.on('response', response => {
+        if (response.metadata && response.metadata!.transaction && !this.id) {
+          this._update(response.metadata!.transaction);
+        }
+      })
+      .on('error', () => {
+        if (!this.id && this._useInRunner) this.begin();
```
olavloite, line 625: nit: (informal) styleguide requires the use of curly braces for all if statements.

olavloite (approved):

surbhigarg92 (approved):

src/transaction.ts
ko3a4ok: sure. fixed

PR #1693, opened 2022-08-08
Fgac changes

asthamohta:
The PR includes generated changes as well, please just review the files in the following folders
1. Samples
2. src/database,session,session-pool
3. system-test
4. test


samples/enable-fine-grained-access.js
```diff
@@ -0,0 +1,78 @@
...
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  async function enableFineGrainedAccess() {
+    // Gets a reference to a Cloud Spanner instance and database.
+    const instance = spanner.instance(instanceId);
+    const database = instance.database(databaseId);
+
+    const [policy] = await database.getIamPolicy({requestedPolicyVersion: 3});
```
olavloite, line 52: What's this magic version number '3'? I think we need a small comment to explain this to users.

samples/enable-fine-grained-access.js
```diff
@@ -0,0 +1,78 @@
...
+    if (policy.version < 3) {
+      policy.version = 3;
+    }
+
+    const newBinding = {
+      role: 'roles/spanner.fineGrainedAccessUser',
```
olavloite, line 58: I think this could also use a small comment. Why are we specifically using this role? Is it automatically available for all databases?

samples/read-data-with-database-role.js
```diff
@@ -0,0 +1,78 @@
...
+
+  // Instantiates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  async function readDataWithDatabaseRole() {
+    // Gets a reference to a Cloud Spanner instance and database.
+    const instance = spanner.instance(instanceId);
+    // Creating new Database with creator role set and using that
```
olavloite, line 44: I don't understand what this comment means. What happens when I set a 'creator role' on a database?

I also think that we should not call it 'Creating new Database' here, as we are not creating a new database, we are only creating a __reference__ to a database. So maybe something like:

`Connect to a database using the 'parent' role. This means that the connection will only have the permissions that have explicitly been granted to the 'parent' role.`

src/database.ts
```diff
@@ -187,6 +207,12 @@ export type CreateSessionResponse = [
 
 export interface CreateSessionOptions {
   labels?: {[k: string]: string} | null;
+  databaseRole?: string | null;
+  gaxOptions?: CallOptions;
+}
+
+export interface GetIamPolicyOptions {
+  requestedPolicyVersion?: number | null;
```
olavloite, line 215: This should probably include a comment so customers know what this version does and which one they should be using.

src/session-pool.ts
```diff
@@ -497,6 +499,10 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     this.database = database;
     this.options = Object.assign({}, DEFAULTS, options);
     this.options.min = Math.min(this.options.min!, this.options.max!);
+    this.options.databaseRole =
+      this.options.databaseRole !== null
```
olavloite, line 503: Should this not be just `this.options.databaseRole` instead of `this.options.databaseRole !== null` so that we don't let for example `undefined` override what is set in `database.databaseRole`.

src/v1/database_admin_client.ts
```diff
@@ -3410,6 +3419,210 @@ export class DatabaseAdminClient {
...
+    options.otherArgs.headers['x-goog-request-params'] =
+      gax.routingHeader.fromParams({
+        parent: request.parent || '',
+      });
+    this.initialize();
+    return this.innerApiCalls.listDatabaseRoles(request, options, callback);
+  }
+
+  /**
+   * Equivalent to `method.name.toCamelCase()`, but returns a NodeJS Stream object.
```
olavloite, line 3527: ?

src/v1/database_admin_client_config.json
```diff
@@ -118,6 +118,10 @@
           "timeout_millis": 3600000,
           "retry_codes_name": "idempotent",
           "retry_params_name": "e9fafda332ce8a1702dc1575de3ca81c4feb4799"
+        },
+        "ListDatabaseRoles": {
+          "retry_codes_name": "non_idempotent",
```
olavloite, line 123: I know that this is auto-generated, so not something that we can change here, but maybe you can put the question to the team that created the proto changes: Why is this `non_idempotent`? I would expect this method to be idempotent.

system-test/spanner.ts
```diff
@@ -1588,6 +1588,295 @@ describe('Spanner', () => {
...
+              done();
+            });
+          })
+        );
+      };
+
+      it('GOOGLE_STANDARD_SQL should create a user defined role', done => {
+        createUserDefinedDatabaseRole(done, DATABASE, 'CREATE ROLE parent');
+      });
+      //
```
olavloite, line 1610: remove commented code? Or enable it?

system-test/spanner.ts
```diff
@@ -5204,6 +5586,8 @@ describe('Spanner', () => {
         }
       );
     });
+
+    it('');
```
olavloite, line 5590: remove?

src/v1/database_admin_client.ts
asthamohta: This is autogenerated.

src/v1/database_admin_client_config.json
asthamohta: This is autogenerated. But I can check and find out

src/database.ts
```diff
@@ -270,6 +301,8 @@ class Database extends common.GrpcServiceObject {
   queryOptions_?: spannerClient.spanner.v1.ExecuteSqlRequest.IQueryOptions;
   resourceHeader_: {[k: string]: string};
   request: DatabaseRequest;
+  ale?: string | null;
```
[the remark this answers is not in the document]
asthamohta: Removed

PR #1699, opened 2022-08-26
Jsonb

asthamohta:
Can you please review the handwritten changes in the following files:
1. samples/pg-jsonb-add-column.js
2. samples/pg-jsonb-query-parameter.js
3. samples/pg-jsonb-update-data.js
4. src/codec.ts
5. src/index.ts
6. system-test/spanner.ts
7. test/codec.ts
8. test/index.ts


src/codec.ts
```diff
@@ -457,6 +483,10 @@ function encodeValue(value: Value): Value {
     return value.map(encodeValue);
   }
 
+  if (value instanceof PGJsonb) {
+    return JSON.stringify(value.value);
```
surbhigarg92, line 487: Can we use toString() method instead from PGJsonb class?

asthamohta: Done

PR #1739, opened 2022-11-08
fix: cleanup different types of session pools

ko3a4ok:


src/session-pool.ts
```diff
@@ -143,7 +143,6 @@ export interface SessionPoolOptions {
   max?: number;
   maxIdle?: number;
   min?: number;
-  writes?: number;
```
olavloite, line 146: Removing a property from an (exported) interface is a breaking change. We either need to do a major version bump and set up an LTS branch for the current version, or just deprecate the property without removing it.

src/session-pool.ts
```diff
@@ -302,21 +300,10 @@ const enum errors {
...
+  sessions: Session[];
   borrowed: Set<Session>;
 }
 
-interface Waiters {
-  [types.ReadOnly]: number;
-  [types.ReadWrite]: number;
-}
-
-export interface CreateSessionsOptions {
```
olavloite, line 315: Same here: Removing this interface all together is a breaking change.

src/session-pool.ts
```diff
@@ -405,19 +372,6 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     return this.size >= this.options.max!;
   }
 
-  /**
-   * Total number of read sessions.
-   * @type {number}
-   */
-  get reads(): number {
```
olavloite, line 412: Although these functions are not part of any interfaces, they could be used by users to gather statistics for their session pool usage, which means that also this would be a breaking change.

src/session-pool.ts
```diff
@@ -644,21 +549,10 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
-    //    needs read/write sessions, and the pool should try to keep that number
-    //    of read/write sessions dynamically.
-    // 3. The user has set a write fraction and that fraction has been reached.
-    const shouldBeWrite =
-      (session.type === types.ReadWrite && this.options.writes === 0) ||
-      (this.options.writes! > 0 &&
-        this.currentWriteFraction < this.options.writes!);
-    if (this.totalWaiters > this.available || !shouldBeWrite) {
-      session.type = types.ReadOnly;
+    // Release it into the pool as a session in if there are more waiters than
```
olavloite, line 552: nit: 
```suggestion
    // Release it into the pool as a session if there are more waiters than
```

src/session-pool.ts
```diff
@@ -644,21 +549,10 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+    // there are sessions available. Releasing it will be unblocked as soon as
```
olavloite, line 553:
```suggestion
    // there are sessions available. Releasing it will unblock a waiter as soon as
```

src/session-pool.ts
```diff
@@ -617,14 +525,11 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
 
   /**
-   * Releases session back into the pool. If the session is a write session it
-   * will also prepare a new transaction before releasing it.
+   * Releases session back into the pool.
    *
    * @throws {Error} For unknown sessions.
    * @emits SessionPool#available
    * @emits SessionPool#error
-   * @emits SessionPool#readonly-available
```
olavloite, line 626: Also a breaking change

src/session-pool.ts
```diff
@@ -667,25 +561,18 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
     this._traces.delete(session.id);
-    this._pendingPrepare++;
-    session.type = types.ReadWrite;
-    this._prepareTransaction(session)
-      .catch(() => (session.type = types.ReadOnly))
-      .then(() => {
-        this._pendingPrepare--;
-        this._release(session);
-      });
+    this._prepareTransaction(session);
```
olavloite, line 564: Is this still necessary? I assume that it does not actually call the `BeginTransaction` RPC, but why do we need it for all sessions that are released into the pool here? And why do we not do it if the session is released into the pool above to unblock a waiter (on line 556)?

src/session-pool.ts
```diff
@@ -747,24 +615,21 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
 
     this._inventory.borrowed.add(session);
-    this._inventory[type].splice(index, 1);
+    this._inventory.sessions.splice(index, 1);
   }
 
   /**
-   * Borrows the first session from specific group. This method may only be called if the inventory
-   * actually contains a session of the desired type.
+   * Borrows the first session from specific group.
```
olavloite, line 625: nit: 
```suggestion
   * Borrows the first session from the inventory.
```

src/session-pool.ts
```diff
@@ -774,82 +639,59 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
-      return this._borrowFrom(types.ReadWrite);
-    }
-
-    return this._borrowFrom(types.ReadOnly);
+  _borrowNextAvailableSession(): Session {
+    return this._borrowFrom();
   }
 
   /**
    * Attempts to create a single session of a certain type.
```
olavloite, line 649:
```suggestion
   * Attempts to create a single session.
```

src/session-pool.ts
```diff
@@ -774,82 +639,59 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
-    const kind = type === types.ReadOnly ? 'reads' : 'writes';
-    const options = {[kind]: 1};
-
-    return this._createSessions(options);
+  _createSession(): Promise<void> {
+    return this._createSessions(1);
   }
 
   /**
    * Batch creates sessions and prepares any necessary transactions.
```
olavloite, line 660:
```suggestion
   * Batch creates sessions.
```

src/session.ts
```diff
@@ -42,14 +42,6 @@ import IRequestOptions = google.spanner.v1.IRequestOptions;
 
 export type GetSessionResponse = [Session, r.Response];
 
-/**
- * enum to capture the possible session types
- */
-export const enum types {
```
olavloite, line 48: Also a breaking change

olavloite (commented):
I think the biggest question here is how we want to handle this change, as it would currently be a breaking change. Do we:
1. Deprecate fields / interfaces / events that we are no longer using instead of removing them?
2. Mark this as a breaking change, do a major version bump and set up an LTS branch for the current version?

src/session-pool.ts
ko3a4ok: Good point. Let's keep it for now, and it can be removed next time for the major version bump.
I marked this field as deprecated.

src/session-pool.ts
ko3a4ok: interesting, I didn't consider this. Thanks for letting me know, reverted back.

src/session-pool.ts
ko3a4ok: Done.

src/session-pool.ts
ko3a4ok: No, it's not necessary any more. Transactions will be prepared(initialized) anyway during the session acquiring anyway.

src/session-pool.ts
ko3a4ok: Done.

src/session.ts
ko3a4ok: Reverted.

src/session-pool.ts
ko3a4ok: Reverted back.

src/session-pool.ts
ko3a4ok: Done.

src/session-pool.ts
ko3a4ok: Done.

src/session-pool.ts
ko3a4ok: Done.

src/session-pool.ts
ko3a4ok: Reverted back.

src/session-pool.ts
asthamohta: Can you update the comment above as well as it is used as part of automatically generated documentation

src/session-pool.ts
```diff
@@ -952,56 +806,34 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
       throw new SessionPoolExhaustedError(this._getLeaks());
     }
 
     let removeOnceCloseListener: Function;
     let removeListener: Function;
 
-    // Wait for the requested type of session to become available.
-    const availableEvent = type + '-available';
+    // Wait for a session to become available.
+    const availableEvent = 'available';
```
asthamohta, line 836: We need to change the name of what is emitted here as 'available' is already an event caught in database: https://github.com/googleapis/nodejs-spanner/blob/1fadc69540015b696d678612e8aa370b13daa878/src/database.ts#L352
We can change it to 'session-available' here and also emit it at line number 1012: https://github.com/googleapis/nodejs-spanner/pull/1739/files#diff-d6f67152c7ddf6462713d84c5314cab643222c347a9b1c87f59498d8cd9c84abR1012
rather than 'readonly-available' and 'readwrite-available' as those are currently emitted an not caught anywhere

src/session-pool.ts
```diff
@@ -715,26 +618,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     };
 
     const session = await this._acquires.add(getSession);
-
-    if (type === types.ReadWrite && session.type === types.ReadOnly) {
-      this._numInProcessPrepare++;
-      try {
-        await this._prepareTransaction(session);
-      } catch (e) {
-        if (isSessionNotFoundError(e as ServiceError)) {
```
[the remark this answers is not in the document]
ko3a4ok: `_prepareTransaction` isn't a promise method anymore: it's responsible to prepare a new instance of the transaction object; no RPC calls happen at the time, so there is no way to catch any `ServiceError` now.

src/session-pool.ts
```diff
@@ -568,10 +499,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @param {GetReadSessionCallback} callback The callback function.
    */
   getReadSession(callback: GetReadSessionCallback): void {
```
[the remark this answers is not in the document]
ko3a4ok: Yes, good point, thanks!

src/session-pool.ts
ko3a4ok: I think @olavloite made a comment https://github.com/googleapis/nodejs-spanner/pull/1739#discussion_r1019983391 that even if we're not catching this `readonly-available` and `readwrite-available` in our codebase, it can be a breakable changes and it can affect our customers that rely on these events. 

we have some options:
1) remove `readonly-available` and `readwrite-available`
2) replace `readonly-available` and `readwrite-available` with `session-available`
3) add `session-available` (also mark `readonly-available` and `readwrite-available` as deprecated). 

Please confirm which option is the right to move forward, and I'm happy to update to code.

surbhigarg92: @ko3a4ok I had a discussion with @olavloite and we should go ahead with option **3**. We do not want to remove `readonly-available` and `readwrite-available` as it might cause a breaking change for the customer.

ko3a4ok: thanks for checking. Done

src/session-pool.ts
```diff
@@ -366,27 +377,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
-    const writes = this._inventory[types.ReadWrite];
-
-    return reads.length + writes.length;
-  }
-
-  /**
-   * Current fraction of write-prepared sessions in the pool.
-   * @type {number}
-   */
-  get currentWriteFraction(): number {
```
surbhigarg92, line 379: This is a public method , removing this will also be a breaking change. We can mark this method as deprecated and probably return 0 ?

src/session-pool.ts
```diff
@@ -458,28 +433,12 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     return this._pending + this._pendingPrepare;
   }
 
-  /**
-   * Current number of waiters for a read-only session.
-   * @type {number}
-   */
-  get numReadWaiters(): number {
```
surbhigarg92, line 465: Same case as `currentWriteFraction` , `numReadWaiters` and `numWriteWaiters` are public methods and removing these will also be a breaking change. 
We can mark these methods as deprecated and probably return 0 ?

src/session-pool.ts
surbhigarg92: @ko3a4ok We will also have to modify the event name like below 
```const availableEvent = 'session-available';```

File: src/session-pool.ts [hunk 17]
```diff
@@ -644,48 +599,24 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
-      session.type = types.ReadOnly;
-      this._release(session);
-      return;
-    }
-
     // Delete the trace associated with this session to mark the session as checked
     // back into the pool. This will prevent the session to be marked as leaked if
     // the pool is closed while the session is being prepared.
     this._traces.delete(session.id);
-    this._pendingPrepare++;
```
surbhigarg92, line 670: @ko3a4ok This was the only place where `_pendingPrepare` flag was getting updated , now since we have removed this reference , we can remove the property `_pendingPrepare` . Also we should mark `pendingPrepare()` method as deprecated.

src/session-pool.ts
ko3a4ok: Done.

src/session-pool.ts
ko3a4ok: Thanks. done.

src/session-pool.ts
ko3a4ok: done.

src/session-pool.ts
ko3a4ok: ah.. yes, thanks! done

src/session-pool.ts
ko3a4ok: done

surbhigarg92 (approved):

src/session-pool.ts
ko3a4ok: Yes, I updated it previously in https://github.com/googleapis/nodejs-spanner/pull/1739/files#diff-d6f67152c7ddf6462713d84c5314cab643222c347a9b1c87f59498d8cd9c84abR145

src/session-pool.ts
```diff
@@ -80,15 +85,23 @@ export interface SessionPoolInterface extends EventEmitter {
...
    * @name SessionPoolInterface#open
    */
+  /**
+   * When called returns a session.
+   *
+   * @name SessionPoolInterface#getSession
+   * @param {GetSessionCallback} callback The callback function.
+   */
   /**
    * When called returns a read-only session.
```
olavloite, line 95: nit: I think we should update this part of the documentation as well to indicate that it always just returns a session. The same also applies to the GetWriteSessionCallback below.

src/session-pool.ts
```diff
@@ -80,15 +85,23 @@ export interface SessionPoolInterface extends EventEmitter {
...
    *
+   * @deprecated Use getSession instead.
    * @name SessionPoolInterface#getReadSession
    * @param {GetReadSessionCallback} callback The callback function.
    */
   /**
    * When called returns a read-write session with prepared transaction.
```
olavloite, line 102: nit: also update this

olavloite (approved):

src/session-pool.ts
ko3a4ok: done

src/session-pool.ts
ko3a4ok: done

PR #1783, opened 2023-01-31
feat: Leader aware routing

surbhigarg92:
Add x-goog-spanner-route-to-leader header to Spanner RPC contexts for RW/PDML transactions. The header is added to support leader-aware-routing feature, which aims at reducing cross-regional latency for RW/PDML transactions in a multi-region instance.

This feature is off by default and can be enabled by setting `routeToLeaderEnabled` as true while creating Spanner object. 

```
const spanner = new Spanner({
    projectId: projectId,
    routeToLeaderEnabled: true,
  });
```


src/index.ts
```diff
@@ -310,6 +317,14 @@ class Spanner extends GrpcService {
       packageJson: require('../../package.json'),
     } as {} as GrpcServiceConfig;
     super(config, options);
+
+    if (
```
asthamohta, line 321: Why not replace with `routeToLeaderEnabled===false`

src/session.ts
```diff
@@ -378,13 +383,24 @@ export class Session extends common.GrpcServiceObject {
     const reqOpts = {
       name: this.formattedName_,
     };
+
```
asthamohta, line 386: This code looks repetitive, can't we add a function for this such as `header=get_header(this.parent.parent.parent as Spanner, this.resourceHeader_)`

src/index.ts
surbhigarg92: `routeToLeaderEnabled` will be assigned `false` based on SpannerOptions which we are doing in this condition.

src/session.ts
surbhigarg92: Done.

src/index.ts
asthamohta: My bad I meant why not check `options.routeToLeaderEnabled===false`
From what I understand in case options.routeToLeaderEnabled is undefined then it won't be equal to false. And the other condition checks for if it is true

asthamohta (approved):

src/database.ts
```diff
@@ -523,13 +525,18 @@ class Database extends common.GrpcServiceObject {
       sessionCount: count,
     };
 
+    const headers = this.resourceHeader_;
+    if ((this.instance.parent as Spanner).routeToLeaderEnabled) {
```
olavloite, line 529: Could we maybe refactor these blocks a little bit so that the if-statement also goes in a generic method. So that would probably mean:

1. Add a method like `getSpanner` to all classes that need to be able to add these headers.
2. Modify the `addLeaderAwareRoutingHeader` so it takes a `Spanner` instance as an argument and includes this if-statement

src/transaction.ts
```diff
@@ -602,6 +616,15 @@ export class Snapshot extends EventEmitter {
       }
     );
 
+    const headers = this.resourceHeader_;
+    if (
+      (this.session.parent.parent.parent as Spanner).routeToLeaderEnabled &&
+      (this._options.readWrite !== undefined ||
```
olavloite, line 622: Ah... I guess that this is the reason that the if statement is not part of the `addLeaderAwareRoutingHeader` function.

test/spanner.ts
```diff
@@ -1486,6 +1489,69 @@ describe('Spanner with mock server', () => {
...
+          await tx!.runUpdate({
+            sql: insertSql,
+          });
+          return await tx.commit();
+        });
+        await database.close();
+        spannerMock.getMetadata().forEach(metadata => {
+          if (metadata.get(LEADER_AWARE_ROUTING_HEADER)[0] !== undefined) {
+            assert.strictEqual(
+              metadata.get(LEADER_AWARE_ROUTING_HEADER)[0],
```
olavloite, line 1531: Could we also add an assertion that we have seen at least one metadata instance with leader-aware-routing? (The current assertion also succeeds if `metadata.get(LEADER_AWARE_ROUTING_HEADER)[0] === undefined` for all requests)

test/spanner.ts
```diff
@@ -1486,6 +1489,69 @@ describe('Spanner with mock server', () => {
...
+
+      before(() => {
+        spannerWithLARDisabled = new Spanner({
+          servicePath: 'localhost',
+          port,
+          sslCreds: grpc.credentials.createInsecure(),
+          routeToLeaderEnabled: false,
+        });
+        // Gets a reference to a Cloud Spanner instance and database
+        instanceWithEnvVar = spannerWithLARDisabled.instance('instance');
```
olavloite, line 1516: nit:
```suggestion
        instanceWithLARDisabled = spannerWithLARDisabled.instance('instance');
```

src/database.ts
surbhigarg92: Added the method getSpanner in each class.

src/transaction.ts
surbhigarg92: Yes, I have not moved if condition to addLeaderAwareRoutingHeader

src/batch-transaction.ts
```diff
@@ -20,12 +20,11 @@ import * as extend from 'extend';
 import * as is from 'is';
 import {Snapshot} from './transaction';
 import {google} from '../protos/protos';
-import {Session, Database} from '.';
+import {Session, Database, Spanner} from '.';
```
olavloite, line 23: nit: the Spanner import is no longer needed, now that we have moved the logic into a separate function (the `_getSpanner()` function)

olavloite (approved):

PR #1801, opened 2023-02-09
test: emit empty metadata before emitting unspecified error

alexander-fenster:
Since gRPC upgrade to `@grpc/grpc-js` v1.8.x, the server can emit errors without sending a metadata. It breaks some retrying logic in `retry-request`. This is unlikely to happen in production; for the test, let's make our stub gRPC server emit `metadata` event (with an empty `Metadata` instance) before emitting an error.

BEGIN_COMMIT_OVERRIDE
fix: tests emit empty metadata before emitting unspecified error
END_COMMIT_OVERRIDE

### DIFF
--- a/test/mockserver/mockspanner.ts
+++ b/test/mockserver/mockspanner.ts
@@ -616,6 +616,7 @@ export class MockSpanner {
               call.write(MockSpanner.toPartialResultSet(res.updateCount));
               break;
             case StatementResultType.ERROR:
+              call.sendMetadata(new Metadata());
               call.emit('error', res.error);
               break;
             default:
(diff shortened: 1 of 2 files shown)


test/mockserver/mockspanner.ts
```diff
@@ -633,6 +634,7 @@ export class MockSpanner {
         call.end();
       })
       .catch(err => {
+        call.sendMetadata(new Metadata());
```
surbhigarg92, line 637: There are lots of other emit `error` event in the mock spanner file. Do we need to do this at all places ?

alexander-fenster: I don't think so. My goal here is to make tests pass to allow proto updates to go through; if the Spanner teams wants to refactor the mock gRPC server used in unit tests, you can do so separately :)

surbhigarg92 (approved):

Issue #1517, opened 2021-11-24
Generated samples are missing


fhinkel:
For example there's no sample with this referenced region tag: `spanner_v1_generated_Spanner_BatchCreateSessions_async`

This got merged but it looks somewhat wrong to me: https://github.com/googleapis/nodejs-spanner/pull/1484/files

surbhigarg92:
@sofisl Do we have any update on the autogenerated samples for Node js ?

sofisl:
@surbhigarg92, is this something the spanner team wants? https://github.com/googleapis/nodejs-spanner/blob/f1ae51301d4ea9b0ed1ad4d4762c249fef9f8d08/owlbot.py#L66

surbhigarg92:
@sofisl This was done intentionally to avoid autogenerated samples as we are not using autogenerated client , instead have an handwritten layer for same.

sofisl:
Hi @surbhigarg92, I see what you're saying, but I think there's also an issue with how it renders in cloud.google.com. See here, under [example](https://cloud.google.com/nodejs/docs/reference/spanner/latest/spanner/v1.spannerclient#_google_cloud_spanner_v1_SpannerClient_batchCreateSessions_member_1_)

![Screenshot 2023-02-14 at 12 33 09 PM](https://user-images.githubusercontent.com/55454395/218855714-82099fb9-ac6b-464b-a8b7-1429cd634a01.png)


So, for client libraries with a veneer, we do include these comments:

https://github.com/googleapis/gapic-generator-typescript/blob/fbda4e4065224ed99419f68116538f1d49177ef5/templates/typescript_samples/samples/generated/%24version/%24service.%24method.js.njk#L4

Would it be alright if we included samples here, with that notification?

Issue #1813, opened 2023-03-01
Docs have not been publishing for some time


sofisl:
See: https://fusion2.corp.google.com/ci;prev=s/kokoro/prod:cloud-devrel%2Fclient-libraries%2Fnodejs%2Frelease%2Fgoogleapis%2Fnodejs-spanner%2Fdocs-devsite/activity/7dc59bb7-da9b-49db-a3d4-cf588190dc5a/log?q=nodejs-spanner&s=p

I noticed this when investigating:

https://github.com/googleapis/nodejs-spanner/issues/1517

That problem is due to outdated samples tags, it will go away once the newest src code is uploaded.

sofisl:
cc @dansaadati

dansaadati:
Thanks for bringing this up! I opened #1815 to clear up the unused references to generated samples.

dansaadati:
It is failing due to `ERROR: Internal Error: Unexpected AST case`. I was able to replicate this error by copying a copy on local and running `api-extractor`. It looks like the easiest fix is to update `api-extractor` which is able to parse the AST successfully. We should merge https://github.com/googleapis/nodejs-cloud-rad/pull/50  and release a new version, and then regenerate docs for `nodejs-spanner`.

surbhigarg92:
@dansaadati I can see the latest build is passing.

PR #1815, opened 2023-03-01
docs: remove unused references to generated samples

dansaadati:
The Spanner team is not using auto-generated samples, so references to them should be removed.

#1517


surbhigarg92 (approved):

sofisl (approved):

sofisl:
This will partially fix #1517, but we still need to fix publishing as well (#1813)

PR #1829, opened 2023-03-28
feat: Support For Proto Column in Spanner

gauravpurohit06:
PR Includes

- Implementation for Proto Column & Enum which includes 
- - Serialisation & Deserialisation Logic 
- - Support to pass proto descriptor while creating database
- Samples
- - DDL 
- - DQL & DML 
- - Mutation & Read 
- Unit Tests


vaibhavagarwal220 (approved):

src/codec.ts
```diff
@@ -30,6 +30,23 @@ export interface Field {
   value: Value;
 }
 
+export interface IProtoMessageParams {
```
asthamohta, line 33: Curious why we are calling this IProtoMessageParams and not just ProtoMessageParams?

src/codec.ts
```diff
@@ -30,6 +30,23 @@ export interface Field {
...
+  value: object;
+  // Function representing the message with helper methods for deserializing
+  // and serializing messages, use from generated file.
+  messageFunction?: Function;
+  // fully qualified name of the proto representing the enum definition
+  fullName: string;
```
asthamohta, line 39: Just a good practice add attributes that must be passed (fullName) before those which are optional(messageFunction)

src/codec.ts
```diff
@@ -231,6 +248,78 @@ export class PGNumeric {
...
+ */
+export class ProtoEnum {
+  value: string;
+  fullName: string;
+  enumObject?: object;
+
+  constructor(protoEnumParams: IProtoEnumParams) {
+    this.fullName = protoEnumParams.fullName;
+
+    if (is.number(protoEnumParams.value)) {
```
asthamohta, line 299: If type is already defined as string for value, why do we need to convert it to string again?

asthamohta: I think we need to convert the type for value to number|string

src/codec.ts
```diff
@@ -231,6 +248,78 @@ export class PGNumeric {
...
+      this.value = protoEnumParams.value.toString();
+    } else if (
+      typeof protoEnumParams.value === 'string' &&
+      /^\d+$/.test(protoEnumParams.value)
```
asthamohta, line 303: is.number(value) also takes a string as a number and returns true. Example: is.number("10") will return true. Hence the code will never come here

src/codec.ts
```diff
@@ -231,6 +248,78 @@ export class PGNumeric {
...
+    ) {
+      this.value = protoEnumParams.value;
+    } else if (protoEnumParams.enumObject) {
+      this.value = protoEnumParams.enumObject[protoEnumParams.value];
```
asthamohta, line 307: Can customers also pass a String example this.value ="ROCK"?

src/codec.ts
```diff
@@ -30,6 +30,23 @@ export interface Field {
   value: Value;
 }
 
+export interface IProtoMessageParams {
+  value: object;
+  // Function representing the message with helper methods for deserializing
+  // and serializing messages, use from generated file.
+  messageFunction?: Function;
```
asthamohta, line 37: Add a description saying that encode decode must be present

src/codec.ts
```diff
@@ -358,6 +456,23 @@ function decode(value: Value, type: spannerClient.spanner.v1.Type): Value {
     case 'BYTES':
       decoded = Buffer.from(decoded, 'base64');
       break;
+    case spannerClient.spanner.v1.TypeCode.PROTO:
+    case 'PROTO':
+      decoded = Buffer.from(decoded, 'base64');
+      decoded = new ProtoMessage({
+        value: decoded,
+        fullName: type.protoTypeFqn,
+        messageFunction: columnInfo as Function,
```
asthamohta, line 465: Add a check here and in enum

samples/database-update-with-proto-descriptor.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    console.log(`Waiting for update on ${database.id} to complete...`);
+    await operation.promise();
+    console.log(`Updated database ${databaseId} with proto column.`);
+  }
+  updateDatabaseWithProtoDescriptor();
+  // [END spanner_update_database_with_proto_descriptors]
+}
+process.on('unhandledRejection', err => {
```
asthamohta, line 84: nit: add line here and in other files

src/index.ts
```diff
@@ -1676,6 +1680,49 @@ class Spanner extends GrpcService {
   static pgJsonb(value): PGJsonb {
     return new codec.PGJsonb(value);
   }
+
+  /**
+   * Helper function to get a Cloud Spanner proMessage object.
```
surbhigarg92, line 1685: nit: `Helper function to get a Cloud Spanner proto Message object.`

src/index.ts
```diff
@@ -1676,6 +1680,49 @@ class Spanner extends GrpcService {
...
+   *
+   * @param {IProtoEnumParams} value The proto enum value params in the format of
+   *     @code{IProtoEnumParams}
+   * @returns {ProtoEnum}
+   *
+   * @example
+   * ```
+   * const {Spanner} = require('@google-cloud/spanner');
+   * const protoEnum = Spanner.protoEnum({
+   *   value: '2uhu',
```
surbhigarg92, line 1716: Normally we expect this value to be a number or number present in string form, right ? 

So try to use a valid value ?

src/partial-result-stream.ts
```diff
@@ -55,6 +55,7 @@ export interface RowOptions {
   json?: boolean;
   jsonOptions?: JSONOptions;
   maxResumeRetries?: number;
+  columnInfo?: object;
```
surbhigarg92, line 58: Add @property for columnInfo in documentation.

src/codec.ts
gauravpurohit06: As we discussed on call, `is.number("10")` evaluates to false, so we need this check.

src/codec.ts
gauravpurohit06: Yes, but in that scenario they have to pass `protoEnumParams.enumObject` using which we will find the integer value associated with the string.

src/codec.ts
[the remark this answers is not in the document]
gauravpurohit06: Yes that would work, but I think calling `toString()` on a string willl create and return another string, which is additional overhead. Please let me know if it's not the case.

src/index.ts
gauravpurohit06: We can use number or Enum String.  Updating it to use valid enum string.

src/codec.ts
asthamohta: Sorry my bad, the line isn't descriptive. In protos, column info is used as Function and inEnums it is an object. Can we check if this is possible(in case user passes an incorrect value) and gracefully throw an error

src/codec.ts
gauravpurohit06: It's a convention to declare interfaces but it looks we are not following it much on the handwritten side.  The gapic auto-generated protos also follow similar convention: https://github.com/googleapis/nodejs-spanner/blob/main/protos/protos.d.ts#L224

src/codec.ts
gauravpurohit06: We don't need to throw an error, so if the `columnInfo ` is missing then still data would be deserialised as bytes and similar to integer in case of enum.

asthamohta: What I meant was what if it is present but in the wrong format

src/index.ts
surbhigarg92: nit: Change `proMessage` to `proto Message`

samples/system-test/spanner.test.js
```diff
@@ -1530,6 +1535,118 @@ describe('Spanner', () => {
...
+      new RegExp(`Waiting for update on ${DATABASE_ID} to complete...`)
+    );
+    assert.match(
+      output,
+      new RegExp(`Updated database ${DATABASE_ID} with proto column.`)
+    );
+  });
+
+  describe('proto columns', () => {
+    const protoDatabaseId = `test-proto-database-${CURRENT_TIME}`;
```
surbhigarg92, line 1569: Why do we need to create new proto database , we already have `PROTO_DATABASE_ID1` and `PROTO_DATABASE_ID2` , can't we reuse these ?

surbhigarg92 (approved):

samples/system-test/spanner.test.js
gauravpurohit06: @surbhigarg92, Yes we can use it but then we need to assume the tests get executed in the sequential order of their definition. I think we can assume it because other tests are also assuming the same.

src/codec.ts
gauravpurohit06: In that case it will throw an error because deserialisation would fail.

asthamohta (approved):

samples/database-create-with-proto-descriptor.js
```diff
@@ -0,0 +1,91 @@
...
+  const database = instance.database(databaseId);
+
+  async function createDatabaseWithProtoDescriptor() {
+    // Reading proto descriptor file
+    const protoDescriptor = fs
+      .readFileSync('./resource/descriptors.pb')
+      .toString('base64');
+
+    // Create a new database with an extra statements which will create
+    // proto bundle and tables with proto column
```
olavloite, line 55: nit:
```suggestion
    // Create a new database with a proto bundle and a table with proto columns
```

samples/database-create-with-proto-descriptor.js
```diff
@@ -0,0 +1,91 @@
...
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+  // Gets a reference to a Cloud Spanner instance and a database. The database does not need to exist.
+  const instance = spanner.instance(instanceId);
+  const database = instance.database(databaseId);
+
+  async function createDatabaseWithProtoDescriptor() {
+    // Reading proto descriptor file
```
olavloite, line 49:
```suggestion
    // Read a proto descriptor file and convert it to a base64 string
```

samples/database-update-with-proto-descriptor.js
```diff
@@ -0,0 +1,89 @@
...
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+  // Gets a reference to a Cloud Spanner instance and a database. The database does not need to exist.
+  const instance = spanner.instance(instanceId);
+  const database = instance.database(databaseId);
+
+  async function updateDatabaseWithProtoDescriptor() {
+    // Reading proto descriptor file
```
olavloite, line 44:
```suggestion
    // Read a proto descriptor file and convert it to a base64 string
```

samples/database-update-with-proto-descriptor.js
```diff
@@ -0,0 +1,89 @@
...
+    const protoDescriptor = fs
+      .readFileSync('./resource/descriptors.pb')
+      .toString('base64');
+
+    // Updated a new database with an extra statements which will create
+    // proto bundle and tables with proto column
```
olavloite, line 50:
```suggestion
    // Add a proto bundle and a table with proto columns to an existing database.
```

olavloite, line 50:
```suggestion
    // Add a proto bundle and a table with proto columns to an existing database.
```

samples/database-update-with-proto-descriptor.js
```diff
@@ -0,0 +1,89 @@
...
+            FirstName  STRING(1024),
+            LastName   STRING(1024),
+            SingerInfo spanner.examples.music.SingerInfo,
+            SingerGenre spanner.examples.music.Genre,
+            SingerInfoArray ARRAY<spanner.examples.music.SingerInfo>,
+            SingerGenreArray ARRAY<spanner.examples.music.Genre>,
+            ) PRIMARY KEY (SingerId)`;
+
+    // Creating empty database
+    await database.create();
```
olavloite, line 67: Move this statement out of the function (for example to line 42) so it does not seem to be part of the sample.

samples/proto-column-dml-insert.js
```diff
@@ -0,0 +1,111 @@
...
+      fullName: 'spanner.examples.music.SingerInfo',
+    });
+
+    const protoEnum = Spanner.protoEnum({
+      value: genre,
+      enumObject: music.Genre,
+      fullName: 'spanner.examples.music.Genre',
+    });
+
+    // Instantiate Spanner table objects.
```
olavloite, line 67: This comment seems out of place

samples/proto-column-dml-insert.js
```diff
@@ -0,0 +1,111 @@
...
+                VALUES (1, 'Virginia', 'Watson', @id1, @id2, @id3, @id4)`,
+          params: {
+            id1: protoMessage,
+            id2: genre,
+            id3: [protoMessage],
+            id4: [protoEnum],
+          },
+        });
+
+        const rowCount = Math.floor(stats[stats.rowCount]);
```
olavloite, line 88: Huh? Why do we need to do round this down?

samples/proto-column-mutation-insert.js
```diff
@@ -0,0 +1,101 @@
...
+      fullName: 'spanner.examples.music.SingerInfo',
+    });
+
+    const protoEnum = Spanner.protoEnum({
+      value: genre,
+      enumObject: music.Genre,
+      fullName: 'spanner.examples.music.Genre',
+    });
+
+    // Instantiate Spanner table object.
```
olavloite, line 67: 'Instantiate' sounds like we are creating it

```suggestion
    // Get a reference to the Singers table
```

samples/proto-column-query.js
```diff
@@ -0,0 +1,93 @@
...
+                   FirstName,
+                   LastName,
+                   SingerInfo,
+                   SingerGenre,
+                   SingerInfoArray,
+                   SingerGenreArray
+            FROM Singers 
+            WHERE SingerId = 1`,
+      // `columnInfo` is an optional parameter and is used to deserialize the proto message and enum object back from bytearray,
+      // if columnInfo is not passed for proto messages and enums then data for these columns will be bytes and int respectively.
```
olavloite, line 58:
```suggestion
      // `columnInfo` is an optional parameter and is used to deserialize the proto message and enum object back from bytearray.
      // If columnInfo is not passed for proto messages and enums, then the data types for these columns will be bytes and int respectively.
```

samples/proto-column-read.js
```diff
@@ -0,0 +1,96 @@
...
+        'FirstName',
+        'LastName',
+        'SingerInfo',
+        'SingerGenre',
+        'SingerInfoArray',
+        'SingerGenreArray',
+      ],
+      keys: ['1'],
+      // `columnInfo` is an optional parameter and is used to deserialize the proto message and enum object back from bytearray,
+      // if columnInfo is not passed for proto messages and enums then data for these columns will be bytes and int respectively.
```
olavloite, line 62:
```suggestion
      // `columnInfo` is an optional parameter and is used to deserialize the proto message and enum object back from bytearray.
      // If columnInfo is not passed for proto messages and enums, then the data types for these columns will be bytes and int respectively.
```

olavloite, line 62:
```suggestion
      // `columnInfo` is an optional parameter and is used to deserialize the proto message and enum object back from bytearray.
      // If columnInfo is not passed for proto messages and enums, then the data types for these columns will be bytes and int respectively.
```

src/codec.ts
```diff
@@ -30,6 +30,27 @@ export interface Field {
   value: Value;
 }
 
+export interface IProtoMessageParams {
+  value: object;
+  // fully qualified name of the proto representing the enum definition
```
olavloite, line 35: Is this field only for enums? Or also for proto objects?

src/codec.ts
```diff
@@ -30,6 +30,27 @@ export interface Field {
...
+  fullName: string;
+  // Function representing the message with helper methods for deserializing
+  // and serializing messages. Method to serialize (i.e. "encode") should be
+  // present when writing data and method to deserialize (i.e. "decode") should
+  // be present while reading/querying data.
+  // User needs to pass typings/js generated by protobufjs.
```
olavloite, line 41: As someone with no experience with proto-columns, I have absolutely no idea what this means. Can we have a more concrete description, maybe with an example?

src/codec.ts
```diff
@@ -30,6 +30,27 @@ export interface Field {
...
+  messageFunction?: Function;
+}
+
+export interface IProtoEnumParams {
+  value: string | number;
+  // fully qualified name of the proto representing the enum definition
+  fullName: string;
+  // Object containing enum string to id mappings.
+  // Users can use it from typings/js generated by protobufjs.
```
olavloite, line 50: Could we have an example for what is expected here?

src/codec.ts
```diff
@@ -343,9 +438,15 @@ function convertValueToJson(value: Value, options: JSONOptions): Value {
  *
  * @param {*} value Value to decode
  * @param {object[]} type Value type object.
+ * @param {Function | object} [columnInfo] Additional parameter which
+ * may help in deserialize data coming from backend.
```
olavloite, line 442: When should I use this? How should I use this?

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
   static pgJsonb(value): PGJsonb {
     return new codec.PGJsonb(value);
   }
+
+  /**
+   * @typedef IProtoMessageParams
+   * @property {object} value Proto Message Value. Serialized Buffer or message object.
+   * @property {string} fullName Proto message Full Qualified Path Name.
+   * @property {Function} [messageFunction] Function generated by protobufs which
+   * has helper methods to deserializing and serializing messages.
```
olavloite, line 1689:
```suggestion
   * @property {Function} [messageFunction] Function generated by protobufjs containing
   * helper methods for deserializing and serializing messages.
```

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
...
+   */
+  /**
+   * Helper function to get a Cloud Spanner proto Message object.
+   *
+   * @param {IProtoMessageParams} value The proto message value params in the format of
```
olavloite, line 1694: In the format of .... icecream? I'll have vanilla, please! ;-)

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
...
+  }
+
+  /**
+   * @typedef IProtoEnumParams
+   * @property {string | number} value Proto Enum Value. Enum String Constant
+   * or Integer Constant.
+   * @property {string} fullName Proto Enum Full Qualified Path Name.
+   * @property {object} [enumObject] An object which have mapping from enum
+   * string to integer constant. User can use it from typings/js generated by
+   * protobufjs.
```
olavloite, line 1719: Maybe simplify this a bit?

```suggestion
   * @property {object} [enumObject] An enum object generated by protobufjs.
```

src/partial-result-stream.ts
```diff
@@ -50,11 +50,16 @@ interface RequestFunction {
  *     that it is not ready for any more data. Increase this value if you
  *     experience 'Stream is still not ready to receive data' errors as a
  *     result of a slow writer in your receiving stream.
+ * @property {object} [columnInfo] An object map using which we can pass
+ * additional properties for each column type which can help in deserializing
+ * the data coming from backend. (Eg: We need to pass Proto Function and Enum
+ * map to deserialize proto messages and enums respectively)
```
olavloite, line 56: Could this be made more concrete? When do I use it? How do I use it?

src/codec.ts
```diff
@@ -231,6 +252,76 @@ export class PGNumeric {
...
+ */
+export class ProtoEnum {
+  value: string;
+  fullName: string;
+  enumObject?: object;
+
+  constructor(protoEnumParams: IProtoEnumParams) {
+    this.fullName = protoEnumParams.fullName;
+
+    if (/^\d+$/.test(protoEnumParams.value.toString())) {
```
skuruppu, line 303: Add a comment here to explain that a value could also be a number so we're checking that here. Otherwise a reader unfamiliar with regexes might not realize what's going on here.

skuruppu: Actually I'm not sure why `value` can't be `string | number` as it is in the interface. Is there a reason why it's a string?

I think I can see below that in `encodeValue()` you can then just return the value. But still wonder if the value can be stored as is and converted to other formats as required.

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
   static pgJsonb(value): PGJsonb {
     return new codec.PGJsonb(value);
   }
+
+  /**
+   * @typedef IProtoMessageParams
+   * @property {object} value Proto Message Value. Serialized Buffer or message object.
```
skuruppu, line 1686: Could we be more specific about the type here instead of it being `object`?

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
...
+   * @property {string} fullName Proto message Full Qualified Path Name.
```
skuruppu, line 1687: Does the upper casing on the comments refer to something special like a type name? E.g. "Serialized Buffer", "Full Qualified Path Name". If not, then just writing a sentence like "Fully-qualified path name of proto message." is preferable.

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
...
+   * ```
+   * const {Spanner} = require('@google-cloud/spanner');
+   * const protoMessage = Spanner.protoMessage({
+   *   value: singerInfo,
+   *   messageFunction: music.SingerInfo,
+   *   fullName: "spanner.examples.music.SingerInfo"
+   * });
+   * ```
+   */
+  static protoMessage(value: IProtoMessageParams): ProtoMessage {
```
skuruppu, line 1708: `params` might be a better variable name here than `value`.

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
...
+    return new codec.ProtoMessage(value);
+  }
+
+  /**
+   * @typedef IProtoEnumParams
+   * @property {string | number} value Proto Enum Value. Enum String Constant
+   * or Integer Constant.
+   * @property {string} fullName Proto Enum Full Qualified Path Name.
```
skuruppu, line 1716: Same here regarding capitalization of words. It's preferable to have sentences.

```
value Proto Enum value as a string constant or an integer constant.
...
```

src/index.ts
skuruppu, line 1719:
```suggestion
   * @property {object} [enumObject] An object which has a mapping from an Enum
   * string to an integer constant. User can use it from typings/js generated by
   * protobufjs.
```

src/index.ts
```diff
@@ -1676,6 +1680,65 @@ class Spanner extends GrpcService {
...
+   * ```
+   * const {Spanner} = require('@google-cloud/spanner');
+   * const protoEnum = Spanner.protoEnum({
+   *   value: 'ROCK',
+   *   enumObject: music.Genre,
+   *   fullName: "spanner.examples.music.Genre"
+   * });
+   * ```
+   */
+  static protoEnum(value: IProtoEnumParams): ProtoEnum {
```
skuruppu, line 1738: Same here re variable name. `params` would be preferable to `value`.

src/partial-result-stream.ts
```diff
@@ -50,11 +50,16 @@ interface RequestFunction {
  *     that it is not ready for any more data. Increase this value if you
  *     experience 'Stream is still not ready to receive data' errors as a
  *     result of a slow writer in your receiving stream.
+ * @property {object} [columnInfo] An object map using which we can pass
```
skuruppu, line 53:
```suggestion
 * @property {object} [columnInfo] An object map that can be used to pass
```

src/partial-result-stream.ts
skuruppu, line 56:
```suggestion
 * map to deserialize proto messages and enums, respectively.)
```

skuruppu: I'm also confused by what this is and what it's meant to look like. An example in the doc string might help.

src/partial-result-stream.ts
```diff
@@ -50,11 +50,16 @@ interface RequestFunction {
...
+ * additional properties for each column type which can help in deserializing
+ * the data coming from backend. (Eg: We need to pass Proto Function and Enum
+ * map to deserialize proto messages and enums respectively)
  */
 export interface RowOptions {
   json?: boolean;
   jsonOptions?: JSONOptions;
   maxResumeRetries?: number;
+  columnInfo?: object;
```
skuruppu, line 62: Could this have a more specific type? That might be one way to make it clear what's meant to be in this object.

src/transaction.ts
```diff
@@ -59,6 +59,7 @@ export interface RequestOptions {
   jsonOptions?: JSONOptions;
   gaxOptions?: CallOptions;
   maxResumeRetries?: number;
+  columnInfo?: object;
```
skuruppu, line 62: Consider giving this a more specific type. For the other fields, it's easy to look up what fields are meant to be set in those options. Here, we have no idea.

skuruppu (changes requested):
I only ended up reviewing the src/ files for now. Once we discuss the comments, I'm happy to look into tests and samples.

src/codec.ts
gauravpurohit06: This interface one is only for proto messageobjects.

src/codec.ts
```diff
@@ -231,6 +252,76 @@ export class PGNumeric {
...
+    } else if (protoMessageParams.messageFunction) {
+      this.value = protoMessageParams.messageFunction['encode'](
+        protoMessageParams.value
+      ).finish();
+    } else {
+      throw new GoogleError(`protoMessageParams cannot be used for constructing ProtoMessage. 
+          Please pass serializedBytes as value or message object with messageFunction`);
+    }
+
+    this.messageFunction = protoMessageParams.messageFunction;
```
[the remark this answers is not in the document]
gauravpurohit06: We can set it there as well, moving it at the same place.

src/codec.ts
```diff
@@ -30,6 +30,27 @@ export interface Field {
   value: Value;
 }
 
+export interface IProtoMessageParams {
+  value: object;
```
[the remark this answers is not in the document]
gauravpurohit06: Actually not, while creating `ProtoMessage` we have kept the value as `Buffer` but to construct it user can pass the value as `Buffer` if they serialised it on their own and but they can also pass the object then we need to do the deserialisation.

src/codec.ts
gauravpurohit06: In the interface value can be `string | number` because we are giving flexibility to user to either pass enum integer constant or enum-string (like FOLK, ROCK etc in singers example). 

The wire type of ENUM is `string` representation of it's integer constant, so if user pass a numeric value we will assign it as is  and if user passes enum string  we will find it's it's integer constant and then assign it.

src/index.ts
gauravpurohit06: The object here would be message object generated by `protobufjs-cli`,  so I think we can't more specific here. Adding docs to make it more clear.

src/partial-result-stream.ts
```diff
@@ -50,11 +50,30 @@ interface RequestFunction {
  *     that it is not ready for any more data. Increase this value if you
  *     experience 'Stream is still not ready to receive data' errors as a
  *     result of a slow writer in your receiving stream.
+ * @property {object} [columnsMetadata] An object map that can be used to pass
+ * additional properties for each column type which can help in deserializing
+ * the data coming from backend. (Eg: We need to pass Proto Function and Enum
+ * map to deserialize proto messages and enums, respectively.)
```
olavloite, line 56: This and also the comment below on `columnsMetadata` is too abstract. Could we add a concrete example of what we expect here?

src/partial-result-stream.ts
```diff
@@ -50,11 +50,30 @@ interface RequestFunction {
...
+   * An object where column names as keys and custom objects as corresponding
+   * values for deserialization. It's specifically useful for data types like
+   * protobuf where deserialization logic is on user-specific code. When provided,
+   * the custom object enables deserialization of backend-received column data.
+   * If not provided, data remains serialized as buffer for Proto Messages and
+   * integer for Proto Enums.
+   *
+   * @example
+   * To obtain Proto Messages and Proto Enums as JSON objects, the customer needs
+   * to supply a custom parameter. This parameter should include the protobufjs-cli
```
olavloite, line 72:
```suggestion
   * To obtain Proto Messages and Proto Enums as JSON objects, you must
   * supply additional metadata. This metadata should include the protobufjs-cli
```

src/transaction.ts
```diff
@@ -59,6 +59,21 @@ export interface RequestOptions {
...
+   * An object where column names as keys and custom objects as corresponding
+   * values for deserialization. It's specifically useful for data types like
+   * protobuf where deserialization logic is on user-specific code. When provided,
+   * the custom object enables deserialization of backend-received column data.
+   * If not provided, data remains serialized as buffer for Proto Messages and
+   * integer for Proto Enums.
+   *
+   * @example
+   * To obtain Proto Messages and Proto Enums as JSON objects, the customer needs
+   * to supply a custom parameter. This parameter should include the protobufjs-cli
```
olavloite, line 72: Also this comment and also the one above could really use a concrete example of what we expect users to pass in. The current description is very abstract and will just be confusing for users who are not using proto columns.

I would suggest that we:
1. Change the wording to 'this is only needed for proto columns' (as opposed to 'it is specifically useful for data types like protobu', as that sounds like it could also be used for other data types)
2. Add a concrete example of what people should pass in based on our commonly used examples that should be recognizable for our users.

```suggestion
   * To obtain Proto Messages and Proto Enums as JSON objects, you must
   * supply a additional metadata. This metadata should include the protobufjs-cli
```

samples/database-update-with-proto-descriptor.js
```diff
@@ -0,0 +1,88 @@
...
+  // const databaseId = 'my-database-id';
+
+  // Imports the Google Cloud client library
+  const {Spanner} = require('@google-cloud/spanner');
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+  // Gets a reference to a Cloud Spanner instance and a database. The database does not need to exist.
```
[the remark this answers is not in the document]
gauravpurohit06: In this case,  it won't fail because we are updating the schema with new DDL which doesn't include `Create` statement.

src/transaction.ts
gauravpurohit06: Thanks Knut for the suggestion. I have added an example. Once the PR is merged. I will also add a link to the generated file.

samples/database-update-with-proto-descriptor.js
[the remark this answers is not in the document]
olavloite: But there is a `database.create()` statement on line 44. Won't that fail if the database already exists?

olavloite (approved):

PR #1845, opened 2023-05-09
docs: update docs-devsite.sh to use latest node-js-rad version

dansaadati:
Towards #1813 , this should resolve the AST issues that were causing the docs job fail. We should also remove unused samples from the documentation, as they are causing further failures.


dansaadati:
> _Warning_: This pull request is touching the following templated files:
> 
> * .kokoro/release/docs-devsite.sh - `.kokoro` files are templated and should be updated in [`synthtool`](https://github.com/googleapis/synthtool)

Will update in synthtool instead.

PR #1847, opened 2023-05-11
fix: set grpc useragent

surbhigarg92:


src/common-grpc/service.ts
```diff
@@ -417,6 +417,12 @@ export class GrpcService extends Service {
 
     this.maxRetries = options.maxRetries;
     this.userAgent = util.getUserAgentFromPackageJson(config.packageJson);
+
+    if (this.providedUserAgent) {
+      this.userAgent = `${this.providedUserAgent} ${this.userAgent}`;
+    }
+    options['grpc.primary_user_agent'] = this.userAgent;
```
olavloite, line 424: Were you able to confirm that this value is actually being sent to the server?

surbhigarg92: Yes. Customer running a sample using this PR, and they were able to see the user-agent. 

Also confirmed with JS-Team , refer https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/transport.ts#L129

olavloite: Ack, thanks for confirming

olavloite (approved):

PR #1863, opened 2023-06-07
feat: cleaning long running transactions by recycling sessions

asthamohta:
Automatically cleaning long running transactions


src/session-pool.ts
```diff
@@ -1036,6 +1151,71 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+   *
+   * @private
+   *
+   * @param {number} longRunningTransactionTimeout Timeout for recycling sessions.
+   */
+  async _deleteLongRunningTransactions(
+    longRunningTransactionTimeout: number = 1000 * 60 * 60
+  ): Promise<void> {
+    // Check lock, only one cleanup should take place at a time
+    if (this._ongoingTransactionDeletion) return;
```
surbhigarg92, line 1165: Add a test to verify that multiple background tasks are not getting triggered. Something like below
```
session1 = getsession()
session2= getsession()
await promise.all(session1, session2)
```

src/session-pool.ts
```diff
@@ -1036,6 +1151,71 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+      this._ongoingTransactionDeletion = false;
+      return;
+    }
+
+    // Traverse each ongoing transaction and check if it needs to be recycled
+    for (const session of this._traces.keys()) {
+      if (session.lastUsed && !session.longRunningTransaction) {
+        const diff = Date.now() - session?.lastUsed;
+        if (
+          !session.longRunningTransaction &&
```
surbhigarg92, line 1182: Condition `!session.longRunningTransaction` is already added in the above if condition. 

Can we merge both if statement into one like `session.lastUsed && !session.longRunningTransaction && diff > longRunningTransactionTimeout`

src/session-pool.ts
```diff
@@ -1036,6 +1151,71 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+            'Long running transaction! Transaction has been closed as it was running for ' +
+              'more than 60 minutes. For long running transactions use batch or partitioned transactions.'
+          );
+          if (this.options.closeInactiveTransactions) {
+            // Recycle session if closeInactiveTransactions is true and log if logging is true.
+            if (this.options.logging) {
+              this.database.logger?.log('warn', sessionTrace);
+            }
+
+            this._recycledTransactions.set(session.txn!, sessionTrace);
```
surbhigarg92, line 1196: Why do we need to store sessionTrace object with transaction object ?

src/session-pool.ts
```diff
@@ -1036,6 +1151,71 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+
+    // Traverse each ongoing transaction and check if it needs to be recycled
+    for (const session of this._traces.keys()) {
+      if (session.lastUsed && !session.longRunningTransaction) {
+        const diff = Date.now() - session?.lastUsed;
+        if (
+          !session.longRunningTransaction &&
+          diff > longRunningTransactionTimeout
+        ) {
+          let sessionTrace = SessionPool.formatTrace(
```
surbhigarg92, line 1185: Move this inside if condition to avoid calling format trace when logging is not enabled https://github.com/googleapis/nodejs-spanner/pull/1863/files#diff-d6f67152c7ddf6462713d84c5314cab643222c347a9b1c87f59498d8cd9c84abR1192

src/transaction.ts
```diff
@@ -359,6 +359,11 @@ export class Snapshot extends EventEmitter {
     const callback =
       typeof gaxOptionsOrCallback === 'function' ? gaxOptionsOrCallback : cb!;
 
+    if (!this.session) {
+      throw new GoogleError(
+        'Transaction has been closed as it was running for more than 60 minutes'
```
surbhigarg92, line 364: Should we add this error message `Transaction has been closed as it was running for more than 60 minutes` in common file and use it at all the places. This would avoid change at multiple places if we have to change it later

surbhigarg92 (changes requested):

src/session-pool.ts
```diff
@@ -681,25 +766,49 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+   * @param {boolean} longRunningTransaction Is transaction long-running.
    * @private
    *
    * @return {Session}
    */
-  _borrowFrom(): Session {
+  _borrowFrom(longRunningTransaction: boolean): Session {
     const session = this._inventory.sessions.pop()!;
     this._inventory.borrowed.add(session);
+    session.lastUsed = Date.now();
```
olavloite, line 777: Would it be better to call this property something like `checkedOutAt`. `lastUsed` sounds like the last time a statement was executed on this session.

src/session-pool.ts
```diff
@@ -681,25 +766,49 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+    // Check if more than 95% of the pool is exhausted and if cleanup is required.
+    if (
+      (this.options.closeInactiveTransactions || this.options.logging) &&
+      this.borrowed / maxSessions >= 0.95
+    ) {
+      this._startCleaningLongRunningSessions();
+    }
+
+    // Log if entire session pool is exhausted.
+    if (this.options.logging && this.borrowed / maxSessions === 1.0) {
```
olavloite, line 791:
```suggestion
    if (this.options.logging && this.borrowed === maxSessions) {
```

src/session-pool.ts
```diff
@@ -681,25 +766,49 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+      this.database.logger?.log(
+        'info',
+        '100% of the session pool is exhausted.' +
+          'Increase max sessions in Session Pool Options as per requirement.'
```
olavloite, line 795:
```suggestion
          ` Increase max sessions in SessionPoolOptions if your application needs to execute more than ${maxSessions} transactions in parallel. Check your application for session leaks if this warning is unexpected, for example if your application would normally not execute more than ${maxSessions} transactions in parallel.`
```

And maybe also add hints on 'how to check for session leaks' in the above warning? Maybe even log the traces of the currently checked out sessions?

src/session-pool.ts
```diff
@@ -1036,6 +1151,71 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     session.txn = transaction;
   }
 
+  /**
+   * recycles sessions with long-running transactions
```
olavloite, line 1155: Are we recycling the sessions (as per this comment)? Or are we deleting them (as per the function name)? Or are we removing them from the session pool?

src/session-pool.ts
```diff
@@ -1036,6 +1151,71 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+  async _deleteLongRunningTransactions(
+    longRunningTransactionTimeout: number = 1000 * 60 * 60
+  ): Promise<void> {
+    // Check lock, only one cleanup should take place at a time
+    if (this._ongoingTransactionDeletion) return;
+    this._ongoingTransactionDeletion = true;
+
+    // Stop cleanup after 60 minutes until triggered again unless a session has
+    // been recycled in the last 60 minutes.
+    if (Date.now() - this._lastSessionRecycle! > 60 * 60 * 1000) {
```
olavloite, line 1170: Are we sure that `this._lastSessionRecycle` has a value here? I see that it is being set to `null` below.

src/session-pool.ts
```diff
@@ -1036,6 +1151,71 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+      this._stopCleaningLongRunningSessions();
+      this._lastSessionRecycle = null;
+      this._ongoingTransactionDeletion = false;
+      return;
+    }
+
+    // Traverse each ongoing transaction and check if it needs to be recycled
+    for (const session of this._traces.keys()) {
+      if (session.lastUsed && !session.longRunningTransaction) {
+        const diff = Date.now() - session?.lastUsed;
```
olavloite, line 1180: nit:
```suggestion
        const diff = Date.now() - session.lastUsed;
```

src/database.ts
```diff
@@ -427,6 +430,7 @@ class Database extends common.GrpcServiceObject {
       Object.assign({}, queryOptions),
       Database.getEnvironmentQueryOptions()
     );
+    this.loggingEnabled = false;
```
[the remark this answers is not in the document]
asthamohta: This is set false by default here but will get set as true when loggingEnabled function is called from the session pool constructor. If logging is false in the session pool options, then this will remain to be false

src/database.ts
```diff
@@ -1994,11 +2057,18 @@ class Database extends common.GrpcServiceObject {
   getTransaction(
     callback?: GetTransactionCallback
   ): void | Promise<[Transaction]> {
-    this.pool_.getSession((err, session, transaction) => {
+    this.pool_.getSession(false, (err, session, transaction) => {
       if (!err) {
         this._releaseOnEnd(session!, transaction!);
+        session!.txn = transaction;
+      }
+      if (
```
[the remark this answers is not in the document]
asthamohta: This is because typescripts throws an error now that transaction can be of snapshot type as well: https://github.com/googleapis/nodejs-spanner/pull/1863/files/141b8e9140697c987b4b9ec0438fedbb11baa5e6#diff-96998a2148aa3d52ad3e80303d0ad88fc43a4afd08966b5330a5657021ae52bbR117

src/database.ts
```diff
@@ -304,6 +305,8 @@ class Database extends common.GrpcServiceObject {
   queryOptions_?: spannerClient.spanner.v1.ExecuteSqlRequest.IQueryOptions;
   resourceHeader_: {[k: string]: string};
   request: DatabaseRequest;
+  logger?: winston.Logger;
+  private loggingEnabled: boolean;
```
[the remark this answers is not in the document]
asthamohta: They can do so by calling only the loggingEnabled function which works as a setter for this and disable using the disableLogging

src/database.ts
```diff
@@ -1127,6 +1136,59 @@ class Database extends common.GrpcServiceObject {
...
+   *         transports: [new winston.transports.Console()],
+   *         format: winston.format.combine(
+   *           winston.format.colorize(),
+   *           winston.format.simple()
+   *         ),
+   *       }
+   * database.enableLogging(loggingOptions);
+   * ```
+   */
+  enableLogging(loggingOptions?: winston.LoggerOptions): winston.Logger {
```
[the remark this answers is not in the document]
asthamohta: Currently its only for cleaning up transactions but we might add more functionality in the future for logging.

src/database.ts
```diff
@@ -785,12 +789,14 @@ class Database extends common.GrpcServiceObject {
         ? (optionsOrCallback as TimestampBounds)
         : {};
 
-    this.pool_.getSession((err, session) => {
+    this.pool_.getSession(true, (err, session) => {
```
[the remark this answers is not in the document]
asthamohta: Adding

src/database.ts
```diff
@@ -2218,7 +2288,7 @@ class Database extends common.GrpcServiceObject {
     callback?: PoolRequestCallback
   ): void | Promise<Session> {
     const pool = this.pool_;
-    pool.getSession((err, session) => {
+    pool.getSession(false, (err, session) => {
```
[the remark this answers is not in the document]
asthamohta: Added a check in release: https://github.com/googleapis/nodejs-spanner/pull/1863/files#diff-d6f67152c7ddf6462713d84c5314cab643222c347a9b1c87f59498d8cd9c84abR686

src/session-pool.ts
```diff
@@ -34,7 +34,13 @@ export interface SessionPoolCloseCallback {
 }
 
 /** @deprecated. Use GetSessionCallback instead. */
-export type GetReadSessionCallback = NormalCallback<Session>;
+export type GetReadSessionCallback = {
```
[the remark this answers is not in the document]
asthamohta: This isn't used externally so I don't think so, and the already existing only is a subset of the new one with only optional params added

src/session-pool.ts
```diff
@@ -87,20 +93,20 @@ export interface SessionPoolInterface extends EventEmitter {
    */
   /**
    * When called returns a session.
-   *
+   * @param {boolean} longRunningTransaction Is transaction long-running.
```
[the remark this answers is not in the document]
asthamohta: Adding this as well: i.e. lasts longer than 60 minutes.

src/session-pool.ts
```diff
@@ -351,15 +383,19 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
   _waiters = 0;
   _pingHandle!: NodeJS.Timer;
   _requests: PQueue;
-  _traces: Map<string, trace.StackFrame[]>;
+  _traces: Map<Session, trace.StackFrame[]>;
```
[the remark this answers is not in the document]
asthamohta: As discussed we use this to traverse existing sessions

src/database.ts
```diff
@@ -3068,7 +3142,7 @@ class Database extends common.GrpcServiceObject {
     // eslint-disable-next-line no-constant-condition
     while (true) {
       try {
-        const [session, transaction] = await promisify(getSession)();
+        const [session, transaction] = await promisify(getSession)(false);
```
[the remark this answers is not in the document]
asthamohta: Yes, adding changes for it in constructor for Runner

src/session-pool.ts
asthamohta: Removing _recycledTransactions

src/session-pool.ts
```diff
@@ -369,7 +405,21 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+
+  /**
+   * Checks is the session for a transaction has been recycled and
+   * transaction has been closed.
+   *
+   * @param {Snapshot} transaction The transaction to be checked.
+   * @return {boolean}
+   */
+  transactionClosed(transaction?: Snapshot): boolean {
+    if (transaction && this._recycledTransactions.has(transaction)) {
```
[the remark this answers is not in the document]
asthamohta: Removing this approach so we don't have to store this info

src/session-pool.ts
asthamohta: This value already existed and is also used by pinging mechanism to check is the session needs to be pinged

src/session-pool.ts
asthamohta: It is set in _startCleaningLongRunningSessions

src/session-pool.ts
asthamohta: We recycle the sessions and delete the transaction. Maybe a better name should be used

system-test/spanner.ts
```diff
@@ -6339,6 +6347,49 @@ describe('Spanner', () => {
...
+        max: 1,
+        maxIdle: 1,
+        min: 1,
+        incStep: 1,
+        closeInactiveTransactions: true,
+        logging: false,
+        databaseRole: null,
+      };
+      const database = instance.database(DATABASE.formattedName_, options);
+      _setLongRunningTransactionTimeout(1000 * 60);
```
surbhigarg92, line 6368: Will this test run for 60 secs ? Can we reduce this timeout value ?

src/session-pool.ts
```diff
@@ -451,6 +479,18 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
   }
 
+  /**
+   * Max sessions for session pool.
+   * @type {number}
+   */
+  get maxSessions(): number {
+    return this.options.max
+      ? this.options.max
+      : DEFAULTS.max
```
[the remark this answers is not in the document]
asthamohta: The problem isn't getting solved with the !
I think we can't change this code

surbhigarg92 (approved):

surbhigarg92:
Closing this PR , as we are not going ahead with this solution

PR #1869, opened 2023-06-20
feat: untyped param types

asthamohta:
This features allows customers to not pass paramType in parametrised queries. It gives backend to flexibly match the type. Example: Previously is timestamp was passed as string, it would be passed to backend with string type code and hence an error would be thrown. Now it is passed without any type code and hence backend handles checking its time and the code succeeds.


system-test/spanner.ts
```diff
@@ -275,6 +275,47 @@ describe('Spanner', () => {
...
       });
     }
+    function readUntypedData(column, value, dialect, callback) {
+      const id = generateName('id');
+      const insertData = {
+        Key: id,
+        [column]: value,
+      };
+
+      let table = googleSqlTable;
```
surbhigarg92, line 285: This should also go into if block. `if (dialect === Spanner.GOOGLE_STANDARD_SQL)`

samples/dml.js
```diff
@@ -288,6 +288,21 @@ function updateUsingDmlWithStruct(instanceId, databaseId, projectId) {
         params: {
           name: nameStruct,
         },
+        types: {
```
surbhigarg92, line 291: I am not able to recall, with this change it is mandatory to pass type for dml queries ?

system-test/spanner.ts
asthamohta: We override on line 293 if it is postgres. This is the default value

samples/dml.js
asthamohta: Only for struct. not for any other

surbhigarg92 (approved):

PR #1873, opened 2023-07-04
feat(default_retention_period): Changes to backup.create API to suppo…

satyaprakashg:
…rt the default retention period feature

With the default retention period feature, null value for expiration time is a valid value. If the expiration time is not specified (null) at the time of create backup then the backup will be retained for the duration of maximum retention period.

In this commit, expireTime is made optional in CreateBackupOptions and backup.create API is fixed to handle the null value in the CreateBackupOptions.


satyaprakashg:
@surbhigarg92 Could you please review this diff? Also, please add the "do not merge" label since the backend changes are not merged.

surbhigarg92 (approved):

surbhigarg92:
closing this PR as it no longer required with admin autogeneration

PR #1874, opened 2023-07-04
feat(default_retention_period): Changes to backup.copy API to support the default retention period feature

satyaprakashg:
With the default retention period feature, null value for expiration time is a valid value.
    If the expiration time is not specified (null) at the time of copy backup then the
    source backup expiration time will be used for expiring the destination backup as well.
    
    In this commit: expireTime is already optional in CopyBackupOptions and no changes are
    required to handle the case of missing expireTime in CopyBackupOptions. Adding unit-tests
    for copy backup for the current scenario and a few other scenarios (since they seem to
    be missing).


satyaprakashg:
@surbhigarg92 Could you please review this diff. Also, please add "do not merge" tag on this diff since the backend changes are not merged yet.

test/backup.ts
```diff
@@ -322,6 +323,114 @@ describe('Backup', () => {
...
+          assert.deepStrictEqual(config.headers, backup.resourceHeader_);
+          done();
+        };
+
+        backup.create(
+          {
+            parent: INSTANCE_NAME,
+            backupId: DST_BACKUP,
+            sourceBackup: SRC_BACKUP,
+            expireTime: timestamp,
```
surbhigarg92, line 363: Can't we use `BACKUP_EXPIRE_TIME` directly here, like we are doing in create backup tests ?

test/backup.ts
```diff
@@ -322,6 +323,114 @@ describe('Backup', () => {
...
+      const INSTANCE_NAME = 'instance-name';
+      const SRC_BACKUP = 'src-backup';
+      const DST_BACKUP = 'dst-backup';
+
+      beforeEach(() => {
+        fakeCodec.encode = util.noop;
+        backup = new Backup(INSTANCE, DST_BACKUP, SRC_BACKUP);
+      });
+
+      it('should succeed with basic params', done => {
```
surbhigarg92, line 336: nit: Change test name to `should make the correct request` to keep it consistent.

test/backup.ts
```diff
@@ -322,6 +323,114 @@ describe('Backup', () => {
...
+            parent: INSTANCE_NAME,
+            backupId: DST_BACKUP,
+            sourceBackup: SRC_BACKUP,
+            encryptionConfig: encryptionConfig,
+          },
+          assert.ifError
+        );
+      });
+
+      it('should succeed without expire time', done => {
```
surbhigarg92, line 405: nit: Change name to "should accept null expire time"

test/backup.ts
```diff
@@ -322,6 +323,114 @@ describe('Backup', () => {
...
+            parent: INSTANCE_NAME,
+            backupId: DST_BACKUP,
+            sourceBackup: SRC_BACKUP,
+            expireTime: timestamp,
+          },
+          assert.ifError
+        );
+      });
+
+      it('should succeed with encryption config', done => {
```
surbhigarg92, line 369: nit: Change name to "should accept an encryption config"

surbhigarg92 (approved):

test/backup.ts
surbhigarg92: Got it. Thanks

PR #1904, opened 2023-08-17
test: add test for aborted blind writes

olavloite:


test/spanner.ts
```diff
@@ -3194,6 +3194,56 @@ describe('Spanner with mock server', () => {
...
 
+    it('should apply blind writes only once', async () => {
+      const database = newTestDatabase();
+      let attempts = 0;
+      await database.runTransactionAsync(async tx => {
+        attempts++;
+        if (attempts == 1) {
+          spannerMock.abortTransaction(tx);
+        }
+        tx!.insert('foo', {id: 1, value: 'One'});
```
surbhigarg92, line 3205: this mutation should be added before abort , right ?

olavloite: That does not really matter. The mutation is only buffered in the client, meaning that whether the transaction has already been aborted or not is not relevant for it.

surbhigarg92 (approved):

PR #1910, opened 2023-08-29
fix: idwaiter with multiple requests

### LINKED ISSUE #1908: Transaction Promise.all with multiple queries results in Error: stream.push() after EOF
Upgrading to 6.5.0, we started encountering an issue when using Promise.all with multiple queries in a transaction.

This issue only started as of 6.5.0, (6.4.0 and earlier do not encounter this issue).  

Possible culprits:
- this MR as it was the major change in 6.5.0:  https://github.com/googleapis/nodejs-spanner/pull/1692
- partial results streams getting closed while queries are still in flight

#### Environment details

  - OS: MacOS
  - Node.js version: 18.16.0
  - npm version: 9.5.1
  - `@google-cloud/spanner` version: 6.16.0

#### Steps to reproduce

  1. Create an async transaction that does 3 queries in a promise all:
```
		const spanner = new Spanner(<projectId>);
		const instance = spanner.instance(<instanceName>);
		const database = instance.database(<databaseName>);

		console.log('MJR instanceName', instanceName);
		console.log('MJR databaseName', databaseName);

		await database.runTransactionAsync(async (t) => {
			try {
				async function doRun(query: string): Promise<void> {
					console.log('Starting ', query);
					await t.run(query);
					console.log('Finished ', query);
				}

				const query1 = 'select * from trigger_offer';
				const query2 = 'select * from offer_purchases';
				const query3 = 'select * from active_offer';

				await Promise.all([doRun(query1), doRun(query2), doRun(query3)]);
				await t.commit();
			} catch (err) {
				console.log('MJR snapshot err', JSON.stringify(err));
			} finally {
				t.end();
			}
		});
		// Close the database when finished.
		await database.close();
		spanner.close();
```
  3. Running the code produced this output:
```
[node] MJR instanceName development
[node] MJR databaseName development_mrichard_amp
[node] Starting  select * from trigger_offer
[node] Starting  select * from offer_purchases
[node] Starting  select * from active_offer
[node] Finished  select * from trigger_offer
[node] uncaughtException stream.push() after EOF Error: stream.push() after EOF
[node]     at new NodeError (node:internal/errors:399:5)
[node]     at readableAddChunk (node:internal/streams/readable:285:30)
[node]     at PassThrough.Readable.push (node:internal/streams/readable:234:10)
[node]     at Immediate._onImmediate (/opt/amp/node_modules/@google-cloud/spanner/build/src/partial-result-stream.js:320:25)
[node]     at processImmediate (node:internal/timers:476:21)
[node]     at process.callbackTrampoline (node:internal/async_hooks:130:17)
[node] child process died with code 1
[node] restarting processes
```

surbhigarg92:
Fixes #1908 🦕

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -17,7 +17,7 @@
 import {DateStruct, PreciseDate} from '@google-cloud/precise-date';
 import {promisifyAll} from '@google-cloud/promisify';
 import arrify = require('arrify');
-import Long = require("long")
+import Long = require('long');
 import {EventEmitter} from 'events';
 import {grpc, CallOptions, ServiceError, Status, GoogleError} from 'google-gax';
 import * as is from 'is';
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -17,6 +17,7 @@
 import {DateStruct, PreciseDate} from '@google-cloud/precise-date';
 import {promisifyAll} from '@google-cloud/promisify';
 import arrify = require('arrify');
+import Long = require("long")
 import {EventEmitter} from 'events';
 import {grpc, CallOptions, ServiceError, Status, GoogleError} from 'google-gax';
 import * as is from 'is';
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -101,7 +101,7 @@ export interface ReadRequest extends RequestOptions {
   keys?: string[] | string[][];
   ranges?: KeyRange[];
   keySet?: spannerClient.spanner.v1.IKeySet | null;
-  limit?: number | Long | null;
+  limit?: number | Long | string | null;
   resumeToken?: Uint8Array | null;
   partitionToken?: Uint8Array | null;
   requestOptions?: Omit<IRequestOptions, 'transactionTag'>;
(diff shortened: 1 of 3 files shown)


src/transaction.ts
```diff
@@ -1326,12 +1324,30 @@ export class Snapshot extends EventEmitter {
...
+        .on('data', (chunk) => streamProxy.emit('data', chunk))
+        .on('end', () => streamProxy.emit('end'));
+      });
+
+      return streamProxy;
+    };
+  }
+
+  _releaseWaitingRequests() {
+    this._inlineBeginStarted = false;
```
olavloite, line 1346: I don't think we should be resetting this here. It seems that it goes well even if we do reset it here, but it seems a bit strange, as `inlineBeginStarted` is still true for this transaction.

surbhigarg92: This was an accidental push. I was trying something. It should not be here.

src/transaction.ts
```diff
@@ -101,7 +102,7 @@ export interface ReadRequest extends RequestOptions {
   keys?: string[] | string[][];
   ranges?: KeyRange[];
   keySet?: spannerClient.spanner.v1.IKeySet | null;
-  limit?: number | Long | null;
+  limit?: number | Long | string | null;
```
olavloite, line 105: This seems like an unrelated change?

olavloite (approved):
LGTM

PR #1948, opened 2023-11-14
feat(spanner): add PG.OID support

skuruppu:
The PG.OID type is an `int64` value with a `PG_OID` type annotation, mostly used by pg_catalog tables. This PR adds support for encoding/decoding PG.OID values.

Tests are added for being able to query a PG.OID value, with and without parameters.

User tables can't be created with PG.OID as a column type so no tests were added for DML or mutations. Since the currently available set of pg_catalog tables don't have PG.OID columns, we also don't test being able to read a PG.OID from a table.


system-test/spanner.ts
```diff
@@ -758,6 +758,23 @@ describe('Spanner', () => {
       });
     });
 
+    describe('oids', () => {
+      it('POSTGRESQL should read pgOid values', function (done) {
```
surbhigarg92, line 762: Should we also add a test for null pgOid value like we have added for all other types ?

skuruppu: I added a test for reading a null value. As for writing, OIDs are never meant to be written, which is why there are no insert tests.

src/codec.ts
```diff
@@ -249,6 +249,25 @@ export class PGJsonb {
...
 
+/**
+ * @typedef PGOid
+ * @see Spanner.pgOid
+ */
+export class PGOid extends WrappedNumber {
+  value: string;
+  constructor(value: string) {
+    super();
+    this.value = value.toString();
```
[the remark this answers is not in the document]
skuruppu: I thought this would work but it actually makes the behaviour different. I'm not sure I understand why. But this logic is the same as how we represent the `Int` type on line 144.

surbhigarg92 (approved):

PR #1965, opened 2023-12-08
feat: Managed Autoscaler

alkatrivedi:
This PR contains code changes for manages autoscaler.


samples/instance-with-autoscaling-config.js
```diff
@@ -0,0 +1,93 @@
+/**
+ * Copyright 2023 Google LLC
```
surbhigarg92, line 2:
```suggestion
 * Copyright 2024 Google LLC
```

src/index.ts
```diff
@@ -481,17 +482,21 @@ class Spanner extends GrpcService {
           displayName,
           nodeCount: config.nodes,
           processingUnits: config.processingUnits,
+          autoscalingConfig: config.autoscalingConfig,
```
surbhigarg92, line 485: We need to do changes for [Update Instance](https://github.com/googleapis/nodejs-spanner/blob/main/src/instance.ts#L1598) to handle below use case 
When autoscaling config is set then node and processing unit should not be passed

surbhigarg92 (changes requested):

alkatrivedi:
Closing this PR, as there is no more need of this sample. I will add the similar samples using admin client.

PR #1966, opened 2023-12-08
feat: Support for Directed Reads

surbhigarg92:


asthamohta (approved):
LGTM

src/transaction.ts
```diff
@@ -1288,6 +1325,25 @@ export class Snapshot extends EventEmitter {
     return {params, paramTypes};
   }
 
+  /**
+   * Get directed read options
+   * @private
+   * @param {google.spanner.v1.IDirectedReadOptions} directedReadOptions Request directedReadOptions object.
+   */
+  protected _getDirectedReadOptions(
+    directedReadOptions: IDirectedReadOptions | null | undefined
```
harshachinta, line 1334: nit: In the interfaces we have mentioned `directedReadOptions?: IDirectedReadOptions | null `.
1. Here we are allowing `undefined` as well. In which case does this undefined get passed?
2. Also what would be the default value of `IDirectedReadOptions` when it is not set by user.

src/transaction.ts
```diff
@@ -1288,6 +1325,25 @@ export class Snapshot extends EventEmitter {
...
+  ) {
+    if (
+      !directedReadOptions &&
+      this._getSpanner().directedReadOptions &&
+      this._options.readOnly
```
harshachinta, line 1339: nit: double checking whether the `readOnly` handwritten property is set for partitioned reads and queries class too?

src/transaction.ts
```diff
@@ -457,6 +469,8 @@ export class Snapshot extends EventEmitter {
    *     PartitionReadRequest message used to create this partition_token.
    * @property {google.spanner.v1.RequestOptions} [requestOptions]
    *     Common options for this request.
+   * @property {object} [directedReadOptions]
```
harshachinta, line 472: Customers will be using autogenerated code to build the directed read options. I think in nodejs we allow json object as input. But should this be marked also as `google.spanner.v1.IDirectedReadOptions` so that they know what type of class should be passed as input.
I am not sure of nodejs convention, but mentioning is as object could mean anything and is not clear.

src/transaction.ts
harshachinta: Also single use has this property set right?
To summarize all the below interfaces should have this readOnly property set to true,
1. Single use
2. Read only
3. Partitioned reads
4. Partitioned Query

samples/system-test/spanner.test.js
```diff
@@ -1985,5 +1985,25 @@ describe('Spanner', () => {
         )
       );
     });
+
+    // directed_read_transaction
```
harshachinta, line 1989:
```suggestion
    // directed_read_options
```

test/batch-transaction.ts
```diff
@@ -315,6 +347,7 @@ describe('BatchTransaction', () => {
         table: QUERY.table,
         keySet: fakeKeySet,
         dataBoostEnabled: true,
+        directedReadOptions: fakeDirectedReadOptionsForRequest,
       };
 
       const stub = sandbox.stub(batchTransaction, 'createPartitions_');
```
harshachinta, line 353: Can we write a unit test to validate `transaction.execute(partition) `?
Basically check, if the ReadPartition or QueryPartition has the directed read options set, then the transaction.execute(partition), should make correct ReadRequest/ExecuteSqlRequest with directedReadOptions set in its request.

test/database.ts
```diff
@@ -2525,6 +2561,35 @@ describe('Database', () => {
     it('should accept requestOptions', () => {
       const fakeCallback = sandbox.spy();
 
+      database.runPartitionedUpdate(
+        {
+          sql: QUERY.sql,
+          params: QUERY.params,
+          requestOptions: {priority: RequestOptions.Priority.PRIORITY_LOW},
+          directedReadOptions: fakeDirectedReadOptions,
```
harshachinta, line 2569: Why are we setting directedReadOptions  in runPartitionedUpdate?

test/transaction.ts
```diff
@@ -392,6 +409,70 @@ describe('Transaction', () => {
...
+
+        snapshot.id = id;
+        snapshot.createReadStream(TABLE);
+
+        const {reqOpts} = REQUEST_STREAM.lastCall.args[0];
+
+        assert.deepStrictEqual(reqOpts, expectedRequest);
+      });
+
+      it('should override directedReadOptions set for client when passed', () => {
```
harshachinta, line 438:
```suggestion
      it('should override directedReadOptions set at client level when passed from request', () => {
```

harshachinta (commented):
Added few minor comments.

test/database.ts
surbhigarg92: Removed

src/transaction.ts
surbhigarg92: Done. Using `google.spanner.v1.IDirectedReadOptions` instead of custom object

src/transaction.ts
surbhigarg92: Yes. This has been validated

test/batch-transaction.ts
surbhigarg92: In unit test we can not have e2e test. But I have added the test case for below scenarios , create partitions should return directedReadOptions and execute should take directedReadOptions

src/transaction.ts
```diff
@@ -956,6 +966,15 @@ export class Snapshot extends EventEmitter {
       .on('end', () => callback!(null, rows, stats, metadata));
   }
 
+  /**
+   * Object for Directed Read Options.
+   *
+   * @typedef {object} IDirectedReadOptions
+   * @property {google.spanner.v1.IRequestOptions} includeReplicas Contains a repeated set of
```
harshachinta, line 973: Why `google.spanner.v1.IRequestOptions`?

src/transaction.ts
```diff
@@ -86,6 +86,7 @@ export interface ExecuteSqlRequest extends Statement, RequestOptions {
   queryOptions?: IQueryOptions;
   requestOptions?: Omit<IRequestOptions, 'transactionTag'>;
   dataBoostEnabled?: boolean | null;
+  directedReadOptions?: google.spanner.v1.IDirectedReadOptions;
```
harshachinta, line 89: just checking if customers can still pass `null` here right?

harshachinta (approved):
LGTM

src/transaction.ts
surbhigarg92: We have '?' which marks the property optional

src/transaction.ts
surbhigarg92: This was a miss. Removed this

PR #1983, opened 2024-01-18
docs: Samples and test for backups admin apis

alkatrivedi:
This PR contains samples and its integration tests for admin client backups APIs.


samples/v2/backups-cancel.js
```diff
@@ -0,0 +1,81 @@
...
+  // const instanceId = 'my-instance';
+  // const databaseId = 'my-database';
+  // const backupId = 'my-backup';
+
+  // Creates a client
+  const spanner = new Spanner({
+    projectId: projectId,
+  });
+
+  const databaseAdminClient = spanner.get_database_admin_client();
```
arpan14: Does this method now need to change based on latest updates?

PR #1987, opened 2024-01-20
docs: Samples and tests for Instance Admin Client APIs

alkatrivedi:
This PR contains sample files and their integration tests for the instance admin client APIs.


samples/v2/instance.js
```diff
@@ -0,0 +1,110 @@
...
+      instanceId: instanceID,
+      instance: {
+        config: instanceAdminClient.instanceConfigPath(
+          projectID,
+          'regional-us-central1'
+        ),
+        nodeCount: 1,
+        displayName: 'Display name for the instance.',
+        labels: {
+          ['cloud_spanner_samples']: 'true',
```
olavloite, line 54: Why is `cloud_spanner_samples` within quotes and brackets?

samples/v2/instance.js
```diff
@@ -0,0 +1,110 @@
...
+          created: Math.round(Date.now() / 1000).toString(), // current time
+        },
+      },
+      parent: instanceAdminClient.projectPath(projectID),
```
olavloite, line 58: nit: I think it would be better to have this line at the top together with the `instanceId`

samples/v2/instance.js
```diff
@@ -0,0 +1,110 @@
...
+    });
+
+    console.log(`Waiting for operation on ${instanceID} to complete...`);
+    await operation.promise();
```
olavloite, line 62: Does `await operation.promise()` return anything? If so, we should show what.

samples/v2/instance-config-create.js
```diff
@@ -0,0 +1,98 @@
...
+          ),
+          baseConfig: instanceAdminClient.instanceConfigPath(
+            projectID,
+            baseInstanceConfigID
+          ),
+          displayName: instanceConfigID,
+          replicas: baseInstanceConfig.replicas,
+          optionalReplicas: baseInstanceConfig.optionalReplicas,
+        },
+        parent: instanceAdminClient.projectPath(projectID),
```
olavloite, line 74: nit: I think it would be easier to read if this was put at the top together with the `instanceConfigId`

samples/v2/instance-config-create.js
```diff
@@ -0,0 +1,98 @@
...
+          name: instanceAdminClient.instanceConfigPath(
+            projectID,
+            instanceConfigID
+          ),
+          baseConfig: instanceAdminClient.instanceConfigPath(
+            projectID,
+            baseInstanceConfigID
+          ),
+          displayName: instanceConfigID,
+          replicas: baseInstanceConfig.replicas,
```
olavloite, line 71: This just creates a copy of an existing instance config. I think it would be more helpful if we showed how to create an instance config that is actually different from an existing config (so add at least one replica)

samples/v2/instance-config-get-operations.js
```diff
@@ -0,0 +1,78 @@
...
+    // Lists the instance config operations.
+    try {
+      console.log(
+        `Getting list of instance config operations on project ${projectId}...\n`
+      );
+      const [instanceConfigOperations] =
+        await instanceAdminClient.listInstanceConfigOperations({
+          parent: instanceAdminClient.projectPath(projectId),
+          filter:
+            '(metadata.@type=type.googleapis.com/google.spanner.admin.instance.v1.CreateInstanceConfigMetadata)',
```
olavloite, line 48: We should add a comment here that explains that this filters for only `InstanceConfig` operations that create a new instance config.

samples/v2/instance-config-update.js
```diff
@@ -0,0 +1,89 @@
...
+      );
+      const [operation] = await instanceAdminClient.updateInstanceConfig({
+        instanceConfig: {
+          name: instanceAdminClient.instanceConfigPath(
+            projectId,
+            instanceConfigId
+          ),
+          displayName: 'updated custom instance config',
+          labels: {
+            ['updated']: 'true',
```
olavloite, line 59: Why is `updated` in quotes and brackets, while `created` is without?

samples/v2/instance-config-update.js
```diff
@@ -0,0 +1,89 @@
...
+            created: Math.round(Date.now() / 1000).toString(), // current time
+          },
+        },
+        updateMask: (protos.google.protobuf.FieldMask = {
+          paths: ['display_name', 'labels'],
```
olavloite, line 64: We should add a comment that explains why you need to add a `FieldMask`.

samples/v2/instance-with-autoscaling-config.js
```diff
@@ -0,0 +1,112 @@
...
+          'regional-us-central1'
+        ),
+        displayName: 'Display name for the instance.',
+        autoscalingConfig: autoscalingConfig,
+        labels: {
+          ['cloud_spanner_samples']: 'true',
+          created: Math.round(Date.now() / 1000).toString(), // current time
+        },
+      },
+      parent: instanceAdminClient.projectPath(projectID),
```
olavloite, line 81: nit: Put this at the top

samples/v2/instance-with-processing-units.js
```diff
@@ -0,0 +1,76 @@
...
+          'regional-us-central1'
+        ),
+        displayName: 'Display name for the instance.',
+        processingUnits: 500,
+        labels: {
+          ['cloud_spanner_samples']: 'true',
+          created: Math.round(Date.now() / 1000).toString(), // current time
+        },
+      },
+      parent: instanceAdminClient.projectPath(projectID),
```
olavloite, line 56: to the top

samples/v2/get-instance-config.js
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * TODO(developer): Uncomment the following line before running the sample.
+   */
+  // const projectId = 'my-project-id';
+
+  // Imports the Google Cloud client library
+  const {InstanceAdminClient} = require('@google-cloud/spanner/build/src/v1');
+
+  // Creates a client
+  const instanceAdminClient = new InstanceAdminClient({
```
olavloite, line 34: Also based on discussion https://github.com/googleapis/python-spanner/pull/1065#discussion_r1461941195

We need to define a strategy for how this should work with the emulator. Currently, we tell our customers that they can just set the environment variable `SPANNER_EMULATOR_HOST=localhost:9010` and everything will use the emulator. That is no longer true with this setup. We should be consistent across all languages regarding this, both to ease our documentation job, but also to make it easier for customers who do use multiple programming languages.
The simplest solution is probably to instruct customers to create a 'normal' hand-written client, and then get the auto-generated admin client from that hand-written client. That allows us to return the same auto-generated admin client that we would otherwise have used in combination with the hand-written client.

samples/v2/instance-config-get-operations.js
```diff
@@ -0,0 +1,81 @@
...
+  async function getInstanceConfigOperations() {
+    // Lists the instance config operations.
+    try {
+      console.log(
+        `Getting list of instance config operations on project ${projectId}...\n`
+      );
+      const [instanceConfigOperations] =
+        await instanceAdminClient.listInstanceConfigOperations({
+          parent: instanceAdminClient.projectPath(projectId),
+          // filtering on CreateInstanceConfigMetadata fields
```
olavloite, line 49:
```suggestion
          // This filter ensures that only operations with metadata type CreateInstanceConfigMetadata
          // will be returned, which is the same as only returning CreateInstanceConfig operations.
```

olavloite (approved):

samples/v2/get-instance-config.js
```diff
@@ -0,0 +1,64 @@
...
+  const instanceAdminClient = spanner.getInstanceAdminClient();
+
+  async function getInstanceConfig() {
+    // Get the instance config for the multi-region North America 6 (NAM6).
+    // See https://cloud.google.com/spanner/docs/instance-configurations#configuration for a list of all available
+    // configurations.
+    const [instanceConfig] = await instanceAdminClient.getInstanceConfig({
+      name: instanceAdminClient.instanceConfigPath(
+        projectId,
+        'custom-my-instance-config'
```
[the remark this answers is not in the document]
alkatrivedi: The existing sample is containing the NAM6 configuration. I have made changes to the configuration to ensure the alignment with the comment provided.
Additionally, the test corresponding to this sample is skipped in the existing test(spanner class) file with the comment` // TODO: Enable when the feature has been released`, hence I have followed the same for the autogenerated samples

PR #1991, opened 2024-02-02
feat: add support for Proto columns

harshachinta:
This PR adds support for Proto Columns


src/partial-result-stream.ts
```diff
@@ -50,11 +50,52 @@ interface RequestFunction {
  *     that it is not ready for any more data. Increase this value if you
  *     experience 'Stream is still not ready to receive data' errors as a
  *     result of a slow writer in your receiving stream.
+ * @property {object} [columnsMetadata] An object map that can be used to pass
```
surbhigarg92, line 53: nit: This property can only be used with Protos right ? Should we mention that in this document.

system-test/spanner.ts
```diff
@@ -124,15 +129,42 @@ describe('Spanner', () => {
...
+    } else {
+      // Reading proto descriptor file
+      const protoDescriptor = fs
+        .readFileSync('test/data/descriptors.pb')
+        .toString('base64');
+
+      const [, googleSqlOperation1] = await DATABASE.create({
+        schema: [
+          `
+        CREATE PROTO BUNDLE (
```
surbhigarg92, line 152: Why this Proto is not added for Emulator. Is it not supported in emulator ?

system-test/spanner.ts
```diff
@@ -1844,6 +1894,219 @@ describe('Spanner', () => {
       });
     });
 
+    describe('protoMessage', () => {
+      before(async function () {
+        if (IS_EMULATOR_ENABLED) {
+          this.skip();
```
surbhigarg92, line 1900: Why skip for emulator ?

src/codec.ts
```diff
@@ -246,6 +283,83 @@ export class PGNumeric {
...
+    this.messageFunction = protoMessageParams.messageFunction;
+
+    if (protoMessageParams.value instanceof Buffer) {
+      this.value = protoMessageParams.value;
+    } else if (protoMessageParams.messageFunction) {
+      this.value = protoMessageParams.messageFunction['encode'](
+        protoMessageParams.value
+      ).finish();
+    } else {
+      throw new GoogleError(`protoMessageParams cannot be used to construct
```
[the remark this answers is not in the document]
harshachinta: Client side validations were removed in cases where client does not do any transformations to the input and it does not apply here.
For Proto columns, client library exposes some interfaces in which the input proto cannot be directly sent to the backend but needs some logic on the client for conversion. So this validation is still needed on client side.
For ex, `proto_type_fqn` is a field that needs to be dynamically computed on client based on input proto object.

src/partial-result-stream.ts
harshachinta: We have mentioned an example already
`Eg: We need to pass Proto Function and Enum map to deserialize proto messages and enums, respectively`

system-test/spanner.ts
harshachinta: There are issues in emulator which is not yet addressed. Will unskip the tests in a seperate PR when emulator is ready.

system-test/spanner.ts
harshachinta: There are issues in emulator which is not yet addressed. Will unskip the tests in a seperate PR when emulator is ready.

samples/system-test/spanner.test.js
```diff
@@ -2052,4 +2053,79 @@
...
+    });
+
+    it('should add proto message and enum columns', async () => {
+      const output = execSync(
+        `node proto-type-add-column.js "${INSTANCE_ID}" "${PROTO_DATABASE_ID}" ${PROJECT_ID}`
+      );
+      assert.match(
+        output,
+        new RegExp(
+          `Altered table "Singers" on database ${PROTO_DATABASE_ID} on instance ${INSTANCE_ID} with proto descriptors.`
```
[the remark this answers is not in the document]
harshachinta: These are used on tests and not directly on production code due to which this can be ignored

src/codec.ts
```diff
@@ -30,6 +30,43 @@ export interface Field {
...
+  fullName: string;
+  /**
+   * Provide a First Class function that includes nested functions named
+   * "encode" for serialization and "decode" for deserialization of the proto
+   * message. The function should be sourced from the JS file generated by
+   * protobufjs-cli for the proto message. Additionally, please review the
+   * sample at {@link www.samples.com} for guidance.
+   *
+   * ToDo: Update the link
+   */
```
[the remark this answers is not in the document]
harshachinta: Removed the text as discussed

surbhigarg92 (approved):

src/codec.ts
```diff
@@ -30,6 +30,40 @@ export interface Field {
...
+  fullName: string;
+  /**
+   * An object containing enum string to id mapping.
+   * @example: { POP: 0, JAZZ: 1, FOLK: 2, ROCK: 3 }
+   *
+   * The object should be sourced from the JS file generated by
+   * protobufjs-cli for the proto message. Additionally, please review the
+   * sample at {@link www.samples.com} for guidance.
+   *
+   * ToDo: Update the link
```
surbhigarg92, line 62: Remove this todo

samples/database-get-ddl.js
```diff
@@ -55,7 +55,10 @@ function main(instanceId, databaseId, projectId) {
         databaseId
       )}:`
     );
-    ddlStatements.statements.forEach(element => {
+    console.log(
```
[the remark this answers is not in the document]
harshachinta: I have removed the changes in this sample to avoid confusion.

src/codec.ts
harshachinta: Removed the TODO and link as we don't follow this for existing methods

PR #1994, opened 2024-02-13
feat(spanner): add emulator support for the admin client autogenerated API samples

alkatrivedi:
This PR introduces emulator support for the API samples, which are using admin client autogenerated methods.

**Sample usage**

For Instance samples:

```
const spanner = new Spanner({
    projectId: projectId,
  });
 const instanceAdminClient = spanner.get_instance_admin_client();
```

For Database samples:

```
const spanner = new Spanner({
    projectId: projectID,
  });
const databaseAdminClient = spanner.get_database_admin_client();
```


src/index.ts
```diff
@@ -345,6 +350,37 @@ class Spanner extends GrpcService {
       [CLOUD_RESOURCE_HEADER]: this.projectFormattedName_,
     };
     this.directedReadOptions = directedReadOptions;
+    this._instance_admin_api = null;
+    this._database_admin_api = null;
```
olavloite, line 354: nit: Isn't this redundant, as they are already set to null when they are defined?

src/index.ts
```diff
@@ -345,6 +350,37 @@ class Spanner extends GrpcService {
...
+    this.emulatorHost = emulatorHost;
+  }
+
+  // Helper for session-related instance admin client API calls
+  instance_admin_api(this): void {
```
olavloite, line 359: Is this the method that a user should call in order to get an instance admin client? Or is it an internal helper method?

If it's an internal helper method: It should start with an underscore to indicate that it is a private method not intended for public use.

If it's the public method that users should call in order to get an admin client: Then it needs more documentation on what it does, and what users are expected to do with the return value. This function creates a new instance each time that it is called, which is a resource-heavy operation. The user is also responsible for closing the instance when they are done with it, which is different from the current situation where all (admin) clients are managed by the client library, and automatically closed when you close this `Spanner` instance. As it is a resource-heavy operation, we should also document that it could make sense for users to cache the instance that is returned by this method, and reuse it if they intend to execute multiple admin operations.
I would also suggest to rename the method to `create_instance_admin_client`, so it is cleared to the caller that it actually creates a new client.

src/index.ts
alkatrivedi: Updated

src/index.ts
olavloite: Sorry, I just realized that I scanned through this method too quickly (and I'm reviewing too many different versions of this PR in different languages). This implementation does __not__ always create a new instance. It only creates a new instance when it has not yet been created before.

The problem here is that it is not clear to the user what they should do with the client that is being returned:
- Is the client managed by the client library, and will the client library close it when the Spanner client is closed?
- Is the client managed by the user, and should they close it when they are done with it?

Currently, none of the above strategies are implemented in this solution. You should make a clear choice between one of the following:
1. The client is managed by the client library. That then also means that the client that you create here, should be closed when the `Spanner` client is closed. You should then also clearly document this function that the client that is being returned is a shared instance, and that the user should __never__ close the client that is returned themselves.
2. The client is managed by the user. That then means that you should never return a shared instance (so you should create a new one every time), and that the method should be clearly documented that the user is responsible for closing the client that is returned.

src/index.ts
```diff
@@ -345,6 +350,37 @@ class Spanner extends GrpcService {
       [CLOUD_RESOURCE_HEADER]: this.projectFormattedName_,
     };
     this.directedReadOptions = directedReadOptions;
+    this._instance_admin_api = this._instance_admin_api;
+    this._database_admin_api = this._database_admin_api;
```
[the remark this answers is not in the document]
alkatrivedi: Removed

src/index.ts
```diff
@@ -345,6 +350,37 @@ class Spanner extends GrpcService {
...
+    this.emulatorHost = emulatorHost;
+  }
+
+  // Helper for session-related instance admin client API calls
+  create_instance_admin_client(this): void {
+    if (this._instance_admin_api === null || this._instance_admin_api === undefined) {
```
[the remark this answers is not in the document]
alkatrivedi: Done

src/index.ts
alkatrivedi: After taking a look at the code my observation was that the management of the client is handled at the user end. Moreover, as there exists no direct correlation between the `Spanner` client and the client obtained from the `create_instance_admin_client()` and `create_database_admin_client()` methods, closing the` Spanner` client will not automatically close the `Admin` client. Additionally, in the existing samples, where hand-written code is utilized, there is no explicit handling for client closure.

alkatrivedi: Based on offline discussions, I have updated the methods to add the created admin clients into the `clients_ `map. Now, closing the `spanner` client will also close the `admin` client that will get created by calling `get_instance_admin_client()` and `get_database_admin_client()` methods

src/index.ts
```diff
@@ -345,6 +349,54 @@ class Spanner extends GrpcService {
       [CLOUD_RESOURCE_HEADER]: this.projectFormattedName_,
     };
     this.directedReadOptions = directedReadOptions;
+    this.emulatorHost = emulatorHost;
+  }
+
+  // Helper for instance admin client API calls
+  get_instance_admin_client(this): void {
+    const clientName = 'InstanceAdminClient';
```
olavloite, line 357: The `prepareGapicRequest_(..)` function just does the following:

```
      if (!this.clients_.has(clientName)) {
        this.clients_.set(clientName, new v1[clientName](this.options));
      }
      const gaxClient = this.clients_.get(clientName)!;
```

I think that we should do exactly the same here, as they should both return the exact same client. That is:
1. If I for example call `getInstances(..)` on this class, then it will call `prepareGapicRequest_` with client name `InstanceAdminClient`. That will call the above code snippet. Any further calls to this `get_instance_admin_client` will then return the instance that was returned by the first call to `getInstances(..)`.
2. The same is also true the other way around: If I first call `get_instance_admin_client`, then the code in this method will be executed to create an `InstanceAdminClient`. Any subsequent call to `getInstances(..)` will then use the client created here.

The above therefore means that they are clearly intended to be the same, so they should also be created using the same code to prevent any differences between them.

That also means that the additional fields for `_instance_admin_api` and `_database_admin_api` are not needed.

src/index.ts
```diff
@@ -345,6 +349,54 @@ class Spanner extends GrpcService {
       [CLOUD_RESOURCE_HEADER]: this.projectFormattedName_,
     };
     this.directedReadOptions = directedReadOptions;
+    this.emulatorHost = emulatorHost;
+  }
+
+  // Helper for instance admin client API calls
```
olavloite, line 355: Methods that are intended for public use should be documented using jsdoc, so using comments in the form `/** method documented here. */

In this case, the documentation should also include an instruction that the returned client is managed by the Spanner client, and that it should not be closed by the user, but will be closed by the client library when `Spanner.close()` is called.

src/index.ts
```diff
@@ -345,6 +349,54 @@ class Spanner extends GrpcService {
...
+  get_instance_admin_client(this): void {
+    const clientName = 'InstanceAdminClient';
+    if (!this.clients_.has(clientName)) {
+      // add the client created to the clients_ map
+      /* NOTE: when the spanner client is closed, all the clients present in the clients_ map will also be closed as well */
```
olavloite, line 360: nit: prefer to stick to one style of comments, and inside a method like this, the most logical is the `// ...` style.

src/index.ts
```diff
@@ -222,6 +222,8 @@ class Spanner extends GrpcService {
   resourceHeader_: {[k: string]: string};
   routeToLeaderEnabled = true;
   directedReadOptions: google.spanner.v1.IDirectedReadOptions | null;
+  _instance_admin_api: null | undefined;
+  _database_admin_api: null | undefined;
```
olavloite, line 226: These fields are probably not needed, see my comment further down.

src/index.ts
```diff
@@ -236,6 +238,8 @@ class Spanner extends GrpcService {
   static GOOGLE_STANDARD_SQL =
     google.spanner.admin.database.v1.DatabaseDialect.GOOGLE_STANDARD_SQL;
 
+  emulatorHost: {endpoint: string; port?: number | undefined} | undefined;
```
olavloite, line 241: This field is probably not needed, see my comment further down.

alkatrivedi: Removed

src/index.ts
alkatrivedi: Removed

src/index.ts
```diff
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2016 Google Inc. All Rights Reserved.
+ * Copyright 2024 Google Inc. All Rights Reserved.
```
olavloite, line 2: nit: we normally do not update the copyright year of an existing file

```suggestion
 * Copyright 2016 Google Inc. All Rights Reserved.
```

src/index.ts
```diff
@@ -347,6 +353,53 @@ class Spanner extends GrpcService {
     this.directedReadOptions = directedReadOptions;
   }
 
+  /**
+   * Gets the InstanceAdminClient object.
+   * The InstanceAdminClient object is also shared for handwritten client and should not be manually closed.
```
olavloite, line 358:
```suggestion
   * The returned InstanceAdminClient object is a shared, managed instance and should not be manually closed.
```

src/index.ts
```diff
@@ -347,6 +353,53 @@ class Spanner extends GrpcService {
...
+    const clientName = 'InstanceAdminClient';
+    if (!this.clients_.has(clientName)) {
+      this.clients_.set(
+        clientName,
+        new v1[clientName](this.options as ClientOptions)
+      );
+    }
+    return this.clients_.get(clientName)! as v1.InstanceAdminClient;
+  }
+  /**
```
olavloite, line 379: nit: add an empty line between the end of the last method and the start of this method

src/index.ts
```diff
@@ -347,6 +353,53 @@ class Spanner extends GrpcService {
...
+   * Gets the DatabaseAdminClient object.
+   * The DatabaseAdminClient object is also shared for handwritten client and should not be manually closed.
```
olavloite, line 381:
```suggestion
   * The returned DatabaseAdminClient object is a managed, shared instance and should not be manually closed.
```

system-test/spanner.ts
```diff
@@ -231,6 +231,115 @@ describe('Spanner', () => {
...
+        } catch (err) {
+          if (!err) {
+            assert.ifError(err);
+          }
+        }
+      });
+
+      it('should list the instances', async () => {
+        assert(projectId);
+        const [operation] = await instanceAdminClient.listInstances({
```
olavloite, line 305: nit:

```suggestion
        const [instances] = await instanceAdminClient.listInstances({
```

olavloite (approved):

surbhigarg92 (approved):

PR #2013, opened 2024-03-05
chore: admin samples

alkatrivedi:
This PR contains missing admin samples/integration tests.


samples/archived/database-get-default-leader.js
```diff
@@ -1,69 +0,0 @@
-/**
```
arpan14, line 1: Just to put it on record - Can you answer why this file is being removed from `archived` folder?

samples/database-get-default-leader.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2024 Google LLC
+ * Copyright 2021 Google LLC
```
arpan14, line 2: Revert? We should have this as 2024.

samples/get-instance-config.js
```diff
@@ -15,7 +15,7 @@
 
 // sample-metadata:
 //  title: Gets the instance config metadata for the configuration nam6
-//  usage: node get-instance-config.js <PROJECT_ID> <INSTANCE_CONFIG_ID>
+//  usage: node get-instance-config.js <PROJECT_ID>
```
arpan14, line 18: Why are we not sending `INSTANCE_CONFIG_ID` ?

samples/schema.js
```diff
@@ -119,6 +119,60 @@ async function addColumn(instanceId, databaseId, projectId) {
   // [END spanner_add_column]
 }
 
+async function queryDataWithNewColumn(instanceId, databaseId, projectId) {
```
arpan14, line 122: * Is this the only function which was skipped? Are there more samples which got skipped? How have we validated this? 

* I am a bit curious on why this was the only sample which was skipped?

samples/README.md
```diff
@@ -663,7 +663,7 @@ View the [source code](https://github.com/googleapis/nodejs-spanner/blob/main/sa
 __Usage:__
 
 
-`node get-instance-config.js <PROJECT_ID> <INSTANCE_CONFIG_ID>`
```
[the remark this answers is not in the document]
alkatrivedi: The old samples is getting NAM6 cofiguration. In the old samples we are not passing `INSTANCE_CONFIG_ID`. Hence, in the new sample I am keeping the same format. In the sample <INSTANCE_CONFIG_ID> was got added earlier. Hence, fixing it here in this PR

samples/database-get-default-leader.js
alkatrivedi: Since, there is no any change in the old sample. We do not need this update, if done, header-checks will get fail in presubmit

samples/get-instance-config.js
alkatrivedi: No, in the old sample we are not seding the parameter. We are just getting the NAM6 configurations

samples/schema.js
alkatrivedi: Several samples(which were using data client) in `datatypes.js` file were unintentionally got skipped during my previous updates. I have rectified this by ensuring that all the data client samples are now included in the file. And this was not the only file in which the data client samples got skipped, some samples got missed in `schema.js` file too. Samples using the data client were unintentionally got overlooked as I'd focused on updating the samples to utilize the autogen methods in the new samples.

I have now rectified what all samples was got missed adding, I have added them all to make sure they all are now present in file.

arpan14 (approved):

PR #2018, opened 2024-03-08
chore: Skip archived test

surbhigarg92:


harshachinta (approved):
LGTM, but can we make sure we have a plan to run these archived samples in some way? We could otherwise not detect issues with those samples.
Few thoughts on doing this,
1. Run the archived samples in continuous and nightly builds thus leaving out presubmits.
2. Run active and archived samples in presubmit only when there are changes in samples directory.

There could be some other better strategy but having this would help us find any regressions in archived samples.

harshachinta (approved):

surbhigarg92:
> LGTM, but can we make sure we have a plan to run these archived samples in some way? We could otherwise not detect issues with those samples. Few thoughts on doing this,
> 
> 1. Run the archived samples in continuous and nightly builds thus leaving out presubmits.
> 2. Run active and archived samples in presubmit only when there are changes in samples directory.
> 
> There could be some other better strategy but having this would help us find any regressions in archived samples.

Added the change to make sure we are running archived tests  in continuous and nightly builds

PR #2020, opened 2024-03-12
feat(spanner): add support for float32

alkatrivedi:
This PR contains support for FLOAT32 type on Cloud Spanner.


src/codec.ts
```diff
@@ -531,6 +550,7 @@ const TypeCode: {
   bool: 'BOOL',
   int64: 'INT64',
   pgOid: 'INT64',
+  float32: 'FLOAT32',
```
surbhigarg92, line 553: Should also add documentation in "paramtype", refer https://github.com/googleapis/nodejs-spanner/blob/main/src/codec.ts#L570

system-test/spanner.ts
```diff
@@ -4863,6 +4863,235 @@ describe('Spanner', () => {
           });
         });
 
+        describe('float32', () => {
+          const float32Query = (done, database, query, value) => {
```
surbhigarg92, line 4867: All the testcase are using float64 not float32. If param type is not passed then it take float64 as default. Add below for all tests
```
types: {
                v: 'float32',
              },
```

src/codec.ts
alkatrivedi: Done

rahul2393 (approved):

system-test/spanner.ts
```diff
@@ -4300,9 +4456,10 @@ describe('Spanner', () => {
       insertThenUpdateRow(done, postgreSqlTable);
     });
 
-    describe('insert & query', () => {
+    describe.skip('insert & query', () => {
```
surbhigarg92, line 4459: Do not skip this, this is also used for other types. Instead comment the part where we are adding Float32 and add a comment for later

system-test/spanner.ts
alkatrivedi: Done

system-test/spanner.ts
alkatrivedi: Done

surbhigarg92 (approved):

PR #2028, opened 2024-04-04
feat: optimisticLock option for getTransaction method

### LINKED ISSUE #2021: database.getTransaction() not accepting RunTransactionOptions
[RunTransactionOptions](https://github.com/googleapis/nodejs-spanner/blob/fc99d8eec14dc568d94cac6362c362be56c71ebe/src/transaction-runner.ts#L44C18-L44C39) has a property optimisticLock mode where a lock is acquired during a commit. More details [here](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.Spanner.V1/latest/Google.Cloud.Spanner.V1.TransactionOptions.Types.ReadWrite.Types.ReadLockMode).

This option is missing in [database.getTransaction()](https://github.com/googleapis/nodejs-spanner/blob/fc99d8eec14dc568d94cac6362c362be56c71ebe/src/database.ts#L1992).

alkatrivedi:
fixes: #2021


samples/insert-query-with-get-trasanction.js
```diff
@@ -0,0 +1,75 @@
...
+                    (4, 'Dylan', 'Shaw')`,
+      });
+      console.log(
+        `Successfully inserted ${rowCount} records into the Singers table`
+      );
+    } catch (err) {
+      console.error('ERROR:', err);
+    }
+
+    await transaction.commit();
```
olavloite, line 63: It seems strange to me that the sample commits the transaction regardless whether the insert statement succeeded or not. Put another way: Would it not make more sense to put the `commit()` call inside the try block? And then also add a `finally` block and put the `transaction.end()` call there.

samples/insert-query-with-get-trasanction.js
```diff
@@ -0,0 +1,75 @@
...
+  async function writeTransaction() {
+    // Gets a reference to a Cloud Spanner instance and database
+    const instance = spanner.instance(instanceId);
+    const database = instance.database(databaseId);
+
+    const options = {
+      optimisticLock: true,
+    };
+
+    const promise = await database.getTransaction(options).then();
```
olavloite, line 45: The use of `getTransaction(..)` means that there is no automatic retrying of the transaction if it is aborted by Spanner. Transaction aborts can happen at any time for any read/write transaction (not only because of lock conflicts, but also for more random, transient problems). That again means that this sample function will fail at random for a (very small) percentage of the times that it is called. Would it not make more sense to make this sample in such a way that it shows that the user is responsible for retrying transactions when they use `getTransaction(..)`?

samples/system-test/spanner.test.js
```diff
@@ -825,6 +825,25 @@ describe('Autogenerated Admin Clients', () => {
     );
   });
 
+  // dml_standard_insert_transaction_using_getTransaction
+  it('should insert rows into an example table using getTransaction method', async () => {
+    const output = execSync(
+      `node insert-query-with-get-trasanction.js ${INSTANCE_ID} ${DATABASE_ID} ${PROJECT_ID}`
```
olavloite, line 831: This can fail at random some times if the transaction is aborted by Spanner. (see also my comment above) That will make this test case a (bit) flaky.

src/database.ts
```diff
@@ -224,6 +224,14 @@ export interface GetIamPolicyOptions {
   gaxOptions?: CallOptions;
 }
 
+/**
+ * @typedef {object} GetTransactionOptions
+ * * @property {boolean} [optimisticLock] The optimistic lock a
+ *     {@link Transaction} should use while running.
+ */
+interface GetTransactionOptions extends RunTransactionOptions {
+  optimisticLock?: boolean;
```
olavloite, line 233: Why do we not want this as a field in `RunTransactionOptions`? Shouldn't it be possible to use this with the normal `runTransaction` method as well?

src/database.ts
```diff
@@ -1989,12 +1997,27 @@ class Database extends common.GrpcServiceObject {
    * });
    * ```
    */
-  getTransaction(): Promise<[Transaction]>;
```
olavloite, line 1992: Isn't removing this overload of this function a breaking change?

src/database.ts
```diff
@@ -1989,12 +1997,27 @@ class Database extends common.GrpcServiceObject {
...
-  getTransaction(callback: GetTransactionCallback): void;
+  getTransaction(callback: GetTransactionCallback): Promise<[Transaction]>;
```
olavloite, line 2000: (I'm not super familiar with how `promisify` works, so it could be that I'm overly worried here) This overload looks a bit strange. It takes both a callback as an argument, __and__ returns a promise. Normally it is either accepting a callback __or__ returning a promise.

src/database.ts
```diff
@@ -1989,12 +1997,27 @@ class Database extends common.GrpcServiceObject {
...
+      typeof optionsOrRunFn === 'function'
+        ? (optionsOrRunFn as RunTransactionCallback)
+        : fn;
+    const options =
+      typeof optionsOrRunFn === 'object' && optionsOrRunFn
+        ? (optionsOrRunFn as RunTransactionOptions)
+        : {};
     this.pool_.getSession((err, session, transaction) => {
+      if (options.optimisticLock) {
+        transaction!.useOptimisticLock();
```
olavloite, line 2019: Can we add a test that verifies that this option is actually sent to Spanner? From the PR description, this seems to be the most important part of the PR.

samples/insert-query-with-get-trasanction.js
```diff
@@ -0,0 +1,79 @@
+// Copyright 2024 Google LLC
```
surbhigarg92, line 1: Why are we adding samples here. Samples normally need to be discussed with documentation team and tags needs to be discussed with them. 
We recommend customers to use `runTransaction` method.

samples/insert-query-with-get-trasanction.js
alkatrivedi: Done

src/database.ts
alkatrivedi: Done

samples/insert-query-with-get-trasanction.js
alkatrivedi: I have removed the samples, as discussed

src/database.ts
alkatrivedi: This isn't a breaking change here since we have modified the signature to: 
```
getTransaction(
    optionsOrCallback?: GetTransactionOptions
  ): Promise<[Transaction]>;
```

where optionsOrCallback is an optional parameter which we may skip to pass while calling getTransaction()

src/database.ts

samples/system-test/spanner.test.js
alkatrivedi: Removed these tests since we are not adding a sample here

src/database.ts

samples/insert-query-with-get-trasanction.js
alkatrivedi: I have removed the sample, since we are recommending clients to use the runTransaction method also adding samples to use getTransaction method will need discussions with the documentation team for the tags

src/database.ts
```diff
@@ -224,6 +224,14 @@ export interface GetIamPolicyOptions {
   gaxOptions?: CallOptions;
 }
 
+/**
+ * @typedef {object} GetTransactionOptions
+ * * @property {boolean} [optimisticLock] The optimistic lock a
+ *     {@link Transaction} should use while running.
+ */
+
```
olavloite, line 232: nit: remove empty line

src/database.ts
olavloite: ack

system-test/spanner.ts
```diff
@@ -8894,6 +8894,24 @@ describe('Spanner', () => {
         mismatchedColumnError(done, DATABASE, googleSqlTable);
       });
 
+      it('GOOGLE_STANDARD_SQL should use getTransaction for executing sql', async () => {
+        const promise = await DATABASE.getTransaction({optimisticLock: true});
+        const transaction = promise[0];
```
olavloite, line 8899: nit:
```suggestion
        const transaction = (await DATABASE.getTransaction({optimisticLock: true}))[0];
```

(The name `promise` seems to indicate that the returned type is a `Promise`, but it is an array.)

system-test/spanner.ts
```diff
@@ -8894,6 +8894,24 @@ describe('Spanner', () => {
...
+
+        try {
+          const [rows] = await transaction!.run('SELECT * FROM TxnTable');
+          assert.strictEqual(rows.length, googleSqlRecords.length);
+        } catch (err) {
+          if ((err as grpc.ServiceError).code === grpc.status.ABORTED) {
+            assert.ok(err, 'Transaction is aborted');
+          } else {
+            console.log('ERROR: ', err);
```
olavloite, line 8908: Shouldn't this also fail the test? I don't think we expect any errors here.

system-test/spanner.ts
```diff
@@ -8894,6 +8894,24 @@ describe('Spanner', () => {
...
+      it('GOOGLE_STANDARD_SQL should use getTransaction for executing sql', async () => {
+        const promise = await DATABASE.getTransaction({optimisticLock: true});
+        const transaction = promise[0];
+
+        try {
+          const [rows] = await transaction!.run('SELECT * FROM TxnTable');
+          assert.strictEqual(rows.length, googleSqlRecords.length);
+        } catch (err) {
+          if ((err as grpc.ServiceError).code === grpc.status.ABORTED) {
+            assert.ok(err, 'Transaction is aborted');
```
olavloite, line 8906: Pure theoretically, this could happen for other, valid reasons as well. Although optimistic locking means that Spanner won't take any exclusive locks during the query, it does not mean that Spanner will never abort the transaction for other reasons. That means that this test could still fail in a few cases.

Maybe we should add a comment here to clarify that a flaky failures on this line is acceptable, as long as the error message indicates that the error was not due to a lock conflict?

olavloite (approved):

system-test/spanner.ts
alkatrivedi: Updated

system-test/spanner.ts
alkatrivedi: Updated

system-test/spanner.ts
alkatrivedi: Added the comment

src/database.ts
alkatrivedi: Done

surbhigarg92 (approved):

PR #2065, opened 2024-06-20
feat: add support for blind writes

alkatrivedi:
This PR contains the support for Blind Writes API in Spanner.


src/database.ts
```diff
@@ -3330,6 +3333,89 @@ class Database extends common.GrpcServiceObject {
...
+   *
+   * try {
+   *  const [response] = await database.blindWrite(mutation, {});
+   *  console.log(response.commitTimestamp);
+   * } catch(err) {
+   *  console.log("Error: ", err);
+   * }
+   * ```
+   */
+  blindWrite(mutation: Mutation): Promise<CommitResponse>;
```
surbhigarg92, line 3385: use the same method name `writeAtLeastOnce`

src/database.ts
```diff
@@ -3330,6 +3333,89 @@ class Database extends common.GrpcServiceObject {
...
+  ): void | Promise<CommitResponse> {
+    const cb =
+      typeof optionsOrCallback === 'function'
+        ? (optionsOrCallback as CommitCallback)
+        : callback;
+    const options =
+      typeof optionsOrCallback === 'object' && optionsOrCallback
+        ? (optionsOrCallback as CallOptions)
+        : {};
+    if (mutation.proto().length === 0) {
```
surbhigarg92, line 3401: We don't need this condition

test/database.ts
```diff
@@ -762,6 +782,83 @@ describe('Database', () => {
...
+      try {
+        database.blindWrite(fakeMutation, (err, res) => {});
+      } catch (error) {
+        assert(error instanceof Error);
+      }
+      done();
+    });
+
+    it('should return error when passing null mutation', done => {
+      const fakeError = new Error('err');
```
surbhigarg92, line 835: Use the actual error returned from backend

test/database.ts
```diff
@@ -762,6 +782,83 @@ describe('Database', () => {
...
+        callback(fakeErr, null, null)
+      );
+
+      database.blindWrite(mutation, err => {
+        assert.deepStrictEqual(err, fakeErr);
+        done();
+      });
+    });
+
+    it('should return error when passing an empty mutation', done => {
```
surbhigarg92, line 823: Modify this test and check nothing happened .

src/database.ts
```diff
@@ -3330,6 +3333,89 @@ class Database extends common.GrpcServiceObject {
...
+   * Apply Blind Write of the Mutations
+   *
+   * blindWrite requests are not replay protected, meaning that it may apply mutations more
+   * than once, if the mutations are not idempotent, this may lead to a failure being
+   * reported when the mutation was applied once. Replays of non-idempotent mutations may
+   * have undesirable effects. For example, replays of an insert mutation may produce an
+   * already exists error. For this reason, most users of the library will prefer to use
+   * {@link runTransaction} instead.
+   *
+   * However, {@code blindWrite()} requires only a single RPC, whereas {@code runTransaction()}
```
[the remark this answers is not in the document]

src/database.ts
```diff
@@ -3330,6 +3333,89 @@ class Database extends common.GrpcServiceObject {
...
+   *  LastName: 'Richards',
+   *  });
+   * mutation.update('Singers', {
+   *  SingerId: '1',
+   *  FirstName: 'John',
+   *  LastName: 'Richards',
+   *  });
+   *
+   * try {
+   *  const [response] = await database.blindWrite(mutation, {});
```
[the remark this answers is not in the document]

src/database.ts
alkatrivedi: Done

test/database.ts
alkatrivedi: Updated

surbhigarg92 (approved):

src/database.ts
```diff
@@ -3346,6 +3349,91 @@ class Database extends common.GrpcServiceObject {
     return proxyStream as NodeJS.ReadableStream;
   }
 
+  /**
+   * Apply Blind Write of the Mutations
```
olavloite, line 3353:
```suggestion
   * Write mutations using a single RPC invocation without replay protection.
```

src/database.ts
```diff
@@ -3346,6 +3349,91 @@ class Database extends common.GrpcServiceObject {
...
+   *
+   * writeAtLeastOnce(Blind Write) requests are not replay protected, meaning that it may apply mutations more
```
olavloite, line 3355:
```suggestion
   * writeAtLeastOnce writes mutations to Spanner using a single Commit RPC.
   * These requests are not replay protected, meaning that it may apply mutations more
```

src/database.ts
```diff
@@ -3346,6 +3349,91 @@ class Database extends common.GrpcServiceObject {
...
+   * than once, if the mutations are not idempotent, this may lead to a failure being
+   * reported when the mutation was applied once. Replays of non-idempotent mutations may
```
olavloite, line 3357:
```suggestion
   * than once. Replays of non-idempotent mutations may
```

src/database.ts
```diff
@@ -3346,6 +3349,91 @@ class Database extends common.GrpcServiceObject {
...
+   * mutation.insert('Singers', {
+   *  SingerId: '1',
+   *  FirstName: 'Marc',
+   *  LastName: 'Richards',
+   *  });
+   * mutation.update('Singers', {
+   *  SingerId: '1',
+   *  FirstName: 'John',
+   *  LastName: 'Richards',
+   *  });
```
olavloite, line 3391: (Note: It could be that this is copy-pasted from somewhere. If so, then we should fix it there as well, instead of repeating it here.)

This is an example of mutations that are not very-well suited for blind writes, as they include an `insert` mutation. The example is also a bit strange, as it first inserts and then updates the same row.
Can we instead change this example to one that executes two `upsert` mutations for two different Singer rows? Upsert mutations is a much more typical scenario for blind writes than this.

src/database.ts
```diff
@@ -3346,6 +3349,91 @@ class Database extends common.GrpcServiceObject {
...
+        : callback;
+    const options =
+      typeof optionsOrCallback === 'object' && optionsOrCallback
+        ? (optionsOrCallback as CallOptions)
+        : {};
+    this.pool_.getSession((err, session?, transaction?) => {
+      if (err && isSessionNotFoundError(err as grpc.ServiceError)) {
+        cb
+          ? this.writeAtLeastOnce(mutation, cb)
+          : this.writeAtLeastOnce(mutation, options);
```
olavloite, line 3424: Can't this method also be called with __both__ options and a callback? It seems that in that case we are ignoring the options.

src/transaction.ts
```diff
@@ -2530,6 +2541,110 @@ function buildDeleteMutation(
...
+ * try {
+ *  database.writeAtLeastOnce(mutation, (err, res) => {
+ *    console.log("RESPONSE: ", res);
+ *  });
+ * } catch(err) {
+ *  console.log("ERROR: ", err);
+ * }
+ * ```
+ */
+export class Mutation {
```
olavloite, line 2577: I find the name of this class a bit confusing. This does not represent a single `Mutation`, but a (mutable) set of mutations. Maybe rename to `MutationSet`?

test/database.ts
```diff
@@ -40,7 +40,13 @@ import {protos} from '../src';
 import * as inst from '../src/instance';
 import RequestOptions = google.spanner.v1.RequestOptions;
 import EncryptionType = google.spanner.admin.database.v1.RestoreDatabaseEncryptionConfig.EncryptionType;
-import {BatchWriteOptions} from '../src/transaction';
+import {
```
olavloite, line 43: Can we add a test (potentially to the `spanner.ts` file using the mock server) that verifies that:
1. Using `writeAtLeastOnce` does not trigger a `BeginTransaction` RPC.
2. Using `writeAtLeastOnce` triggers a `Commit` RPC that uses a single-use read/write transaction.

src/database.ts
alkatrivedi: Done

src/database.ts
alkatrivedi: Done

src/database.ts
alkatrivedi: Done

src/transaction.ts
alkatrivedi: Updated

src/database.ts
alkatrivedi: Done

src/database.ts
alkatrivedi: Typescript not able to infer the correct types from the overload signature. Hence we need to make it clear to typescript that correct type is pass

olavloite: Yeah, that was not my point. My point was that the original method call could be something like this:

```
database.writeAtLeastOnce(myMutations, myOptions, myCallback);
```

That then leads to the line that we are talking about here calling the following:

```
        cb
          ? this.writeAtLeastOnce(mutations, cb)
          : this.writeAtLeastOnce(mutations, options);
```

And as __both__ options and a callback was passed in as an argument, the ternary will resolve to:

```
this.writeAtLeastOnce(mutations, cb)
```

Put another way:
1. If the original method is called with __both__ options and a callback.
2. And a `SessionNotFound` error occurs.
3. Then the actual method call that is being executed during the retry is __without__ options.

src/transaction.ts
```diff
@@ -2530,6 +2541,110 @@ function buildDeleteMutation(
   return mutation as spannerClient.spanner.v1.Mutation;
 }
 
+/**
+ * MutationSet represent a set of changes to be applied atomically to a cloud spanner
```
olavloite, line 2545: nit: 
```suggestion
 * MutationSet represent a set of changes to be applied atomically to a Cloud Spanner
```

src/transaction.ts
```diff
@@ -2530,6 +2541,110 @@ function buildDeleteMutation(
...
+ * const database = instance.database('my-database');
+ *
+ * const mutation = new Mutation();
+ * mutation.insert('Singers', {SingerId: '123', FirstName: 'David'});
+ * mutation.update('Singers', {SingerId: '123', FirstName: 'Marc'});
+ *
+ * try {
+ *  database.writeAtLeastOnce(mutation, (err, res) => {
+ *    console.log("RESPONSE: ", res);
+ *  });
```
olavloite, line 2571: This code example needs to be updated to use `MutationSet`.

test/spanner.ts
```diff
@@ -3260,6 +3261,48 @@ describe('Spanner with mock server', () => {
...
+        LastName: 'Terry',
+      });
+      mutations.upsert('Singers', {
+        SingerId: 2,
+        FirstName: 'Marc',
+      });
+      await database.writeAtLeastOnce(mutations, {});
+      await database.close();
+
+      // Verify that we don't have a BeginTransaction request for the retry.
```
olavloite, line 3279:
```suggestion
      // Verify that we don't have a BeginTransaction request for the transaction.
```

test/spanner.ts
```diff
@@ -3260,6 +3261,48 @@ describe('Spanner with mock server', () => {
...
+      const beginTxnRequest = spannerMock.getRequests().find(val => {
+        return (val as v1.BeginTransactionRequest).options?.readWrite;
+      }) as v1.BeginTransactionRequest;
+      assert.deepStrictEqual(beginTxnRequest, undefined);
+
+      // Verify that we have a single Commit request, and that the Commit request contains only one mutation.
```
olavloite, line 3285:
```suggestion
      // Verify that we have a single Commit request, and that the Commit request contains two mutations and uses a single-use read/write transaction.
```

test/spanner.ts
```diff
@@ -3260,6 +3261,48 @@ describe('Spanner with mock server', () => {
...
+        }).length
+      );
+      const commitRequest = spannerMock.getRequests().find(val => {
+        return (val as v1.CommitRequest).mutations;
+      }) as v1.CommitRequest;
+      assert.ok(commitRequest, 'Commit was called');
+      assert.strictEqual(commitRequest.mutations.length, 2);
+      const request = spannerMock.getRequests().find(val => {
+        return (val as v1.CommitRequest).singleUseTransaction?.readWrite;
+      }) as v1.CommitRequest;
```
olavloite, line 3299: You can combine the two `spannerMock.getRequests()` into one and filter for __both__ `mutations` and `singleUseTransaction?.readWrite`. That simplifies the code a bit, but it also ensures that there is really just one commit request. (In theory, the current verification also covers the strange case where we have two commit requests; one that contains the mutations, but without a single-use read/write transaction, and one that contains no mutations, but has a single-use read/write transaction.)

olavloite (approved):
LGTM, with a couple of minor nits.

test/spanner.ts
alkatrivedi: Updated.

test/spanner.ts
alkatrivedi: Done

test/spanner.ts
alkatrivedi: Done

src/transaction.ts
alkatrivedi: Done

src/transaction.ts
alkatrivedi: Done

src/database.ts
```diff
@@ -3346,6 +3349,91 @@ class Database extends common.GrpcServiceObject {
...
+   * const mutations = new MutationSet();
+   * mutations.upsert('Singers', {
+   *    SingerId: 1,
+   *    FirstName: 'Scarlet',
+   *    LastName: 'Terry',
+   *  });
+   * mutations.upsert('Singers', {
+   *    SingerId: 2,
+   *    FirstName: 'Marc',
+   *  });
```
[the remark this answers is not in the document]
alkatrivedi: Done

src/database.ts
```diff
@@ -3346,6 +3349,91 @@ class Database extends common.GrpcServiceObject {
...
+  /**
+   * Write mutations using a single RPC invocation without replay protection.
+   *
+   * writeAtLeastOnce writes mutations to Spanner using a single Commit RPC.
+   * These requests are not replay protected, meaning that it may apply mutations more
+   * than once, if the mutations are not idempotent, this may lead to a failure being
+   * reported when the mutation was applied once. Replays non-idempotent mutations may
+   * have undesirable effects. For example, replays of an insert mutation may produce an
+   * already exists error. For this reason, most users of the library will prefer to use
+   * {@link runTransaction} instead.
```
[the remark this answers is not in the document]
alkatrivedi: Done

src/database.ts
alkatrivedi: Done

PR #2078, opened 2024-07-09
chore: callback method in getDatabaseDialect

### LINKED ISSUE #2075: Database.getDatabaseDialect's "cb" argument is never used
If we examine the method "getDatabaseDialect" https://github.com/googleapis/nodejs-spanner/blob/30151cad2723d6a9f2c70342c81f1a578ee02f6a/src/database.ts#L1530-L1553
we can see that the callback passed in is never used. I found this while working on trace instrumentation for these libraries.

alkatrivedi:
fixes: #2075

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -1533,23 +1533,39 @@ class Database extends common.GrpcServiceObject {
   ): void;
   async getDatabaseDialect(
     optionsOrCallback?: CallOptions | GetDatabaseDialectCallback,
-    cb?: GetDatabaseDialectCallback
+    callback?: GetDatabaseDialectCallback
   ): Promise<
     | EnumKey<typeof databaseAdmin.spanner.admin.database.v1.DatabaseDialect>
     | undefined
   > {
     const gaxOptions =
-      typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+      typeof optionsOrCallback === 'object'
+        ? (optionsOrCallback as CallOptions)
+        : {};
 
-    if (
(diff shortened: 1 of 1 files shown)


src/database.ts
```diff
@@ -1549,6 +1556,10 @@ class Database extends common.GrpcServiceObject {
       const [metadata] = await this.getMetadata(gaxOptions);
       this.databaseDialect = metadata.databaseDialect;
     }
+    if (cb) {
+      cb!(null, this.databaseDialect);
```
odeke-em, line 1560: At this point do we still need `cb!` given that we confirmed above that the callback is non-nil?

alkatrivedi: Yeah, we don't need. Updated

odeke-em (approved):
Thank you @alkatrivedi!

alkatrivedi:
Thanks for approving the PR @odeke-em

surbhigarg92:
@alkatrivedi Please add a unit test case .

src/database.ts
```diff
@@ -1549,6 +1556,10 @@ class Database extends common.GrpcServiceObject {
       const [metadata] = await this.getMetadata(gaxOptions);
       this.databaseDialect = metadata.databaseDialect;
     }
```
surbhigarg92, line 1558:
```suggestion
    }
    
```

alkatrivedi:
> @alkatrivedi Please add a unit test case .

A unit test already exists for this

surbhigarg92 (approved):

PR #2083, opened 2024-07-24
feat(spanner): add samples for instance partitions

varuncnaik:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-spanner/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #<issue_number_goes_here> 🦕


samples/instance-partition-create.js
```diff
@@ -0,0 +1,84 @@
...
+    // [START spanner_create_instance_partition]
+    // Imports the Google Cloud client library
+    const {Spanner} = require('@google-cloud/spanner');
+
+    /**
+     * TODO(developer): Uncomment the following lines before running the sample.
+     */
+    // const projectId = 'my-project-id';
+    // const instanceId = 'my-instance';
+    // const instancePartitionId = 'my-instance-partition';
```
alkatrivedi, line 37: nit: can we change my-instance-partition to an actual example of a instance partition id. That will make it easier for users to understand how they can use it.
Something like this maybe -  https://github.com/alkatrivedi/nodejs-spanner/blob/main/samples/database-create-with-default-leader.js#L30

surbhigarg92 (approved):

samples/instance-partition-create.js
varuncnaik: Actually, "my-instance-partition" is a valid instance partition id. The requirements for instance partition id are very similar to the requirements for instance id.

alkatrivedi: Got it!

alkatrivedi (approved):

PR #2091, opened 2024-08-07
fix: unhandled exception error catch

### LINKED ISSUE #1972: Uncaught exception with concurrent queries and inline begin
The client throws an unhandled exception if the initial queries in the transaction are called concurrently and one of the queries fails(except for the first one). You can see the example below. The issue is resolved if `transaction.begin()` is called explicitly before the concurrent queries. 

#### Environment details

  - OS: macOS 14.2 (23C64)
  - Node.js version: v20.1.0
  - npm version: 9.6.4
  - `@google-cloud/spanner` version: reproducible on master 

#### Steps to reproduce

Minimal repro

```ts
// We discovered this issue with ABORTED errors, but
// it's easier to reproduce with this non-existent table query.
const nonExistantTableQuery = 'SELECT * FROM nonExistingTable';
  try {
    await Promise.all([
      tx!.run(selectSql),
      tx!.run(nonExistantTableQuery),
    ]);
    await tx.commit();
  } catch (err) {
    await tx.rollback();
  }
});

```

This code will result in an uncaught exception and it will crash node process.

I think the issue is a missing error handler in this code. 

https://github.com/googleapis/nodejs-spanner/blob/6c0eef16d85e6da4993b246337483a818a4521b5/src/transaction.ts#L1336-L1338

Adding this handler fixes the issue
```TS
.on('error', err => streamProxy.emit('error', err))
```

Happy to open PR with test if this seems correct

alkatrivedi:
fixes: #1972

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -712,8 +712,16 @@ export class Snapshot extends EventEmitter {
           this._update(response.metadata!.transaction);
         }
       })
-      .on('error', () => {
-        if (!this.id && this._useInRunner) {
+      .on('error', err => {
+        const isServiceError = err && typeof err === 'object' && 'code' in err;
+        if (
+          !this.id &&
+          this._useInRunner &&
+          !(
+            isServiceError &&
+            (err as grpc.ServiceError).code === grpc.status.ABORTED
+          )
+        ) {
           this.begin();
         }
       });
@@ -1219,8 +1227,16 @@ export class Snapshot extends EventEmitter {
           this._update(response.metadata!.transaction);
         }
       })
-      .on('error', () => {
-        if (!this.id && this._useInRunner) {
+      .on('error', err => {
+        const isServiceError = err && typeof err === 'object' && 'code' in err;
+        if (
+          !this.id &&
+          this._useInRunner &&
+          !(
+            isServiceError &&
+            (err as grpc.ServiceError).code === grpc.status.ABORTED
+          )
+        ) {
           this.begin();
         }
       });
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -1437,6 +1437,7 @@ export class Snapshot extends EventEmitter {
       this._waitingRequests.push(() => {
         makeRequest(resumeToken)
           .on('data', chunk => streamProxy.emit('data', chunk))
+          .on('error', err => streamProxy.emit('error', err))
           .on('end', () => streamProxy.emit('end'));
       });
 
--- a/test/spanner.ts
+++ b/test/spanner.ts
@@ -2978,35 +2978,24 @@ describe('Spanner with mock server', () => {
         .filter(val => (val as v1.ExecuteSqlRequest).sql)
         .map(req => req as v1.ExecuteSqlRequest);
 
-      assert.strictEqual(requests.length, 6);
-      assert.ok(
-        requests[0].transaction?.begin!.readWrite,
-        'inline txn is not set.'
-      );
-      assert.ok(
-        requests[1].transaction?.begin!.readWrite,
-        'inline txn is not set.'
-      );
-      assert.ok(
-        requests[2].transaction?.begin!.readWrite,
-        'inline txn is not set.'
-      );
-      assert.ok(
-        requests[3].transaction!.id,
-        'Transaction ID is not used for retries.'
-      );
-      assert.ok(
-        requests[4].transaction!.id,
-        'Transaction ID is not used for retries.'
-      );
-      assert.ok(
-        requests[5].transaction!.id,
-        'Transaction ID is not used for retries.'
-      );
+      assert.strictEqual(requests.length, 4);
+      requests.slice(0, 1).forEach((request, index) => {
+        assert.ok(
+          request.transaction?.begin!.readWrite,
+          `Inline txn is not set in request ${index}.`
+        );
+      });
+      requests.slice(1, 4).forEach((request, index) => {
+        assert.ok(
+          request.transaction!.id,
+          `Transaction ID is not used for retries. ${index}.`
+        );
+      });
       const beginTxnRequest = spannerMock
         .getRequests()
         .filter(val => (val as v1.BeginTransactionRequest).options?.readWrite)
         .map(req => req as v1.ExecuteSqlRequest);
+      // TODO: There is a bug which is initiating BeginTransaction RPC twice.
       assert.deepStrictEqual(beginTxnRequest.length, 2);
       const commitRequests = spannerMock
         .getRequests()
--- a/test/spanner.ts
+++ b/test/spanner.ts
@@ -3224,6 +3224,56 @@ describe('Spanner with mock server', () => {
       });
     });
 
+    it('should catch an exception error even after retrying on aborted query', async () => {
+      let attempts = 0;
+      const database = newTestDatabase();
+
+      try {
+        await database.runTransactionAsync(async tx => {
+          if (!attempts) {
+            spannerMock.abortTransaction(tx);
+          }
+          attempts++;
+    
+          try {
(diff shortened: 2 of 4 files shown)


test/spanner.ts
```diff
@@ -3211,6 +3211,21 @@ describe('Spanner with mock server', () => {
...
+    it('should catch exception error during invalid queries while using inline begin transaction', async () => {
+      const database = newTestDatabase();
+      try {
+        await database.runTransactionAsync(async tx => {
+          await tx!.run(selectSql);
+          await tx!.run(invalidSql);
+          await tx.commit();
+        });
+      } catch (err) {
+        (err as grpc.ServiceError).message.includes('Table FOO not found');
```
surbhigarg92, line 3223: Need an assert here.

alkatrivedi: Done

surbhigarg92 (approved):

test/spanner.ts
```diff
@@ -1314,6 +1314,43 @@ describe('Spanner with mock server', () => {
...
+          assert.ok(
+            requests[1].transaction!.id,
+            'Transaction ID is not used for retries.'
+          );
+          assert.ok(
+            requests[1].resumeToken,
+            'Resume token is not set for the retried'
+          );
+        });
+
```
olavloite, line 1353: Can we add a test that verifies the behavior if the initial `ExecuteStreamingSql` call succeeds, but the stream then fails with `UNAVAILABLE` after receiving the first result? That should lead to the following behavior:
1. `ExecuteStreamingSql` is retried, but __without__ a `transaction.begin` and __with__ a `transaction.id` (and with a resume token)
2. There's also no extra `BeginTransaction` RPC

test/spanner.ts
```diff
@@ -2890,6 +2927,30 @@ describe('Spanner with mock server', () => {
...
+          ]).then(([rows]) => {
+            let count = 0;
+            rows.forEach(() => count++);
+            return transaction.commit().then(() => count);
+          });
+        }
+      );
+      assert.strictEqual(rowCount, 3);
+      assert.strictEqual(attempts, 2);
+      await database.close();
```
olavloite, line 2951: Can we add verifications for the requests that we receive here? My expectation would be the following list of requests:

1. `ExecuteStreamingSql` (with `transaction.begin`)
2. `BeginTransaction`
3. `ExecuteStreamingSql` with transaction.id
4. `ExecuteStreamingSql` with transaction.id
5. `Commit`

test/spanner.ts
```diff
@@ -2890,6 +2927,30 @@ describe('Spanner with mock server', () => {
...
+            transaction!.run(selectSql),
+            transaction!.run(selectSql),
+          ]).then(([rows]) => {
+            let count = 0;
+            rows.forEach(() => count++);
+            return transaction.commit().then(() => count);
+          });
+        }
+      );
+      assert.strictEqual(rowCount, 3);
```
olavloite, line 2949: Why is the row count 3? Shouldn't that be 2, as each query returns 1 row.

test/spanner.ts
olavloite: (It could be that we already have such a test, so check for that first)

test/spanner.ts
```diff
@@ -1314,6 +1314,43 @@ describe('Spanner with mock server', () => {
...
+            streamIndex: index,
+          } as MockError;
+          spannerMock.setExecutionTime(
+            spannerMock.executeStreamingSql,
+            SimulatedExecutionTime.ofError(err)
+          );
+          const database = newTestDatabase();
+
+          await database.runTransactionAsync(async tx => {
+            await Promise.all([tx!.run(selectSql), tx!.run(selectSql)]);
```
olavloite, line 1330: Can we verify here that we get the expected (number of) rows?

test/spanner.ts
```diff
@@ -2890,6 +2952,70 @@ describe('Spanner with mock server', () => {
...
+      );
+      assert.strictEqual(rowCount, 6);
+      assert.strictEqual(attempts, 2);
+      const requests = spannerMock
+        .getRequests()
+        .filter(val => (val as v1.ExecuteSqlRequest).sql)
+        .map(req => req as v1.ExecuteSqlRequest);
+
+      assert.strictEqual(requests.length, 6);
+      assert.ok(
```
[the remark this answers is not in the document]
alkatrivedi: Done

test/spanner.ts
```diff
@@ -1314,6 +1314,68 @@ describe('Spanner with mock server', () => {
...
+            requests[1].resumeToken,
+            'Resume token is not set for the retried'
+          );
+          const commitRequests = spannerMock
+            .getRequests()
+            .filter(val => (val as v1.CommitRequest).mutations)
+            .map(req => req as v1.CommitRequest);
+          assert.strictEqual(commitRequests.length, 1);
+          assert.deepStrictEqual(
+            requests[1].transaction!.id,
```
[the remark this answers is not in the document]
alkatrivedi: Updated

test/spanner.ts
alkatrivedi: Done

test/spanner.ts
alkatrivedi: Updated

src/transaction.ts
```diff
@@ -1219,8 +1227,16 @@ export class Snapshot extends EventEmitter {
           this._update(response.metadata!.transaction);
         }
       })
-      .on('error', () => {
-        if (!this.id && this._useInRunner) {
+      .on('error', err => {
```
olavloite, line 1230: For a different refactor PR: It seems that `runStream` and `createReadStream` have a lot of duplicated code. Could we combine more of those?

test/spanner.ts
```diff
@@ -1314,6 +1314,43 @@ describe('Spanner with mock server', () => {
...
+            'inline txn is not set.'
+          );
+          assert.ok(
+            requests[1].transaction!.id,
+            'Transaction ID is not used for retries.'
+          );
+          assert.ok(
+            requests[1].resumeToken,
+            'Resume token is not set for the retried'
+          );
```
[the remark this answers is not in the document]
olavloite: Ack

test/spanner.ts
olavloite: Ack

test/spanner.ts
```diff
@@ -3211,6 +3326,19 @@ describe('Spanner with mock server', () => {
       assert.ok(!beginTxnRequest, 'beginTransaction was called');
     });
 
+    it('should catch an exception error during invalid queries while using inline begin transaction', async () => {
+      const database = newTestDatabase();
+      await database.runTransactionAsync(async tx => {
+        try {
+          await Promise.all([tx!.run(selectSql), tx!.run(invalidSql)]);
```
olavloite, line 3333: Do we already have a test for the same with these requests in the other order? That is; with the invalid SQL first?

If not, can we add that as well, and verify that:
1. The initial call to ExecuteStreamingSql fails, which also means that there is no transaction ID that is returned.
2. The transaction is retried with an explicit BeginTransaction and then the invalid SQL + working SQL statement.
3. Catch the error from the invalid SQL statement.
4. Commit the transaction (even though one of the statements failed).
5. The result should be (I think):

```
ExecuteSql(transaction{begin: {readwrite=true}})
BeginTransaction(readwrite=true)
ExecuteSql(transaction{id:...})
ExecuteSql(transaction{id:...})
Commit
```

olavloite (approved):

src/transaction.ts
surbhigarg92: Sure.

test/spanner.ts
surbhigarg92: Will be doing this in a separate PR

PR #2087, opened 2024-08-01
DO NOT MERGE: feat: instrument src/ with traces using OpenTelemetry

odeke-em:
This change implements tracing of RPC calls using
OpenTelemetry to aid in providing observability.

Fixes #2079

Issue #2079, opened 2024-07-13
all: instrument with traces for observability


odeke-em:
Customers are asking for traces to be added to this package so that they can properly diagnose latency issues and have observability for their systems.

I am tasked with implementing tracing for this library and this issue serves as the reference. I have implemented this functionality in my fork https://github.com/orijtech/nodejs-spanner/pull/3 and I shall progressively send code changes to this repository to make the software development process much more scalable and easier to merge.

## Samples of results
<img width="1410" alt="Screenshot 2024-07-14 at 1 52 25 AM" src="https://github.com/user-attachments/assets/76b12c30-42c5-4ae7-9269-ea736a781e5e">

with PII acknowledgement and annotations added
<img width="1410" alt="Screenshot 2024-07-14 at 1 52 47 AM" src="https://github.com/user-attachments/assets/fbed9d97-ebbd-4d08-872a-04e1c214771b">

odeke-em:
And completed!

PR #2097, opened 2024-08-17
observability: add base for instrumentation + tests

odeke-em:
This change adds the scaffolding for instrumentation along with tests for the
tracing mechanisms that'll be used to trace the entire package.

Built from #2087
Updates #2079


odeke-em:
Kindly /cc-ing @surbhigarg92

src/instrument.ts
```diff
@@ -0,0 +1,262 @@
...
+// have been already created.
+export function getTracer(config?: traceConfig) {
+  if (config && config.opts && config.opts.tracerProvider) {
+    return config.opts.tracerProvider.getTracer(TRACER_NAME);
+  }
+  if (defaultTracerProvider) {
+    return defaultTracerProvider.getTracer(TRACER_NAME);
+  }
+  // Otherwise use the global tracer.
+  return trace.getTracer(TRACER_NAME);
```
surbhigarg92, line 94: Does this mean that users will need to register global tracer with 'nodejs-spanner' tracer_name ?

src/instrument.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+}
+
+let defaultTracerProvider: typeof TracerProvider = undefined;
+
+// setTracerProvider allows the caller to hook up an OpenTelemetry
+// TracerProvider that spans generated from this library shall be attached to,
+// instead of using the global configuration. Later on if `getTracer` is invoked and
+// the default tracerProvider is unset, it'll use the global tracer
+// otherwise it'll use the set TracerProvider.
```
surbhigarg92, line 76:
```suggestion
// setTracerProvider allows the caller to specify a custom OpenTelemetry
// TracerProvider to be used for Spanner Traces. Global TracerProvider will be used
// as a fallback . 
```

src/instrument.ts
```diff
@@ -0,0 +1,262 @@
...
+export function setTracerProvider(freshTracerProvider: typeof TracerProvider) {
```
surbhigarg92, line 77: Why do we need multiple options to set `TracerProvider`.

I think we should keep only one option `applyObservabilityOptions` using `observabilityOptions` interface.

Same for `optInForSQLStatementOnSpans` and `optOutOfSQLStatementOnSpans`. We don't need these.

Too many options will add confusion for users.

Also IMU these `observabilityOptions` can be set via SpannerOptions , we shouldn't even need to export `applyObservabilityOptions` . This should be set within Spanner class context.

src/instrument.ts
```diff
@@ -0,0 +1,262 @@
...
+  defaultTracerProvider = freshTracerProvider;
+}
+
+const TRACER_NAME = 'nodejs-spanner';
+
+// getTracer fetches the tracer each time that it is invoked to solve
+// the problem of observability being initialized after Spanner objects
+// have been already created.
+export function getTracer(config?: traceConfig) {
```
surbhigarg92, line 86: If observabilityOptions will be injected via SpannerOptions, in that case won't all the classes will have access to these options, in what scenario this will be used ?

src/instrument.ts
```diff
@@ -0,0 +1,262 @@
...
+export function optOutOfSQLStatementOnSpans() {
+  optedInPII = false;
+}
+
+interface traceConfig {
+  sql?: string | SQLStatement;
+  tableName?: string;
+  dbName?: string;
+  opts?: observabilityOptions;
+  enableExtendedTracing?: boolean;
```
surbhigarg92, line 114: Why do we need it again here ? We are already passing `observabilityOptions`

src/instrument.ts
```diff
@@ -0,0 +1,262 @@
...
+    return span;
+  }
+  return new noopSpan();
+}
+
+// noopSpan is a pass-through Span that does nothing and shall not
+// be exported, nor added into any context. It serves as a placeholder
+// to allow calls in sensitive areas like sessionPools to transparently
+// add attributes to spans without lots of ugly null checks.
+class noopSpan implements Span {
```
surbhigarg92, line 220: At least Java opentelmetry already has a Noop span implementation , dont we have something similar for Node ?

https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk_trace_base.NoopSpanProcessor.html

src/instrument.ts
surbhigarg92: Use observabilityOptions instead of traceConfig

src/instrument.ts
odeke-em: I finally found it under "NonRecordingSpan"

src/instrument.ts
odeke-em: No longer exists but thank you for the change suggestion!

src/instrument.ts
odeke-em: I just passed in TracerProvider directly. Thank you!

src/instrument.ts
odeke-em: Fixed and thank you @surbhigarg92!

src/instrument.ts
odeke-em: So following up with our discussion about `(cloud.google.com/nodejs/spanner, VERSION)` to retrieve the tracer.

odeke-em:
@surbhigarg92 @harshachinta the PR is ready to fly, please take a look and thank you!

observability-test/observability.test.ts
```diff
@@ -0,0 +1,417 @@
...
+
+const {disableContextAndManager, setGlobalContextManager} = require('./helper');
+
+const {
+  AsyncHooksContextManager,
+} = require('@opentelemetry/context-async-hooks');
+
+const projectId = process.env.SPANNER_TEST_PROJECTID || 'test-project';
+
+describe('Test startTrace', () => {
```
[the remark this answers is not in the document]
odeke-em: Hey @surbhigarg92, I purposefully setup this directory because running all the tests locally is a huge chore, given that some of them require mutating a live Cloud Spanner instance, and they are unrelated to observability, which makes development so much harder. It also helps to have an accurate view of coverage.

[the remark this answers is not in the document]
odeke-em: Also as you can see, testing for observability is a very involved process for which it is so much easier when its tests are isolated and can be focused on to test rigorously than when mixed with the other functional tests that are related to Cloud Spanner correctness/functionality.

src/instrument.ts
```diff
@@ -0,0 +1,219 @@
...
+}
+
+export type {observabilityOptions as ObservabilityOptions};
+
+const TRACER_NAME = 'cloud.google.com/nodejs/spanner';
+const TRACER_VERSION = '7.14.0'; // Manually hard coded, TODO: remove
+
+// getTracer fetches the tracer each time that it is invoked to solve
+// the problem of observability being initialized after Spanner objects
+// have been already created.
```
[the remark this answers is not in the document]
odeke-em: Thank you, done!

src/instrument.ts
```diff
@@ -0,0 +1,219 @@
...
+  return trace.getTracer(TRACER_NAME, TRACER_VERSION);
+}
+
+interface traceConfig {
+  sql?: string | SQLStatement;
+  tableName?: string;
+  dbName?: string;
+  opts?: observabilityOptions;
+  that?: Object;
+  tracerProvider?: TracerProvider;
```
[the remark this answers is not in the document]
odeke-em: Removed, thank you!

src/instrument.ts
```diff
@@ -0,0 +1,219 @@
...
+      }
+      if (config.dbName) {
+        span.setAttribute(SEMATTRS_DB_NAME, config.dbName);
+      }
+
+      const definedExtendedTracing =
+        config.opts && config.opts.enableExtendedTracing !== undefined;
+      // If they optedInPII but opts.enableExtendedTracing=false, reject it.
+      const explicitlySkipET =
+        definedExtendedTracing && !config.opts?.enableExtendedTracing;
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

src/instrument.ts
```diff
@@ -0,0 +1,219 @@
...
+}
+
+const SPAN_NAMESPACE_PREFIX = 'cloud.google.com/nodejs/spanner';
+export {SPAN_NAMESPACE_PREFIX};
+
+export function startTrace<F extends (span: Span) => ReturnType<F>>(
+  spanNameSuffix: string,
+  config: traceConfig | undefined,
+  cb: F
+): unknown {
```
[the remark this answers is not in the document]
odeke-em: Thank you for the simplification, done!

observability-test/observability.test.ts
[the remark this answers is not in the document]
odeke-em: I updated it slightly please take a look and thanks!

observability-test/observability.test.ts
```diff
@@ -0,0 +1,417 @@
...
+
+      // The span MUST start with the prefix of this library.
+      const gotName = span0.name;
+      const wantName = SPAN_NAMESPACE_PREFIX + '/mySpan';
+      assert.equal(
+        gotName,
+        wantName,
+        'name mismatch:\n\tGot: ' + gotName + '\n\tWant: ' + wantName
+      );
+    });
```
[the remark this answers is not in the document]
odeke-em: Updated, thanks!

observability-test/observability.test.ts
```diff
@@ -0,0 +1,417 @@
...
+  });
+
+  it('startTrace with TracerProvider in options, skips using global TracerProvider', () => {
+    const exporter2 = new InMemorySpanExporter();
+    const provider2 = new NodeTracerProvider({
+      sampler: sampler,
+      exporter: exporter2,
+    });
+    provider2.addSpanProcessor(new SimpleSpanProcessor(exporter2));
+    provider2.register();
```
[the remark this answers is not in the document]
odeke-em: For while it is being used yes, but after that I unregister it.

observability-test/observability.test.ts
```diff
@@ -0,0 +1,417 @@
...
+
+    startTrace('aSpan', {tracerProvider: provider2}, async span => {
+      await new Promise((resolve, reject) => setTimeout(resolve, 400));
+      span.end();
+
+      const gotSpansFromGlobal = exporter.getFinishedSpans();
+      assert.ok(
+        gotSpansFromGlobal.length,
+        0,
+        'no spans should be registered to the global'
```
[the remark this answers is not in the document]
odeke-em: So the assertion log show the got vs expected count though so the whole message is redundant and repeats.

observability-test/observability.test.ts
```diff
@@ -0,0 +1,417 @@
...
+        span.duration(),
+        activeSpan.duration(),
+        'durations must now match'
+      );
+    });
+  });
+});
+
+describe('setError', () => {
+  const exporter = new InMemorySpanExporter();
```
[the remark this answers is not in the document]
odeke-em: Well so the use of "before" in majority of these cases unnecessarily adds typing declarations because the common code must be cleaned up in the after clause so that it won't pollute the other individual tests.

src/instrument.ts
surbhigarg92: What does this mean ? Can we use `NonRecordingSpan` directly instead of creating our `noopSpan` class ?

package.json
```diff
@@ -57,6 +59,10 @@
     "@google-cloud/projectify": "^4.0.0",
     "@google-cloud/promisify": "^4.0.0",
     "@grpc/proto-loader": "^0.7.0",
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/context-async-hooks": "^1.25.1",
+    "@opentelemetry/sdk-trace-node": "^1.25.1",
```
surbhigarg92, line 64: This dependency is not used for the actual code and only required for tests. Is there a way we can only add this dependency as a test dependency ?

observability-test/observability.ts
```diff
@@ -0,0 +1,450 @@
...
+  it('with TracerProvider in options, skips using global TracerProvider', () => {
+    const overridingExporter = new InMemorySpanExporter();
+    const overridingProvider = new NodeTracerProvider({
+      sampler: sampler,
+      exporter: overridingExporter,
+    });
+    overridingProvider.addSpanProcessor(
+      new SimpleSpanProcessor(overridingExporter)
+    );
+    overridingProvider.register();
```
surbhigarg92, line 101: If we are registering this here, how would we make sure if the tests are using the injected provider or the registered one ? 
Shouldn't we remove this line and just inject the tracerprovider ?

src/instrument.ts
odeke-em: Sorry I hadn't completed my response. So NonRecordingSpan isn't exported out of use of opentelemetry as a visible symbol sadly.

package.json
odeke-em: Done, and moved to "devDependencies", thank you!

observability-test/observability.ts
odeke-em: Roger that, I thought about it last evening too, done and thank you @surbhigarg92!

src/instrument.ts
surbhigarg92: Can you please share the link of NonRecordingSpan code, I just want to make sure we have used the same code for the same.

surbhigarg92 (approved):

src/instrument.ts
```diff
@@ -0,0 +1,249 @@
...
+      span.setAttribute(SEMATTRS_DB_SYSTEM, 'spanner');
+
+      if (config.tableName) {
+        span.setAttribute(SEMATTRS_DB_SQL_TABLE, config.tableName);
+      }
+      if (config.dbName) {
+        span.setAttribute(SEMATTRS_DB_NAME, config.dbName);
+      }
+
+      // If they optedInPII but opts.enableExtendedTracing=false, reject it.
```
surbhigarg92, line 113: @odeke-em There is a change which we discussed for env variables. 

ExtendedTracing should be enabled if either env variable is enabled or opts.enableExtendedTracing is enabled

src/instrument.ts
```diff
@@ -0,0 +1,249 @@
...
+
+interface traceConfig {
+  sql?: string | SQLStatement;
+  tableName?: string;
+  dbName?: string;
+  opts?: observabilityOptions;
+  that?: Object;
+}
+
+const SPAN_NAMESPACE_PREFIX = 'cloud.google.com/nodejs/spanner';
```
surbhigarg92, line 79: Why are we adding this prefix in all the traces ?

src/instrument.ts
```diff
@@ -0,0 +1,249 @@
...
+}
+
+export type {observabilityOptions as ObservabilityOptions};
+
+const TRACER_NAME = 'cloud.google.com/nodejs/spanner';
+const TRACER_VERSION = '7.14.0'; // Manually hard coded, TODO: remove
+
+/**
+ * getTracer fetches the tracer from the provided tracerProvider.
+ * @param {TracerProvider} [tracerProvider] -- optional custom tracer provider
```
alkatrivedi, line 58: nit: please remove `--` from the comments
you can check the documentation format here - https://github.com/googleapis/nodejs-spanner/blob/main/src/codec.ts#L495

src/instrument.ts
```diff
@@ -0,0 +1,249 @@
...
+const SPAN_NAMESPACE_PREFIX = 'cloud.google.com/nodejs/spanner';
+export {SPAN_NAMESPACE_PREFIX};
+
+/**
+ * startTrace begins an active span in the current active context
+ * and passes it back to the set callback function. Each span will
+ * be prefixed with "cloud.google.com/nodejs/spanner". It is the
+ * responsibility of the caller to invoke [span.end] when finished tracing.
+ *
+ * @returns {Span} -- The created span.
```
alkatrivedi, line 88: nit: please remove `--` from here as well.

src/instrument.ts
odeke-em: To correctly identify all the spans as originating from this library and namespaces correctly. For example all gRPC spans emanating from the gRPC instrumentation are prefixed with "grpc.google.spanner.v1" and for libraries with proper instrumentation like Go, every span is prefixed with the proper "cloud.google.com/go/spanner".

src/instrument.ts
odeke-em: Yes it is enabled for either, but this case handles when the environment variable is true but the library caller explicitly passes in `{opts:{enableExtendedTracing:false}}`

src/instrument.ts
odeke-em: Done, thanks @alkatrivedi!

src/instrument.ts
odeke-em: Done, thanks @alkatrivedi!

src/instrument.ts
surbhigarg92: That can be done via `Scope name` as we are adding this in instrumentation scope. 

@harshachinta `for libraries with proper instrumentation like Go, every span is prefixed with the proper "cloud.google.com/go/spanner".` Is it really happening ?

src/instrument.ts
surbhigarg92: No even in this case we should make `enableExtendedTracing ` as true. So consider customer has set  `{opts:{enableExtendedTracing:false}}` but during some issue, they dont want to redeploy their aplication just set the environment variable and restart . 

Sorry I did not inform this earlier, this is a discussion we had yesterday.

src/instrument.ts
harshachinta: Yes, and historically it could be because opencensus didn't have the scope name support and it was hard to identify which library emits a particular trace.
Whereas in opentelemetry we have a scope name that can be set. So lets **not add the full name as prefix** to every span in python and node since it could make it hard to read the spans.

odeke-em: @harshachinta @surbhigarg92 the reason for using the namespace is more to aid in searching for spans than the use of scope name: most tracing tools like Google Cloud Trace, Zipkin, Jaeger have a search by span name prefix. It'll be impossible to find spaces produced by a library without their namespace. The use now of "scope" will be a drastic change for tools and visualizers. I highly recommend we stick with historical precedent because even the gRPC OpenTelemetry instrumentation that's modernized uses full namespace.

odeke-em: I suppose we can reduce it to "CloudSpanner." (like java-spanner and python-spanner do) but that'll also mean that we couldn't easily find spans from a language or library and systems like Google Cloud Trace do not support searching by attributes but rather by search by name
Screenshot 2024-08-28 at 3 02 35 PM

odeke-em: @surbhigarg92 @harshachinta  I think that to avoid ruffling too many feathers, we can use `CloudSpanner.` as the prefix to match java-spanner and python-spanner, but we are definitely going to have to make a standardization decision and discussion later on across all Cloud Spanner libraries, or have to change go-spanner which uses the fully qualified names.

src/instrument.ts
odeke-em: @surbhigarg92 alright, thank you, I've simplified it even more.

PR #2101, opened 2024-08-30
observability: add SessionPool span annotations

odeke-em:
This change adds span annotations on the active span which could either be the noopSpan or the top-level active one, for whenever we progressively add more spans.

Built from #2087
Updates #2079


odeke-em:
## Exhibit
On running this code per instructions of sample https://gist.github.com/odeke-em/35fb0e58b3a73dca59730b688817be1b, we get back
Screenshot 2024-08-30 at 6 47 59 AM
Screenshot 2024-08-30 at 6 48 29 AM
Screenshot 2024-08-30 at 7 07 24 AM

src/session-pool.ts
```diff
@@ -1048,6 +1112,9 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @param {Session} session The session object.
    */
   _release(session: Session): void {
+    const span = getActiveOrNoopSpan();
+    span.addEvent('releasing session back to the pool', {id: session.id});
```
surbhigarg92, line 1116: Remove this span. Removed session is enough.

src/session-pool.ts
```diff
@@ -736,6 +755,8 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     while (amount > 0) {
       let sessions: Session[] | null = null;
 
+      span.addEvent(`trying to create ${amount} sessions`);
```
surbhigarg92, line 758: nit: rename to `Creating ${amount} sessions`

src/session-pool.ts
```diff
@@ -919,9 +963,11 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     // Only create a new session if there are more waiters than sessions already
     // being created. The current requester will be waiter number _numWaiters+1.
     if (!this.isFull && this.totalPending <= this.totalWaiters) {
+      span.addEvent('pool is not full and more waiters than sessions');
```
surbhigarg92, line 966: Remove this, "Creating session" should be enough

src/session-pool.ts
```diff
@@ -990,6 +1044,10 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @returns {Promise}
    */
   async _ping(session: Session): Promise<void> {
+    // NOTE: Please do not trace Ping as it gets quite spammy
```
surbhigarg92, line 1047: How are we making sure this does not happen. ?

src/session-pool.ts
```diff
@@ -1030,10 +1088,16 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @param {object} options The transaction options.
    */
   _prepareTransaction(session: Session): void {
+    const span = getActiveOrNoopSpan();
+    span.addEvent('creating transaction for session');
```
surbhigarg92, line 1092: this actually does not create a transaction. Its just creates a transaction object. transaction.id will always be null here. 

Remove both the events here

surbhigarg92: In database class we have `transaction.begin` calls, there we should add these events. 

We can simply add these in transaction.begin() method.

package.json
```diff
@@ -60,7 +60,6 @@
     "@google-cloud/promisify": "^4.0.0",
     "@grpc/proto-loader": "^0.7.0",
     "@opentelemetry/api": "^1.9.0",
-    "@opentelemetry/context-async-hooks": "^1.25.1",
```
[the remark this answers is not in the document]
surbhigarg92: What does this package do, does it add any extra process which runs within app ?

src/session-pool.ts
```diff
@@ -865,21 +892,35 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @returns {Promise<Session>}
    */
   async _getSession(startTime: number): Promise<Session> {
+    const span = getActiveOrNoopSpan();
+    span.addEvent('Attempting to get session');
+
     if (this._hasSessionUsableFor()) {
-      return this._borrowNextAvailableSession();
+      span.addEvent('Cache hit: has usable session');
+      const sessPromise = this._borrowNextAvailableSession();
```
surbhigarg92, line 900: why this change is required ?  to save results in const 

directly use `return this._borrowNextAvailableSession();`

test/session-pool.ts
```diff
@@ -1349,6 +1365,84 @@ describe('SessionPool', () => {
...
+          spans.length,
+          1,
+          'Exactly 1 span should have been exported'
+        );
+        assert.strictEqual(
+          spans[0].name,
+          `${SPAN_NAMESPACE_PREFIX}.${topLevelSpanName}`,
+          'Expected only the top-level created span'
+        );
+        const span0 = spans[0];
```
surbhigarg92, line 1426: I don't see any assert on the annotations. 

Ideally we should add below tests 

1) Executing a query when no sessions are available

-> These will have "Acquiring Session", "Waiting for session to become available", "Creating Session" , "Acquired session" annotation

2) Executing a query when session is available"

 These will have "Acquiring Session", "Cache hit: has usable session" , "Acquired session" annotation

surbhigarg92: Refer https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java

src/session-pool.ts
```diff
@@ -645,12 +654,17 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       const elapsed = Date.now() - startTime;
 
       if (elapsed >= timeout!) {
+        setSpanErrorAndException(span, 'timed out');
         throw new GoogleError(errors.Timeout);
       }
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

test/session-pool.ts
odeke-em: There are assertions on the annotations for SessionPool,. Also in our discussions, we agreed that a follow-up PR this one or two shall be a PR to add tests against the Spanner Mock server which test the whole grouping hence we'll able to test out all the annotations.

src/session-pool.ts
odeke-em: Sorry about that, it was a vestige from prior code that did something on the promise. Fixed, thanks!

src/session-pool.ts
odeke-em: Keeping that documentation in, then later on once we have traceWrap in, adding explicit tests that ping doesn't produce any spans.

src/session-pool.ts
```diff
@@ -747,16 +768,19 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       } catch (e) {
         this._pending -= amount;
         this.emit('createError', e);
+        setSpanErrorAndException(span, e as Error);
         throw e;
       }
 
+      span.addEvent(`started adding ${amount} sessions to the inventory`);
```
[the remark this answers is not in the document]
odeke-em: Roger that!

src/session-pool.ts
odeke-em: Done, thanks!

src/session-pool.ts
```diff
@@ -931,11 +977,19 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       if (amount + this.size > this.options.max!) {
         amount = this.options.max! - this.size;
       }
+      span.addEvent('creating new session', {
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

src/session-pool.ts
```diff
@@ -931,11 +977,19 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+        amount: amount,
+        min: min,
+      });
+
       if (amount > 0) {
         this._pending += amount;
         promises.push(
           new Promise((_, reject) => {
             this._pending -= amount;
+            span.addEvent('creating a new session');
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

src/database.ts
```diff
@@ -821,7 +822,14 @@ class Database extends common.GrpcServiceObject {
         callback!(err as ServiceError, null, undefined);
         return;
       }
+
+      const span = getActiveOrNoopSpan();
+      span.addEvent('Creating BatchTransaction');
```
surbhigarg92, line 827: Remove this event as I feel once we add the spans , there should be a span for "BatchTransaction" , so this information will be repeated in this case.

src/database.ts
```diff
@@ -821,7 +822,14 @@ class Database extends common.GrpcServiceObject {
...
+
       const transaction = this.batchTransaction({session: session!}, options);
+      span.addEvent('BatchTransaction Creation Done', {
```
surbhigarg92, line 830: This Span needs to move to the "transaction.begin" called below. Till the time transaction.begin is called, there will not be any transaction "id"

src/database.ts
```diff
@@ -2951,6 +2962,9 @@ class Database extends common.GrpcServiceObject {
             if (session) {
               session.lastError = err as grpc.ServiceError;
             }
+            span.addEvent('No session available', {
+              'failed_session.id': session?.id,
```
surbhigarg92, line 2966: keep the name same "session.id" instead of "failed_session.id"

src/database.ts
```diff
@@ -2924,13 +2932,16 @@ class Database extends common.GrpcServiceObject {
...
     const proxyStream: Transform = through.obj();
+    const span = getActiveOrNoopSpan();
 
     this.pool_.getSession((err, session) => {
       if (err) {
         proxyStream.destroy(err);
         return;
       }
 
+      span.addEvent('Using Session', {'session.id': session?.id});
```
surbhigarg92, line 2943: Can you add the same event for all the places, where we are calling `this.pool_.getSession` in this file ? 

Same goes for `No session available` event in case of "Sessionnotfounderror"

src/session-pool.ts
```diff
@@ -630,8 +635,12 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @returns {Promise<Session>}
    */
   async _acquire(): Promise<Session> {
+    const span = getActiveOrNoopSpan();
     if (!this.isOpen) {
-      throw new GoogleError(errors.Closed);
+      span.addEvent('SessionPool is closed');
+      const err = new GoogleError(errors.Closed);
+      setSpanErrorAndException(span, err);
```
surbhigarg92, line 642: As per our discussion you are planning to create `tracewrapper` which will handle all the errors and event. 

Will this error be covered in that?

src/session-pool.ts
```diff
@@ -642,18 +651,31 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     // wrapping this logic in a function to call recursively if the session
     // we end up with is already dead
     const getSession = async (): Promise<Session> => {
+      span.addEvent('Acquiring session');
       const elapsed = Date.now() - startTime;
 
       if (elapsed >= timeout!) {
+        setSpanErrorAndException(span, 'timed out');
```
surbhigarg92, line 658: nit: Change the message from 'timed out' to "Acquiring session timed out"

src/session-pool.ts
```diff
@@ -744,9 +774,14 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
         });
 
         amount -= sessions.length;
+        nReturned += sessions.length;
       } catch (e) {
         this._pending -= amount;
         this.emit('createError', e);
+        span.addEvent(
```
surbhigarg92, line 781: We can remove this span as we are setting exception below.

src/session-pool.ts
```diff
@@ -758,6 +793,8 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
         });
       });
     }
+
+    span.addEvent(`Requested for ${nRequested} sessions returned ${nReturned}`);
```
surbhigarg92, line 797: what is the need of creating `nRequested` and  `nReturned` , we can dircetly use `amount` and `sessions.length` respectively

src/session-pool.ts
```diff
@@ -922,6 +974,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       let amount = this.options.incStep
         ? this.options.incStep
         : DEFAULTS.incStep!;
+
```
surbhigarg92, line 977: nit: remove empty line

src/session-pool.ts
```diff
@@ -931,6 +984,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
       if (amount + this.size > this.options.max!) {
         amount = this.options.max! - this.size;
       }
+
```
surbhigarg92, line 987: nit: remove empty line

src/session-pool.ts
surbhigarg92: understood.

src/session.ts
```diff
@@ -480,7 +481,11 @@ export class Session extends common.GrpcServiceObject {
     options?: TimestampBounds,
     queryOptions?: google.spanner.v1.ExecuteSqlRequest.IQueryOptions
   ) {
-    return new Snapshot(this, options, queryOptions);
+    const span = getActiveOrNoopSpan();
+    span.addEvent('Creating Snapshot');
+    const snapshot = new Snapshot(this, options, queryOptions);
+    span.addEvent('Snapshot creation done', {id: snapshot?.id?.toString()});
```
surbhigarg92, line 487: Again, the transaction is not getting created here. Add this line, only for the places where we are calling "transaction.begin"

src/session.ts
```diff
@@ -480,7 +481,11 @@ export class Session extends common.GrpcServiceObject {
     options?: TimestampBounds,
     queryOptions?: google.spanner.v1.ExecuteSqlRequest.IQueryOptions
   ) {
-    return new Snapshot(this, options, queryOptions);
+    const span = getActiveOrNoopSpan();
+    span.addEvent('Creating Snapshot');
```
surbhigarg92, line 485: remove this

src/session.ts
surbhigarg92: like this place 
https://github.com/googleapis/nodejs-spanner/blob/main/src/database.ts#L2033

and this https://github.com/googleapis/nodejs-spanner/blob/main/src/database.ts#L2773

I would recommend to add below event at https://github.com/googleapis/nodejs-spanner/blob/main/src/transaction.ts#L1419 . This method gets called from all the places whenever transaction gets created and assigns the id. This would mean you can remove the `BatchTransaction creation done` event as well.

`span.addEvent('Transaction creation done', {id: id?.toString()});`

src/session.ts
```diff
@@ -497,7 +502,18 @@ export class Session extends common.GrpcServiceObject {
...
-    return new Transaction(this, undefined, queryOptions, requestOptions);
+    const span = getActiveOrNoopSpan();
+    span.addEvent('Creating Transaction');
+    const transaction = new Transaction(
+      this,
+      undefined,
+      queryOptions,
+      requestOptions
+    );
+    span.addEvent('Transaction Creation Done', {
```
surbhigarg92, line 513: Same here, remove all the create transaction and Transaction Creation Done event and keep at only one place, recommended in above comment

src/transaction.ts
```diff
@@ -1954,6 +1955,9 @@ export class Transaction extends Dml {
     optionsOrCallback?: CommitOptions | CallOptions | CommitCallback,
     cb?: CommitCallback
   ): void | Promise<CommitResponse> {
+    const span = getActiveOrNoopSpan();
+    span.addEvent('Starting Commit');
```
surbhigarg92, line 1959: I know this is part of the scope, but as long as we add this method in "tracewrapper" , this event is not required, same goes for below event `Commit failed` , if the error gets sets, then its not requried.

src/transaction.ts
```diff
@@ -2016,6 +2029,13 @@ export class Transaction extends Dml {
         }
         err = Transaction.decorateCommitError(err as ServiceError, mutations);
 
+        if (err) {
+          setSpanError(span, err);
+          span.addEvent('Commit failed');
```
surbhigarg92, line 2034: no need for these events

test/spanner.ts
```diff
@@ -4985,6 +4993,69 @@ describe('Spanner with mock server', () => {
...
+        return evtA.time < evtB.time;
+      });
+
+      const gotEventNames: string[] = [];
+      events.forEach(event => {
+        gotEventNames.push(event.name);
+      });
+
+      const wantEventNames = [
+        'Requesting 25 sessions',
```
surbhigarg92, line 5042: These event names, doesn't look right? 

In case of success , shouldn't we have event like "Acquired session" and we shouldn't have 'Waiting for a session to become available' as it should do a cache hit.

package.json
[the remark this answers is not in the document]
surbhigarg92: We decided to remove this dependency as we have not released the last changes and its safe to remove.

package.json
surbhigarg92, line 1: We decided to remove this as we have not released the code changes which added this, "@opentelemetry/context-async-hooks"

src/session-pool.ts
odeke-em: Sadly amount gets reduced on each iteration, so we can't use it, hence why we capture it upfront.

package.json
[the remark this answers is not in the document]
odeke-em: Roger that and thanks!

src/database.ts
odeke-em: Alright, thanks! Though in our scope document we needed to annotate when a transaction is started. Correct though that we shall have the span for BatchTransaction.

src/database.ts
odeke-em: Good point, thank you!

src/database.ts
odeke-em: So "session.id" didn't make sense given the event "No session available", "failed_session.id" gave more prescription that it failed when we tried to retrieve using that sessionID, but no biggie I am making that change.

src/session-pool.ts
odeke-em: Indeed, it'll handle setting the error. I suppose let me undo this setSpanErrorAndException here.

src/database.ts
odeke-em: Roger that and done, thank you!

src/session-pool.ts
odeke-em: I think we should keep it despite the exception because if for example it gets caught, that span should still have recorded the status at that terminal point.

test/spanner.ts
odeke-em: So with how fragile the code in this file is, the creation of the Spanner object is precarious and is tightly coupled with the assertions of the test invocations and checks for suffixes with values like `*-instance-3`, so I made a TODO at the top of this `it` to later on refactor this code so that we can create our own Spanner object and check for all the events below. Those mentioned flakes were a huge delay for this review.

odeke-em:
@surbhigarg92 thank you very much for the review! Kindly help me re-run the checks an also with a fresh set of review.

src/session-pool.ts
```diff
@@ -771,6 +800,7 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
    * @returns {Promise}
    */
   async _destroy(session: Session): Promise<void> {
+    const span = getActiveOrNoopSpan();
```
surbhigarg92, line 803: nit: this is not used

odeke-em:
Yes indeed. Let's keep this one focused and smaller @surbhigarg92, thanks!

odeke-em:
@willpoint this is the PR for us to base off. @surbhigarg92 kindly please take a look. In a follow-up PR I shall make the updates to span annotation tests for transactions but this change has transaction creation annotations.

src/session-pool.ts
odeke-em: Done, thanks!

src/transaction.ts
```diff
@@ -1966,12 +1973,24 @@ export class Transaction extends Dml {
...
 
+    const span = getActiveOrNoopSpan();
+
     if (this.id) {
       reqOpts.transactionId = this.id as Uint8Array;
     } else if (!this._useInRunner) {
       reqOpts.singleUseTransaction = this._options;
     } else {
-      this.begin().then(() => this.commit(options, callback), callback);
+      this.begin().then(
```
surbhigarg92, line 1983: This is not required, as we are calling commit here again which will add `commit failed` event if error.

surbhigarg92 (approved):

surbhigarg92:
@odeke-em I have approved the PR. But Please address the new comments before planning to merge.

src/database.ts
```diff
@@ -3195,11 +3242,13 @@ class Database extends common.GrpcServiceObject {
         ? (optionsOrRunFn as RunTransactionOptions)
         : {};
 
+    let sessionId = '';
     const getSession = this.pool_.getSession.bind(this.pool_);
     // Loop to retry 'Session not found' errors.
     // (and yes, we like while (true) more than for (;;) here)
     // eslint-disable-next-line no-constant-condition
     while (true) {
+      const span = getActiveOrNoopSpan();
```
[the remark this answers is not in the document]
odeke-em: I had it in the loop to get the latest current active span in case the code down below invoked this method recursively. Removed though, thank you!

src/database.ts
```diff
@@ -3424,14 +3483,20 @@ class Database extends common.GrpcServiceObject {
         ? (optionsOrCallback as CallOptions)
         : {};
     this.pool_.getSession((err, session?, transaction?) => {
+      const span = getActiveOrNoopSpan();
       if (err && isSessionNotFoundError(err as grpc.ServiceError)) {
+        const span = getActiveOrNoopSpan();
```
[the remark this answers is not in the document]
odeke-em: Definitely left over from a prior change, done thanks!

src/session-pool.ts
```diff
@@ -865,21 +894,28 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
     let removeOnceCloseListener: Function;
     let removeListener: Function;
 
     // Wait for a session to become available.
+    span.addEvent('Waiting for a session to become available');
     const availableEvent = 'session-available';
     const promises = [
       new Promise((_, reject) => {
-        const onceCloseListener = () => reject(new GoogleError(errors.Closed));
+        const onceCloseListener = () => {
```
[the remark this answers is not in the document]
odeke-em: Oops, it was left over from a prior change, thanks and done!

src/transaction.ts
```diff
@@ -413,6 +414,9 @@ export class Snapshot extends EventEmitter {
       options,
     };
 
+    const span = getActiveOrNoopSpan();
+    span.addEvent('Creating Transaction');
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

src/transaction.ts
odeke-em: Done, thanks!

PR #2102, opened 2024-09-04
chore(samples): add samples for spanner editions

alkatrivedi:
This PR contains create and update samples for Spanner Editions.


samples/instance.js
```diff
@@ -91,6 +95,15 @@ require('yargs')
   .example(
     'node $0 createInstanceWithProcessingUnits "my-instance" "my-project-id"'
   )
+  .command(
```
surbhigarg92, line 98: why do we have update instance here, when we have created a new file ?

samples/system-test/spanner.test.js
```diff
@@ -259,21 +259,41 @@ describe('Autogenerated Admin Clients', () => {
...
         )
       );
       assert.match(
-        output,
+        createInstanceOutput,
         new RegExp(`Created instance ${SAMPLE_INSTANCE_ID}.`)
       );
+
+      const updateInstanceOutput = execSync(
+        `${instanceCmd} updateInstance "${SAMPLE_INSTANCE_ID}" ${PROJECT_ID}`
```
surbhigarg92, line 279: Instead of calling the same instanceCmd with "updateInstance" , you can directly invoke the "instance-update.js" file

samples/instance.js
alkatrivedi: updated the sample to rubric

samples/system-test/spanner.test.js
alkatrivedi: done

surbhigarg92 (approved):

nithinsujir (approved):
Thanks for making the changes!

PR #2105, opened 2024-09-10
observability: ensure that each span has otel.scope.name,version

odeke-em:
This change ensures that each span created by the code in this library is annotated with:
* otel.scope.name = "cloud.google.com/nodejs/spanner"
* otel.scope.version = <LIBRARY_VERSION>

Updates #2079

### DIFF
--- a/src/instrument.ts
+++ b/src/instrument.ts
@@ -52,12 +52,10 @@ interface observabilityOptions {
 
 export type {observabilityOptions as ObservabilityOptions};
 
-const LIB_FQNAME = 'cloud.google.com/nodejs/spanner';
-const LIB_VERSION = '7.14.0'; // Manually hard coded, TODO: remove
-const TRACER_NAME = LIB_FQNAME;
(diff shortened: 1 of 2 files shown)


odeke-em:
Kindly cc-ing @surbhigarg92

src/instrument.ts
```diff
@@ -50,8 +52,12 @@ interface observabilityOptions {
 
 export type {observabilityOptions as ObservabilityOptions};
 
-const TRACER_NAME = 'cloud.google.com/nodejs/spanner';
-const TRACER_VERSION = '7.14.0'; // Manually hard coded, TODO: remove
+const LIB_FQNAME = 'cloud.google.com/nodejs/spanner';
```
surbhigarg92, line 55: I don't think we need to maintain two separate variables for `LIB_FQNAME` and `LIB_VERSION`. 

We can reuse `TRACER_NAME` and `TRACER_VERSION`

surbhigarg92 (approved):

Issue #2114, opened 2024-09-17
observability: use manual tracing instead of auto-traceWrapping until opentelemetry-instrumentation-spanner is ported


odeke-em:
This issue serves as a decision marker and record for our method of attack to get observability in for this package. I implemented 
2 kinds of tracing to ensure the team had a buffet of options:
* manual tracing per PR https://github.com/googleapis/nodejs-spanner/pull/2087 complete with tests and end-to-end working samples and docs
* auto-instrumentation with traceWrapping per PR https://github.com/googleapis/python-spanner/pull/1199 complete with tests and end-to-end working samples and docs

Internally at Google it was discussed and resolved with @surbhigarg92 representing us when tabled with the OpenTelemetry team at Google that:
- [ ] traceWrap was well received but we should send that instrumentation to be hosted at opentelemetry-js-contrib
- [ ] manually instrument the libraries currently

One thing to note is that we shall need to add feature guards that will ensure that customers don't doubly import the future opentelemetry-instrumentation-spanner and then also with the manual spans.

PR #2115, opened 2024-09-17
feat: (observability) add spans for BatchTransaction and Table

odeke-em:
This change is part of a series of changes to add
OpenTelemetry traces, focused on BatchTransaction and Table.

While here, made the tests for sessionPool spans much more precise to avoid flakes.

Updates #2079
Built from PR #2087
Updates #2114

### DIFF
--- a/observability-test/batch-transaction.ts
+++ b/observability-test/batch-transaction.ts
@@ -1,0 +1,255 @@
+/*!
+ * Copyright 2024 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-disable prefer-rest-params */
+
+import {util} from '@google-cloud/common';
+import * as pfy from '@google-cloud/promisify';
+import * as assert from 'assert';
+import {before, beforeEach, afterEach, describe, it} from 'mocha';
+import * as extend from 'extend';
+import * as proxyquire from 'proxyquire';
+import * as sinon from 'sinon';
+const {
+  AlwaysOnSampler,
+  NodeTracerProvider,
+  InMemorySpanExporter,
+} = require('@opentelemetry/sdk-trace-node');
+// eslint-disable-next-line n/no-extraneous-require
+const {SimpleSpanProcessor} = require('@opentelemetry/sdk-trace-base');
+import {Session, Spanner} from '../src';
+import * as bt from '../src/batch-transaction';
+
+const fakePfy = extend({}, pfy, {
+  promisifyAll(klass, options) {
+    if (klass.name !== 'BatchTransaction') {
+      return;
+    }
+    assert.deepStrictEqual(options.exclude, ['identifier']);
+  },
+});
+
+class FakeTimestamp {
+  calledWith_: IArguments;
+  constructor() {
+    this.calledWith_ = arguments;
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const fakeCodec: any = {
+  encode: util.noop,
+  Timestamp: FakeTimestamp,
+  Int() {},
+  Float() {},
+  SpannerDate() {},
+  convertProtoTimestampToDate() {},
+};
+
+const SPANNER = {
+  routeToLeaderEnabled: true,
+};
+
+const INSTANCE = {
+  parent: SPANNER,
+};
+
+const DATABASE = {
+  formattedName_: 'database',
+  parent: INSTANCE,
+};
+
+class FakeTransaction {
+  calledWith_: IArguments;
+  session;
+  constructor(session) {
+    this.calledWith_ = arguments;
+    this.session = session;
+  }
+  static encodeKeySet(): object {
+    return {};
+  }
+  static encodeParams(): object {
+    return {};
+  }
+
+  _getSpanner(): Spanner {
+    return SPANNER as Spanner;
+  }
+
+  run() {}
+  read() {}
+}
+
+describe('BatchTransaction', () => {
+  const sandbox = sinon.createSandbox();
+
+  // tslint:disable-next-line variable-name
+  let BatchTransaction: typeof bt.BatchTransaction;
+  let batchTransaction: bt.BatchTransaction;
+
+  before(() => {
+    BatchTransaction = proxyquire('../src/batch-transaction.js', {
+      '@google-cloud/precise-date': {PreciseDate: FakeTimestamp},
+      '@google-cloud/promisify': fakePfy,
+      './codec.js': {codec: fakeCodec},
+      './transaction.js': {Snapshot: FakeTransaction},
+    }).BatchTransaction;
+  });
+
+  const traceExporter = new InMemorySpanExporter();
+  const sampler = new AlwaysOnSampler();
+
+  const provider = new NodeTracerProvider({
+    sampler: sampler,
+    exporter: traceExporter,
+  });
+  provider.addSpanProcessor(new SimpleSpanProcessor(traceExporter));
+
+  afterEach(() => {
+    traceExporter.reset();
+    sandbox.restore();
+  });
+
+  const REQUEST = sandbox.stub();
+  const SESSION = {
+    parent: DATABASE,
+    formattedName_: 'abcdef',
+    request: REQUEST,
+  };
+  const ID = '0xdeadbeef';
+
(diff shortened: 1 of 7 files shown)


src/table.ts
```diff
@@ -1072,29 +1073,34 @@ class Table {
...
+            span.end();
+            callback(err);
+            return;
+          }
 
-        transaction![method](this.name, rows as Key[]);
-        transaction!.commit(options, callback);
-      }
-    );
+          span.end();
```
surbhigarg92, line 1098: Why are we doing `span.end` before commit call ?

surbhigarg92:
@odeke-em I don't see any tests for these spans.

odeke-em:
Hey @surbhigarg92 I hadn't uploaded some of them as I created this PR early morning at around 3AM before bed and  in interludes. I've added comprehensive tests, please take a look.

src/transaction.ts
```diff
@@ -285,6 +286,7 @@ export class Snapshot extends EventEmitter {
   queryOptions?: IQueryOptions;
   resourceHeader_: {[k: string]: string};
   requestOptions?: Pick<IRequestOptions, 'transactionTag'>;
+  observabilityOptions?: ObservabilityOptions;
```
surbhigarg92, line 289: `observabilityOptions` should be injected by customers from [SpannerOptions](https://github.com/googleapis/nodejs-spanner/blob/main/src/index.ts#L134) which should set this property. Do you plan to do that in separate PR ? 
I would recommend to add that in the same PR, so that we can test batch-transaction and mutation flow end to end

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
...
+      batchTransaction.session = SESSION as {} as Session;
+      batchTransaction.id = ID;
+      batchTransaction.observabilityOptions = {tracerProvider: provider};
+
+      REQUEST.callsFake((_, callback) => callback(null, RESPONSE));
+    });
+
+    const GAX_OPTS = {};
+
+    const fakeDirectedReadOptionsForRequest = {
```
surbhigarg92, line 531: This is optional param , so better to remove it to keep the test short. 
Same goes for dataBoostEnabled

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
...
+          // Sort the spans by duration.
+          spans.sort((spanA, spanB) => {
+            spanA.duration < spanB.duration;
+          });
+
+          const gotSpanNames: string[] = [];
+          spans.forEach(span => {
+            gotSpanNames.push(span.name);
+          });
+          const wantSpanNames = [
```
surbhigarg92, line 581: nit: rename to `expectedSpans`

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
...
+            true,
+            'at least 2 spans expected'
+          );
+
+          // Sort the spans by duration.
+          spans.sort((spanA, spanB) => {
+            spanA.duration < spanB.duration;
+          });
+
+          const gotSpanNames: string[] = [];
```
surbhigarg92, line 577: nit: rename to `actualSpans` . Here and elsewhere.

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
...
+
+          const spans = traceExporter.getFinishedSpans();
+          assert.strictEqual(
+            spans.length >= 2,
+            true,
+            'at least 2 spans expected'
+          );
+
+          // Sort the spans by duration.
+          spans.sort((spanA, spanB) => {
```
surbhigarg92, line 573: Why do we need to sort ?

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
...
+        (err, part, resp) => {
+          assert.ifError(err);
+
+          traceExporter.forceFlush();
+
+          const spans = traceExporter.getFinishedSpans();
+          assert.strictEqual(
+            spans.length >= 2,
+            true,
+            'at least 2 spans expected'
```
surbhigarg92, line 569: We should do strictEqual check. We expect exactly 2 spans not at least 2.

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
...
+    it('createQueryPartitions', done => {
+      const REQUEST = sandbox.stub();
+      const response = {};
+      REQUEST.callsFake((_, callback) => callback(null, response));
+
+      const res = batchTransaction.createQueryPartitions(
+        QUERY,
+        (err, part, resp) => {
+          assert.ifError(err);
+
```
surbhigarg92, line 562: nit: remove empty lines

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
...
+          spans.sort((spanA, spanB) => {
+            spanA.duration < spanB.duration;
+          });
+
+          const gotSpanNames: string[] = [];
+          spans.forEach(span => {
+            gotSpanNames.push(span.name);
+          });
+          const wantSpanNames = [
+            'CloudSpanner.BatchTransaction.createPartitions_',
```
surbhigarg92, line 582: Is it possible to test that `createPartitions_` is a child span of `createQueryPartitions`

test/session-pool.ts
```diff
@@ -1412,25 +1412,13 @@ describe('SessionPool', () => {
...
 
-        const spans = exporter.getFinishedSpans();
+        const events = span.events;
+        assert.strictEqual(!events, false, 'Events must be set');
         assert.strictEqual(
-          spans.length,
-          1,
-          'Exactly 1 span should have been exported'
-        );
-        assert.strictEqual(
```
surbhigarg92, line 1421: Why did we remove this check from here ?

test/batch-transaction.ts
odeke-em: We need at least 2 because as we add in spans to other methods, the test shall need modification but sure I can modify it.

test/batch-transaction.ts
odeke-em: Spans aren't guaranteed to be returned in the exact order of their exporting.

test/batch-transaction.ts
odeke-em: Absolutely! Good point.

test/batch-transaction.ts
```diff
@@ -485,4 +491,147 @@ describe('BatchTransaction', () => {
       assert.strictEqual(identifier.timestamp, TIMESTAMP);
     });
   });
+
+  describe('observability traces', () => {
```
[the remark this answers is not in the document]
odeke-em: Done, added for both batch-transaction.ts and table.ts

test/batch-transaction.ts
odeke-em: Done, thanks!

test/batch-transaction.ts
odeke-em: Done, thanks!

test/batch-transaction.ts
odeke-em: Done, thanks!

test/batch-transaction.ts
odeke-em: Done, thanks!

test/batch-transaction.ts
odeke-em: Done, thanks!

test/batch-transaction.ts
odeke-em: Done.

src/transaction.ts
odeke-em: I shall have it plumbed in progressively as we get there. The reason being that this PR is bound to balloon if I add it straight from the top inside `new Spanner({observabilityOptions: opts})`. Just the way for each part I have been adding it in and testing.

test/session-pool.ts
odeke-em: In the testing environment, when ran isolated, the checks pass but when ran as a whole this test flakes here and there, a potential problem with OpenTelemetry. Besides, the point of the code was to ensure that the parent span actually has the required annotations which we do have and now it is very well cut out just for that.

observability-test/batch-transaction.ts
```diff
@@ -0,0 +1,269 @@
...
+    }).BatchTransaction;
+  });
+
+  beforeEach(() => {
+    batchTransaction = new BatchTransaction(SESSION as {} as Session);
+  });
+
+  afterEach(() => sandbox.restore());
+
+  describe('observability traces', () => {
```
surbhigarg92, line 125: nit rename to `BatchTransaction traces`

test/session-pool.ts
```diff
@@ -1367,34 +1352,12 @@ describe('SessionPool', () => {
   });
 
   describe('trace annotations on active span', () => {
```
surbhigarg92, line 1354: maybe for separate PR. But It will be good to extract these tests in `observability-test` folder

surbhigarg92 (approved):

observability-test/batch-transaction.ts
odeke-em: Done, I instead removed the describe given that I moved the code from where I had written it nested inside test/batch-transaction.ts to here. This now what the log looks like
Screenshot 2024-09-19 at 7 36 26 PM

PR #2121, opened 2024-09-21
feat(observability): add tracing spans to Session

odeke-em:
This change adds tracing spans to Session for methods:
* create
* getMetadata
* keepAlive

and corresponding tests.

Updates #2079
Built from PR #2087
Updates #2114

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -44,6 +44,7 @@ import {
 import {grpc, CallOptions} from 'google-gax';
 import IRequestOptions = google.spanner.v1.IRequestOptions;
 import {Spanner} from '.';
+import {ObservabilityOptions, startTrace, setSpanError} from './instrument';
 
 export type GetSessionResponse = [Session, r.Response];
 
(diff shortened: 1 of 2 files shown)


src/session.ts
```diff
@@ -388,23 +402,32 @@ export class Session extends common.GrpcServiceObject {
...
-        reqOpts,
-        gaxOpts,
-        headers: headers,
-      },
-      (err, resp) => {
-        if (resp) {
-          resp.databaseRole = resp.creatorRole;
-          delete resp.creatorRole;
-          this.metadata = resp;
+
```
surbhigarg92, line 405: remove for this method. We dont need it for now. Can add later if we need

src/session.ts
```diff
@@ -247,14 +249,24 @@ export class Session extends common.GrpcServiceObject {
         this.databaseRole =
           options.databaseRole || database.databaseRole || null;
 
-        return database.createSession(options, (err, session, apiResponse) => {
-          if (err) {
-            callback(err, null, apiResponse);
-            return;
-          }
+        const q = {opts: database.observabilityConfig};
+        return startTrace('Session.create', q, span => {
```
surbhigarg92, line 253: I think we dont need this as well, as we dont use single session creation , instead we use batch create session

surbhigarg92 (commented):
I believe we dont need this PR. All the important spans are already there for sessions. 

Lete keep this on hold and move to other PRs.

odeke-em:
Alright, thanks @surbhigarg92! Sure we can put it on ice and move onto others!

odeke-em:
Deemed as unnecessary!

PR #2131, opened 2024-09-28
feat: (observability) Add support for OpenTelemetry traces and allow observability options to be passed. 

odeke-em:
This change plumbs ObservabilityConfig into Spanner, Instance and Database so that
any subsequent traces will use it when beginning spans and later on for metrics.

Updates #2079


src/index.ts
```diff
@@ -138,6 +139,7 @@ export interface SpannerOptions extends GrpcClientOptions {
   sslCreds?: grpc.ChannelCredentials;
   routeToLeaderEnabled?: boolean;
   directedReadOptions?: google.spanner.v1.IDirectedReadOptions | null;
+  observabilityConfig?: ObservabilityOptions | undefined;
```
surbhigarg92, line 142: We don't need to use "undefined" if a field is marked optional using "?"

Also please add the documentation for the new options similar to other options. 

We will also need to define "Observability Options" type

src/instance.ts
```diff
@@ -164,6 +165,7 @@ class Instance extends common.GrpcServiceObject {
   databases_: Map<string, Database>;
   metadata?: IInstance;
   resourceHeader_: {[k: string]: string};
+  observabilityConfig: ObservabilityOptions | undefined;
```
surbhigarg92, line 168: we should define observabilityConfig as optional using "?" instead of using "undefined". Same standard should be used everywhere. 

Also use param name `observabilityOptions` same as the type name.

src/database.ts
```diff
@@ -337,6 +337,7 @@ class Database extends common.GrpcServiceObject {
   databaseDialect?: EnumKey<
     typeof databaseAdmin.spanner.admin.database.v1.DatabaseDialect
   > | null;
+  observabilityConfig: ObservabilityOptions | undefined;
```
surbhigarg92, line 340: should use underscore to refer this property as private. Else customer can override the property at database level . Or is there any other way we can avoid this ?

src/instance.ts
surbhigarg92: Similar to Database , use underscore to mark this property as private.

test/spanner.ts
```diff
@@ -5061,6 +5061,62 @@ describe('Spanner with mock server', () => {
       );
     });
   });
+
+  it('Plumb ObservabilityConfig into Spanner, Instance and Database', () => {
```
surbhigarg92, line 5065: Rename this test `Should use passed ObservabilityOptions in Spanner, Instance and Database"

surbhigarg92, line 5065: Add one more test to verify below 
1) Register a global exporter
2) Inject a different exporter via Spanner 
3) Verify that injected exporter is used and not global.

src/index.ts
odeke-em: What's the precedent on line 141 `directedReadOptions?: google.spanner.v1.IDirectedReadOptions | null;`?

ObservabilityOptions was defined in src/instrument.ts

odeke-em: I've exposed ObservabilityOptions as a type itself.

src/instance.ts
odeke-em: Done, I've marked it as private.

test/spanner.ts
odeke-em: @surbhigarg92 could we please add that test after we've merged in the other code for Database? It's currently so cumbersome and walls of code just to add just an end-to-end routine for such. Currently all the PRs are so laden with tons of code and spiraling out of control.

odeke-em: Done, thanks!

odeke-em: I added a TODO in the tests for it. I spent time trying to plumb it through and that was wasted, meanwhile we could simply wait until we've merged the end-to-end code that uses MockSpanner and then simply have it running end-to-end and much simpler.

odeke-em: And I've added the test, please take a look @surbhigarg92!

src/database.ts
odeke-em: I suffixed it with an underscore so `observabilityConfig_`.

src/instance.ts
odeke-em: Done.

odeke-em:
Kindly help me run the bots @surbhigarg92 @harshachinta, thank you!

src/database.ts
```diff
@@ -344,7 +344,7 @@ class Database extends common.GrpcServiceObject {
   databaseDialect?: EnumKey<
     typeof databaseAdmin.spanner.admin.database.v1.DatabaseDialect
   > | null;
-  observabilityConfig: ObservabilityOptions | undefined;
+  observabilityOptions_?: ObservabilityOptions;
```
surbhigarg92, line 347: Can you try using "private" keyword . Also it should be `_observabilityOptions` right ?

src/index.ts
```diff
@@ -138,6 +139,7 @@ export interface SpannerOptions extends GrpcClientOptions {
   sslCreds?: grpc.ChannelCredentials;
   routeToLeaderEnabled?: boolean;
   directedReadOptions?: google.spanner.v1.IDirectedReadOptions | null;
+  observabilityOptions?: ObservabilityOptions;
```
surbhigarg92, line 142: Add documentation for observabilityOptions above interface similar to other options

src/database.ts
odeke-em: I was following the precedent in this library of `formattedName_` as I hadn't seen other variables beginning with the underscore but sure I can update that. The only problem about declaring them with the keyword private is that between classes we actually need to read and set the value and adding a setter seems like overkill. What I could do is set it to `readonly` and infer it only from the constructor as well as having set it as _observabilityOptions as you suggested.

src/index.ts
odeke-em: Done, thank you @surbhigarg92!

src/database.ts
odeke-em: So @surbhigarg92 I don't believe that reasonable customers expect to set `formattedName_` nor would a reasonable customer change the now `_observabilityOptions`, even more, making a setter or getter is unnecessary exposure.

src/instrument.ts
```diff
@@ -45,12 +45,21 @@ interface SQLStatement {
...
 
-interface observabilityOptions {
+/*
+ * ObservabilityOptions defines the configuration with which
+ * startTrace may be invoked if set.
+ * @property [tracerProvider] is the injected TracerProvider to use,
+ * otherwise the global TracerProvider shall be used.
+ * @property [enableExtendedTracing] when set signifies that spans started
+ * with an accompanying an SQL statement shall be annotated with that SQL.
+ * Alternatively, you could set environment variable `SPANNER_ENABLE_EXTENDED_TRACING=true`.
```
surbhigarg92, line 55:
```suggestion
 * @property [enableExtendedTracing] Sets whether to enable extended OpenTelemetry tracing. Enabling this option will add the
     * following additional attributes to the traces that are generated by the client
     * db.statement: Contains the SQL statement that is being executed.
 * Alternatively, you could set environment variable `SPANNER_ENABLE_EXTENDED_TRACING=true`.
```

src/instrument.ts
```diff
@@ -45,12 +45,21 @@ interface SQLStatement {
   sql: string;
 }
 
-interface observabilityOptions {
+/*
+ * ObservabilityOptions defines the configuration with which
+ * startTrace may be invoked if set.
+ * @property [tracerProvider] is the injected TracerProvider to use,
+ * otherwise the global TracerProvider shall be used.
```
surbhigarg92, line 52:
```suggestion
 * ObservabilityOptions defines the configuration to be used for Spanner OpenTelemetry Traces.
 * @property [tracerProvider] Sets the TracerProvider to be used for traces,
 * Global TracerProvider will be used as a fallback.
```

src/index.ts
```diff
@@ -130,6 +131,8 @@ export type GetInstanceConfigOperationsCallback = PagedCallback<
  * @property {google.spanner.v1.IDirectedReadOptions} [directedReadOptions] Sets the DirectedReadOptions for all ReadRequests and ExecuteSqlRequests for the Client.
  * Indicates which replicas or regions should be used for non-transactional reads or queries.
  * DirectedReadOptions won't be set for readWrite transactions"
+ * @property {ObservabilityOptions} [observabilityOptions] defines the configuration
+ * for OpenTelemetry Tracing.
```
surbhigarg92, line 135:
```suggestion
 * @property {ObservabilityOptions} [observabilityOptions] Sets the observability options to be used for OpenTelemetry tracing
```

surbhigarg92 (approved):

src/instrument.ts
odeke-em: Thank you @surbhigarg92! Isn't hard-coding the variable defeating the purpose of prose documentation for customers or are we more like pointing them to the actual variable to look for?

src/index.ts
odeke-em: thank you, done @surbhigarg92!

odeke-em:
@surbhigarg92 thanks for the review! Kindly help me run the bots

src/instrument.ts
odeke-em: Done and thank you @surbhigarg92! Kindly help me re-run the bots.

odeke-em:
Updated with the latest code from main, kindly please help me re-run the bots @surbhigarg92 @rahul2393 @harshachinta.

PR #2134, opened 2024-09-29
feat: (observability) trace Instance

odeke-em:
Trace instruments the Instance class.

Updates #2079


surbhigarg92 (commented):
Lets keep this as well on hold. 

We will finish everything related to dataClient , which is running transaction , queries etc. Changes in this file are for database operations. So in order to keep things simple, I would like to first complete the dataClient first, test, release and then see if we want to add this.

PR #2143, opened 2024-10-03
fix: exact staleness timebound

### LINKED ISSUE #2129: Stale reads documentation and samples are outdated
Version v3.0.0 introduced a breaking change where timestamp bounds now expect [milliseconds precision](https://github.com/googleapis/nodejs-spanner/blob/main/CHANGELOG.md#timestamp-bounds-now-offer-nanosecond-precision-506) but in the [sample code](https://github.com/googleapis/nodejs-spanner/blob/main/samples/crud.js#L313) there is a comment that staleness is specified in seconds. Same issue is also present in the [documentation ](https://cloud.google.com/spanner/docs/reads#node.js:~:text=const%20options%20%3D%20%7B%0A%20%20//%20Guarantees%20that%20all%20writes%20committed%20more%20than%2015%20seconds%20ago%20are%20visible%0A%20%20exactStaleness%3A%2015%2C%0A%7D%3B) for the client library.

Sample should have following options to perform 15 seconds stale read:
```  
const options = {
    // Guarantees that all writes committed more than 15 seconds ago are visible
    exactStaleness: 15000,
};
```

alkatrivedi:
fixes: #2129


samples/crud.js
```diff
@@ -309,8 +309,8 @@ async function readStaleData(instanceId, databaseId, projectId) {
   };
 
   const options = {
-    // Guarantees that all writes committed more than 15 seconds ago are visible
-    exactStaleness: 15,
+    // Guarantees that all writes committed more than 15 milliseconds ago are visible
```
surbhigarg92, line 312:
```suggestion
    // Guarantees that all writes committed more than 1500 milliseconds ago are visible
```

surbhigarg92 (approved):

samples/crud.js
alkatrivedi: done

PR #2122, opened 2024-09-22
feat(observability): trace Transaction

odeke-em:
This change adds observability tracing for
Transaction along with tests.

Updates #2079
Built from PR #2087
Updates #2114

PR #2145, opened 2024-10-04
feat: (observability) trace Database.batchCreateSessions + SessionPool.createSessions

### LINKED ISSUE #2146: observability: if an async context manager has not been enabled, make sure to register one
Reported offline by @alkatrivedi, whereby their code didn't involve creating a new traces at all but was using this instrumented package and their code used async/await yet spans that are supposed to be nested were disjoint.

OpenTelemetry-JS acknowledges that for code to work correctly and for context to be correctly propagated (as a replacement for threadlocal storage) a context manager must be enabled beforehand https://opentelemetry.io/docs/languages/js/context/

Customers should never have to think about these reasons and do it themselves, instead our package should automatically handle this scenario so that customers can transparently keep their code without any needs to edit
 
## Task
- [ ] Check for if a prior context manager was enabled and if not, create oen like this ideally in src/instrument.ts
```javascript
const {AsyncHooksContextManager} = require('@opentelemetry/context-async-hooks');
const {context} = require('@opentelemetry/api');
const contextManager = new AsyncHooksContextManager();
contextManager.enable();
context.setGlobalContextManager(contextManager);
```

along with regression tests for it.

odeke-em:
This change adds tracing for Database.batchCreateSessions
as well as SessionPool.createSessions which was raised
as a big need.
This change is a premise to finishing up tracing Transaction.

While here, also folded in the async/await fix to avoid day+ long
code review lag and then 3+ hours just to run tests per PR:
OpenTelemetry cannot work correctly for async/await if there isn't
a set AsyncHooksManager, but we should not burden our customers with
this type of specialist knowledge, their code should just work and
this change performs such a check. Later on we shall file a feature
request with the OpenTelemetry-JS API group to give us a hook to detect
if we've got a live asyncHooksManager instead of this mandatory
comparison to ROOT_CONTEXT each time.

Fixes #2146
Updates #2079
Spun out of PR #2122
Supersedes PR #2147

### DIFF
--- a/observability-test/database.ts
+++ b/observability-test/database.ts
@@ -375,6 +375,115 @@ describe('Database', () => {
     });
   });
 
+  describe('batchCreateSessions', () => {
+    it('without error', done => {
+      const ARGS = [null, [{}]];
+      database.request = (config, callback) => {
+        callback(...ARGS);
+      };
+
+      database.batchCreateSessions(10, (err, sessions) => {
+        assert.ifError(err);
+        assert.ok(sessions);
+
+        traceExporter.forceFlush();
+        const spans = traceExporter.getFinishedSpans();
+
+        const actualSpanNames: string[] = [];
+        const actualEventNames: string[] = [];
+        spans.forEach(span => {
+          actualSpanNames.push(span.name);
+          span.events.forEach(event => {
+            actualEventNames.push(event.name);
+          });
+        });
+
+        const expectedSpanNames = ['CloudSpanner.Database.batchCreateSessions'];
+        assert.deepStrictEqual(
+          actualSpanNames,
+          expectedSpanNames,
+          `span names mismatch:\n\tGot:  ${actualSpanNames}\n\tWant: ${expectedSpanNames}`
+        );
+
+        // Ensure that the span didn't encounter an error.
+        const firstSpan = spans[0];
+        assert.strictEqual(
+          SpanStatusCode.UNSET,
+          firstSpan.status.code,
+          'Unexpected span status code'
+        );
+        assert.strictEqual(
+          undefined,
+          firstSpan.status.message,
+          'Mismatched span status message'
+        );
+
+        // We don't expect events.
+        const expectedEventNames = [];
+        assert.deepStrictEqual(
+          actualEventNames,
+          expectedEventNames,
+          `Unexpected events:\n\tGot:  ${actualEventNames}\n\tWant: ${expectedEventNames}`
+        );
+
+        done();
+      });
+    });
+
+    it('with error', done => {
+      const ARGS = [new Error('batchCreateSessions.error'), null];
+      database.request = (config, callback) => {
+        callback(...ARGS);
+      };
+
+      database.batchCreateSessions(10, (err, sessions) => {
+        assert.ok(err);
+        assert.ok(!sessions);
+        traceExporter.forceFlush();
+        const spans = traceExporter.getFinishedSpans();
+
+        const actualSpanNames: string[] = [];
+        const actualEventNames: string[] = [];
+        spans.forEach(span => {
+          actualSpanNames.push(span.name);
+          span.events.forEach(event => {
+            actualEventNames.push(event.name);
+          });
+        });
+
+        const expectedSpanNames = ['CloudSpanner.Database.batchCreateSessions'];
+        assert.deepStrictEqual(
+          actualSpanNames,
+          expectedSpanNames,
+          `span names mismatch:\n\tGot:  ${actualSpanNames}\n\tWant: ${expectedSpanNames}`
+        );
+
+        // Ensure that the span actually produced an error that was recorded.
+        const firstSpan = spans[0];
+        assert.strictEqual(
+          SpanStatusCode.ERROR,
+          firstSpan.status.code,
+          'Expected an ERROR span status'
+        );
+        assert.strictEqual(
+          'batchCreateSessions.error',
+          firstSpan.status.message,
+          'Mismatched span status message'
+        );
+
+        // We don't expect events.
+        const expectedEventNames = [];
+        assert.deepStrictEqual(
+          actualEventNames,
+          expectedEventNames,
+          `Unexpected events:\n\tGot:  ${actualEventNames}\n\tWant: ${expectedEventNames}`
+        );
+
+        done();
+      });
+    });
+  });
+
   describe('getSnapshot', () => {
     let fakePool: FakeSessionPool;
     let fakeSession: FakeSession;
--- a/observability-test/database.ts
+++ b/observability-test/database.ts
@@ -518,7 +518,7 @@ describe('Database', () => {
 
       getSessionStub.callsFake(callback => callback(fakeError, null));
 
-      database.getSnapshot((err, snapshot) => {
+      database.getSnapshot(err => {
         assert.strictEqual(err, fakeError);
         traceExporter.forceFlush();
         const spans = traceExporter.getFinishedSpans();
@@ -1136,7 +1136,6 @@ describe('Database', () => {
     });
 
     it('with error on null mutation should catch thrown error', done => {
-      const fakeError = new Error('err');
       try {
         database.writeAtLeastOnce(null, (err, res) => {});
       } catch (err) {
--- a/observability-test/session-pool.ts
+++ b/observability-test/session-pool.ts
@@ -32,7 +32,6 @@ const {SimpleSpanProcessor} = require('@opentelemetry/sdk-trace-base');
 import {Database} from '../src/database';
 import {Session} from '../src/session';
 import * as sp from '../src/session-pool';
-import {Transaction} from '../src/transaction';
 
 let pQueueOverride: typeof PQueue | null = null;
 
--- a/observability-test/session-pool.ts
+++ b/observability-test/session-pool.ts
@@ -1,0 +1,201 @@
+/*!
+ * Copyright 2024 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as assert from 'assert';
+import {before, beforeEach, afterEach, describe, it} from 'mocha';
+import * as events from 'events';
+import * as extend from 'extend';
+import PQueue from 'p-queue';
+import * as proxyquire from 'proxyquire';
+import * as sinon from 'sinon';
+import stackTrace = require('stack-trace');
+import timeSpan = require('time-span');
+const {
+  AlwaysOnSampler,
+  NodeTracerProvider,
+  InMemorySpanExporter,
+} = require('@opentelemetry/sdk-trace-node');
+// eslint-disable-next-line n/no-extraneous-require
+const {SimpleSpanProcessor} = require('@opentelemetry/sdk-trace-base');
+
+import {Database} from '../src/database';
+import {Session} from '../src/session';
+import * as sp from '../src/session-pool';
+import {Transaction} from '../src/transaction';
+
+let pQueueOverride: typeof PQueue | null = null;
+
+function FakePQueue(options) {
+  return new (pQueueOverride || PQueue)(options);
+}
+
+FakePQueue.default = FakePQueue;
+
+class FakeTransaction {
+  options;
+  constructor(options?) {
+    this.options = options;
+  }
+  async begin(): Promise<void> {}
+}
+
+const fakeStackTrace = extend({}, stackTrace);
+
+function noop() {}
+
+describe('SessionPool', () => {
+  let sessionPool: sp.SessionPool;
+  // tslint:disable-next-line variable-name
+  let SessionPool: typeof sp.SessionPool;
+
+  function noop() {}
+  const DATABASE = {
+    batchCreateSessions: noop,
+    databaseRole: 'parent_role',
+  } as unknown as Database;
+
+  const sandbox = sinon.createSandbox();
+  const shouldNotBeCalled = sandbox.stub().throws('Should not be called.');
+
+  const createSession = (name = 'id', props?): Session => {
+    props = props || {};
+
+    return Object.assign(new Session(DATABASE, name), props, {
+      create: sandbox.stub().resolves(),
+      delete: sandbox.stub().resolves(),
+      keepAlive: sandbox.stub().resolves(),
+      transaction: sandbox.stub().returns(new FakeTransaction()),
+    });
+  };
+
+  before(() => {
+    SessionPool = proxyquire('../src/session-pool.js', {
+      'p-queue': FakePQueue,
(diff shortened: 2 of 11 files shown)


src/session-pool.ts
```diff
@@ -164,6 +169,7 @@ export interface SessionPoolOptions {
   writes?: number;
   incStep?: number;
   databaseRole?: string | null;
+  observabilityOptions?: ObservabilityOptions;
```
surbhigarg92, line 172: We should not add `observabilityOptions` as part of `SessionPoolOptions` . Customer pass the SessionPoolOptions and it will be confusing for them on what to use. Create this as a private property in sessionpool class, similar to _traces property

surbhigarg92:
Can you also add record session leaks in opentelemetry traces. 

Add the [this._getLeaks()](https://github.com/googleapis/nodejs-spanner/blob/main/src/session-pool.ts#L904) in traces, also

src/session-pool.ts
odeke-em: Alright removed, thanks @surbhigarg92!

odeke-em:
And rebased from main, good to get to bots testing and review updates @alkatrivedi @surbhigarg92!

observability-test/spanner.ts
```diff
@@ -457,50 +506,79 @@ describe('EndToEnd', () => {
         done();
       });
     });
+
+    it('batchCreateSessions', done => {
+      const blankMutations = new MutationSet();
```
alkatrivedi, line 511: remove ?

src/instrument.ts
```diff
@@ -93,6 +95,32 @@ interface traceConfig {
...
+ * spans resulting from async/await invocations won't be correctly
+ * associated in their respective hierarchies.
+ */
+function ensureInitialContextManagerSet() {
+  if (context.active() === ROOT_CONTEXT) {
+    // If no active context was set previously, trace context propagation cannot
+    // function correctly with async/await for OpenTelemetry and they acknowledge
+    // this fact per https://opentelemetry.io/docs/languages/js/context/#active-context
+    // but we shouldn't make our customers have to invasively edit their code
+    // nor should they be burdened about these facts, their code should JUST work.
```
alkatrivedi, line 115: remove

src/instrument.ts
```diff
@@ -93,6 +95,32 @@ interface traceConfig {
...
+ * checking if context.active() returns an invalid/unset context
+ * and if so, sets a global AsyncHooksContextManager otherwise
+ * spans resulting from async/await invocations won't be correctly
+ * associated in their respective hierarchies.
+ */
+function ensureInitialContextManagerSet() {
+  if (context.active() === ROOT_CONTEXT) {
+    // If no active context was set previously, trace context propagation cannot
+    // function correctly with async/await for OpenTelemetry and they acknowledge
+    // this fact per https://opentelemetry.io/docs/languages/js/context/#active-context
```
alkatrivedi, line 113: // If no active context was set previously, trace context propagation cannot
// function correctly with async/await for OpenTelemetry
    // See {@link https://opentelemetry.io/docs/languages/js/context/#active-context}

src/instrument.ts
```diff
@@ -93,6 +95,32 @@ interface traceConfig {
...
+    // but we shouldn't make our customers have to invasively edit their code
+    // nor should they be burdened about these facts, their code should JUST work.
+    // Please see https://github.com/googleapis/nodejs-spanner/issues/2146
```
alkatrivedi, line 116: remove

src/instrument.ts
alkatrivedi, line 113:
```suggestion
    // If no active context was set previously, trace context propagation cannot
    // function correctly with async/await for OpenTelemetry
    // See {@link https://opentelemetry.io/docs/languages/js/context/#active-context}
```

src/instrument.ts
alkatrivedi, line 116:
```suggestion
```

src/instrument.ts
```diff
@@ -93,6 +95,32 @@ interface traceConfig {
...
+    context.disable(); // Firstly disable any prior contextManager.
```
alkatrivedi, line 117:
```suggestion
    context.disable(); // Disable any prior contextManager.
```

observability-test/database.ts
```diff
@@ -375,6 +375,115 @@ describe('Database', () => {
     });
   });
 
+  describe('batchCreateSessions', () => {
+    it('without error', done => {
```
alkatrivedi, line 379:
```suggestion
    it('should execute batchCreateSessions without error', done => {
```

observability-test/database.ts
```diff
@@ -375,6 +375,115 @@ describe('Database', () => {
...
+          actualEventNames,
+          expectedEventNames,
+          `Unexpected events:\n\tGot:  ${actualEventNames}\n\tWant: ${expectedEventNames}`
+        );
+
+        done();
+      });
+    });
+
+    it('with error', done => {
```
alkatrivedi, line 433:
```suggestion
    it('should execute batchCreateSessions with error', done => {
```

observability-test/database.ts
```diff
@@ -375,6 +375,115 @@ describe('Database', () => {
...
+          firstSpan.status.code,
+          'Expected an ERROR span status'
+        );
+        assert.strictEqual(
+          'batchCreateSessions.error',
+          firstSpan.status.message,
+          'Mismatched span status message'
+        );
+
+        // We don't expect events.
```
alkatrivedi, line 474:
```suggestion
        // Ensure events are not present
```

observability-test/database.ts
```diff
@@ -375,6 +375,115 @@ describe('Database', () => {
...
+      });
+    });
+
+    it('with error', done => {
+      const ARGS = [new Error('batchCreateSessions.error'), null];
+      database.request = (config, callback) => {
+        callback(...ARGS);
+      };
+
+      database.batchCreateSessions(10, (err, sessions) => {
```
alkatrivedi, line 439: let's not pass the session count directly, better if we make a variable first and then pass that variable to batchCreateSessions

observability-test/database.ts
```diff
@@ -375,6 +375,115 @@ describe('Database', () => {
...
   });
 
+  describe('batchCreateSessions', () => {
+    it('without error', done => {
+      const ARGS = [null, [{}]];
+      database.request = (config, callback) => {
+        callback(...ARGS);
+      };
+
+      database.batchCreateSessions(10, (err, sessions) => {
```
alkatrivedi, line 385: let's not pass the session count directly, better if we make a variable first and then pass that variable to batchCreateSessions

src/database.ts
```diff
@@ -677,30 +682,36 @@ class Database extends common.GrpcServiceObject {
...
-        reqOpts,
-        gaxOpts: options.gaxOptions,
-        headers: headers,
-      },
-      (err, resp) => {
-        if (err) {
-          callback!(err, null, resp!);
-          return;
-        }
+    const q = {opts: this._observabilityOptions};
```
surbhigarg92, line 685: Modify the variable name from `q` to `traceConfig` as discussed in previous PR

src/session-pool.ts
```diff
@@ -485,6 +491,9 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     });
 
     this._traces = new Map();
+    if (!this._observabilityOptions) {
```
surbhigarg92, line 494: Why do we need this if check. We should simply override the observability option ?

src/session-pool.ts
```diff
@@ -752,41 +758,54 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
     let nReturned = 0;
     const nRequested: number = amount;
 
-    // while we can request as many sessions be created as we want, the backend
-    // will return at most 100 at a time, hence the need for a while loop.
-    while (amount > 0) {
-      let sessions: Session[] | null = null;
+    if (!this.database._observabilityOptions) {
```
surbhigarg92, line 761: Why do we need to set. the observability option here again ? We should be doing it in constructor ?

src/session-pool.ts
```diff
@@ -752,41 +758,54 @@ export class SessionPool extends EventEmitter implements SessionPoolInterface {
...
+          amount -= sessions.length;
+          nReturned += sessions.length;
+        } catch (e) {
+          this._pending -= amount;
+          this.emit('createError', e);
+          span.addEvent(
+            `Requested for ${nRequested} sessions returned ${nReturned}`
+          );
+          setSpanErrorAndException(span, e as Error);
+          throw e;
```
surbhigarg92, line 792: There should be a `span.end()` in the catch block ? before throwing error ? 
If this is true , also add a test case which could have caught this

observability-test/spanner.ts
```diff
@@ -815,8 +966,161 @@ describe('ObservabilityOptions injection and propagation', async () => {
...
+    const observabilityOptions: typeof ObservabilityOptions = {
+      tracerProvider: provider,
+      enableExtendedTracing: true,
+    };
+    const setupResult = await setup(observabilityOptions);
+    const spanner = setupResult.spanner;
+    const server = setupResult.server;
+    const spannerMock = setupResult.spannerMock;
+
+    after(async () => {
```
surbhigarg92, line 992: Is this `after` in the test method ? why ?

observability-test/spanner.ts
```diff
@@ -815,8 +966,161 @@ describe('ObservabilityOptions injection and propagation', async () => {
         expectedEventNames,
         `Unexpected events:\n\tGot:  ${actualEventNames}\n\tWant: ${expectedEventNames}`
       );
+    });
+  });
+});
 
-      done();
+describe('Regression tests for fixed bugs', () => {
```
surbhigarg92, line 973: use a better name to describe the test block. Maybe something like `E2E traces with async/await` ?

observability-test/spanner.ts
```diff
@@ -815,8 +966,161 @@ describe('ObservabilityOptions injection and propagation', async () => {
...
+  it('async/await correctly parents trace spans', async () => {
```
surbhigarg92, line 974: Can we also add a test case using "callback"

observability-test/database.ts
odeke-em: This feels pedantic/verbose and repeating also out of context because I am really testing what happens without an error returned and also with an error returned.

src/session-pool.ts
odeke-em: Good catch and thank you! I've added a check and a fix too.

observability-test/spanner.ts
odeke-em: But they are really regression tests, wouldn't E2E be even more obscure? Regression to catch any future bugs given an exact use case per bug.

observability-test/spanner.ts
odeke-em: Yes please, and done, thank you @surbhigarg92, please take a look!

src/session-pool.ts
odeke-em: Not really, I was being cautious in case someone had passed in multiple observability configurations one for the explicitly created SessionPool and then one for the Database separately. It doesn't hurt to be conservative at the beginning before these libraries have been put into heavy usage.

src/session-pool.ts
odeke-em: There was some test case in which the observability options for database weren't set despite having been passed into the Spanner constructor.

observability-test/spanner.ts
odeke-em: No longer relevant but really you are allowed to place an after anywhere, especially to capture nested scope variables.

src/database.ts
odeke-em: Indeed, thank you!

src/instrument.ts
odeke-em: Done, thank you @alkatrivedi!

src/instrument.ts
odeke-em: Done, thank you @alkatrivedi!

observability-test/spanner.ts
odeke-em: Done, thank you @alkatrivedi!

observability-test/database.ts
odeke-em: I don't agree, this is an inlined and singly used value, creating a variable for such a case seems unnecessary and pedantic.

observability-test/database.ts
odeke-em: We aren't testing Cloud Spanner's code but rather the instrumentation added and what happens with and without errors.

observability-test/database.ts
odeke-em: Ditto about that being pedantic yet this is a single usage.

observability-test/spanner.ts
odeke-em: Also btw within this section I've included the callback variant of that bug reproducer

observability-test/spanner.ts
```diff
@@ -1010,52 +1010,43 @@ describe('ObservabilityOptions injection and propagation', async () => {
   });
 });
 
-describe('Regression tests for fixed bugs', () => {
-  it('async/await correctly parents trace spans', async () => {
-    // See https://github.com/googleapis/nodejs-spanner/issues/2146.
-    const traceExporter = new InMemorySpanExporter();
-    const provider = new NodeTracerProvider({
+describe('Regression tests for fixed bugs', async () => {
```
surbhigarg92, line 1013: We can not have test block name as 'Regression tests for fixed bugs' . Block should tell what the scope of testing

src/instrument.ts
surbhigarg92: I think we should remove this code. As we have added the opentelemetry link which should be enough. The bug will be anyway linked so maintainers would be able to track it using the code blame if required.

src/session-pool.ts
surbhigarg92: But we don't allow customers to pass multiple observability configurations right ? They should be able to set it only once via spannerOptions. 

We also modified the _observabilityOptions object to add _ to represent private variable.

src/session-pool.ts
surbhigarg92: Then we should find a way to fix those test cases and not add the code here again.

observability-test/spanner.ts
odeke-em: I think I'll defer to you then for the naming as you are the maintainer; this because for other projects we usually have explicit sections for regression tests where it is highly encouraged that you post the reproducers. What name would you like for this?

observability-test/session-pool.ts
```diff
@@ -0,0 +1,222 @@
...
+import * as proxyquire from 'proxyquire';
+import * as sinon from 'sinon';
+import stackTrace = require('stack-trace');
+const {
+  AlwaysOnSampler,
+  NodeTracerProvider,
+  InMemorySpanExporter,
+} = require('@opentelemetry/sdk-trace-node');
+// eslint-disable-next-line n/no-extraneous-require
+const {SimpleSpanProcessor} = require('@opentelemetry/sdk-trace-base');
```
surbhigarg92, line 30: Why don't we add `@opentelemetry/sdk-trace-base` as a test dependency and remove this lint from everywhere in code

surbhigarg92 (approved):

observability-test/session-pool.ts
odeke-em: @surbhigarg92 it is added in `devDependencies`

surbhigarg92: then why do we need to add this eslint suppress ?

odeke-em: Without it we were getting build failures, it's basically the template for the tests in observability-test from long when we started sending in code changes. I really didn't dig in as it required so much time going through and figuring the build system's problems and it wasn't worth the hours sweating on.

PR #2147, opened 2024-10-04
fix: (observability) make async/await correctly work by setting initial AsyncHooksManager

### LINKED ISSUE #2146: observability: if an async context manager has not been enabled, make sure to register one

odeke-em:
OpenTelemetry cannot work correctly for async/await if there isn't a set AsyncHooksManager, but we should not burden our customers with this type of specialist knowledge, their code should just work and this change performs such a check. Later on we shall file a feature request with the OpenTelemetry-JS API group to give us a hook to detect if we've got a live asyncHooksManager instead of this mandatory comparison to ROOT_CONTEXT each time.

Fixes #2146

### DIFF
--- a/observability-test/spanner.ts
+++ b/observability-test/spanner.ts
@@ -617,8 +617,9 @@ describe('ObservabilityOptions injection and propagation', async () => {
   });
 });
 
-describe('Bug fixes', () => {
+describe('Regression tests for fixed bugs', () => {
   it('async/await correctly parents trace spans', async () => {
+    // See https://github.com/googleapis/nodejs-spanner/issues/2146.
     const traceExporter = new InMemorySpanExporter();
     const provider = new NodeTracerProvider({
       sampler: new AlwaysOnSampler(),
--- a/observability-test/spanner.ts
+++ b/observability-test/spanner.ts
@@ -37,6 +37,9 @@ const {
 
 const {ObservabilityOptions} = require('../src/instrument');
 
+const selectSql = 'SELECT 1';
+const updateSql = 'UPDATE FOO SET BAR=1 WHERE BAZ=2';
+
 /** A simple result set for SELECT 1. */
 function createSelect1ResultSet(): protobuf.ResultSet {
   const fields = [
@@ -141,8 +144,6 @@ describe('EndToEnd', () => {
       server = setupResult.server;
       spannerMock = setupResult.spannerMock;
 
-      const selectSql = 'SELECT 1';
-      const updateSql = 'UPDATE FOO SET BAR=1 WHERE BAZ=2';
       spannerMock.putStatementResult(
         selectSql,
         mock.StatementResult.resultSet(createSelect1ResultSet())
@@ -223,7 +224,13 @@ describe('EndToEnd', () => {
             `span names mismatch:\n\tGot:  ${actualSpanNames}\n\tWant: ${expectedSpanNames}`
           );
 
-          const expectedEventNames = [];
+          const expectedEventNames = [
+            'Acquiring session',
+            'Waiting for a session to become available',
+            'Acquired session',
+            'Begin Transaction',
+            'Transaction Creation Done',
+          ];
           assert.deepStrictEqual(
             actualEventNames,
             expectedEventNames,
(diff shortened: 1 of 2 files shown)


odeke-em:
Kindly help me run the bots @surbhigarg92 @alkatrivedi @harshachinta.

src/instrument.ts
```diff
@@ -93,6 +95,30 @@ interface traceConfig {
...
 
+const {
+  AsyncHooksContextManager,
+} = require('@opentelemetry/context-async-hooks');
+
+/*
+ * This function ensures that async/await functions correctly by
+ * checking if context.active() returns an invalid/unset context
+ * and if so, sets a global AsyncHooksContextManager.
+ */
```
surbhigarg92, line 106:
```suggestion
/**
 * Ensures that an appropriate Context Manager is set to support proper
 * context propagation when using async/await, particularly in OpenTelemetry.
 * 
 * OpenTelemetry's context system relies on `context.active()` to retrieve the 
 * current active context. However, if this returns the default `ROOT_CONTEXT`, 
 * it indicates that no context has been previously set. Without a valid context, 
 * async/await calls will fail to propagate context correctly, which can cause 
 * issues in distributed tracing and telemetry data collection.
 * 
 * This function checks if the active context is invalid (i.e., `ROOT_CONTEXT`). 
 * If so, it resets the Context Manager by disabling any prior manager and then 
 * enabling a new instance of `AsyncHooksContextManager` to properly track async
 * operations and context propagation.
 * 
 * The goal is to ensure that users' code works seamlessly with OpenTelemetry, 
 * without requiring them to modify their existing code to manually manage context.
 * 
 * Further details:
 * - OpenTelemetry context documentation: https://opentelemetry.io/docs/languages/js/context/#active-context
 * - Related issue: https://github.com/googleapis/nodejs-spanner/issues/2146
 */
```

src/instrument.ts
```diff
@@ -93,6 +95,30 @@ interface traceConfig {
...
+function ensureInitialContextManagerSet() {
+  if (context.active() === ROOT_CONTEXT) {
+    // If no active context was set previously, trace context propagation cannot
+    // work correctly with async/await for OpenTelemetry and they acknowledge
+    // this fact per https://opentelemetry.io/docs/languages/js/context/#active-context
+    // but we shouldn't make our customers have to invasively edit their code
+    // nor should they be burdened about these facts, their code should JUST work.
+    // Please see https://github.com/googleapis/nodejs-spanner/issues/2146
```
surbhigarg92, line 114:
```suggestion
    // No active context set, so context propagation won't work with async/await.
    // Fix by disabling any prior context manager and setting a new AsyncHooksContextManager.
    
```

src/instrument.ts
surbhigarg92: @odeke-em I generated this documentation using AI, please review once more. But we need a more structured documentation

observability-test/spanner.ts
```diff
@@ -583,3 +616,118 @@ describe('ObservabilityOptions injection and propagation', async () => {
     });
   });
 });
+
+describe('Bug fixes', () => {
```
surbhigarg92, line 620: Use a better name for this. If you are testing "async/await" use name like `Observability tests with async/await`

src/instrument.ts
odeke-em: Some of this AI generated documentation is not correct like 
```
 * The goal is to ensure that users' code works seamlessly with OpenTelemetry, 
* without requiring them to modify their existing code to manually manage context.
```	

it's nothing about manually managing context but setting a global context manager.

observability-test/spanner.ts
odeke-em: This is the section in which we shall be inserting all regression tests and bug fixes.

odeke-em: I tweaked it, please take a look.

odeke-em:
Kindly help me run the bots @alkatrivedi @surbhigarg92 @harshachinta

odeke-em:
I've instead moved the fix into one PR https://github.com/googleapis/nodejs-spanner/pull/2145 because it takes 3+ hours just to run tests, 1+ day for 1 PR code reviews but we don't have much time.

PR #2155, opened 2024-10-09
feat: (observability) propagate database name for every span generated to aid in quick debugging

odeke-em:
With this change customers shall always be able to identify which database is being connected to.

Updates #2079

## Exhibit
Screenshot 2024-10-08 at 10 52 56 PM

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -109,6 +109,7 @@ import {
   startTrace,
   setSpanError,
   setSpanErrorAndException,
+  traceConfig,
 } from './instrument';
 
 export type GetDatabaseRolesCallback = RequestCallback<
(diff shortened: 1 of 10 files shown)


src/database.ts
```diff
@@ -828,7 +828,10 @@ class Database extends common.GrpcServiceObject {
         ? (optionsOrCallback as TimestampBounds)
         : {};
 
-    const traceConfig = {opts: this._observabilityOptions};
+    const traceConfig = {
```
surbhigarg92, line 831: I really believe that this `traceConfig` creation should move to constructor where we can add all the base configuration and extend it where ever required. Just in this case we had to make code changes at multiple places , it would have been very simple if we would have this in constructor. WDYT ?

surbhigarg92 (approved):

src/database.ts
odeke-em: Sure, I've moved it to the constructor. I would have preferred to do that in another PR but we are already here so let's get it out. Thank you and done!

odeke-em: I wired that up for src/database.ts but doing the same for all of them is quite a hassle and a needle in the haystack that I don't think we should spend time on right now. Kindly please take a look at the latest code.

PR #2156, opened 2024-10-10
feat(observability)): trace Session.create+delete

odeke-em:
This change traces Session methods:
* create
* delete

Updates #2087


surbhigarg92:
We dont need this PR

odeke-em:
@surbhigarg92 we spoke about it but I wanted to ensure an entry for it, given that the checklist asks for it.

PR #2168, opened 2024-10-21
fix: Remove Begin transaction on non retryable error

surbhigarg92:


src/transaction.ts
```diff
@@ -743,21 +743,6 @@ export class Snapshot extends EventEmitter {
...
-            this._useInRunner &&
-            !(
-              isServiceError &&
-              (err as grpc.ServiceError).code === grpc.status.ABORTED
-            )
-          ) {
-            this.begin();
-          }
-          setSpanError(span, err);
-        })
```
odeke-em, line 760: I don't believe that the right fix is to delete this code but rather to check that the error is retryable by checking if these grpc error codes are returned:
* DEADLINE_EXCEEDED
* UNAVAILABLE
* INTERNAL

although I guess the next step is to ask after this.begin() what happens next?

PR #2188, opened 2024-11-12
fix(observability): fix the spans for the abort transaction

surbhigarg92:
This PR fixes spans for the `runTransaction` for error cases and removes the `Database.run` span.


src/database.ts
```diff
@@ -2110,14 +2110,14 @@ class Database extends common.GrpcServiceObject {
...
             } else {
-              span.addEvent('Using Session', {'session.id': session?.id});
               this.pool_.release(session!);
               span.end();
               callback!(err);
             }
             return;
           }
 
+          span.addEvent('Using Session', {'session.id': session?.id});
```
sakthivelmanii, line 2120: nit: can we create a const file and put all these event names

sakthivelmanii (approved):

PR #2191, opened 2024-11-13
chore: add support for class MultiplexedSession

alkatrivedi:
This PR contains the code for the newly introduced class `MultiplexedSession`.


src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+  _pingHandle!: NodeJS.Timer;
+  _refreshHandle!: NodeJS.Timer;
+  constructor(
+    database: Database,
+    multiplexedSessionOptions?: MultiplexedSessionOptions
+  ) {
+    super();
+    this.database = database;
+    this.options = Object.assign({}, DEFAULTS, multiplexedSessionOptions);
+    this.options.databaseRole = this.options.databaseRole
```
surbhigarg92, line 119: Where is this `databaseRole` used ? In session pool its used for `batchCreateSession` .  I believe this needs to be passed to `this.database.createSession`

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+
+  /**
+   * Creates a transaction for a session.
+   *
+   * @private
+   *
+   * @param {Session} session The session object.
+   * @param {object} options The transaction options.
+   */
+  _prepareTransaction(session: Session | null): void {
```
surbhigarg92, line 202: This is again a duplicate method ?

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+   * @throws {Error} If there is an issue with retrieving the session metadata or calculating the expiration time.
+   */
+  async _refresh(): Promise<void> {
+    const metadata = await this._muxSession.multiplexedSession?.getMetadata();
+    const createTime =
+      parseInt(metadata![0].createTime.seconds) * 1000 +
+      metadata![0].createTime.nanos / 1000000;
+
+    // Calculate expiration time (7 days after session creation)
+    const expireTime = createTime + 7 * 24 * 60 * 60 * 1000;
```
surbhigarg92, line 226: Instead of 7, we should be using `refreshRate` ?

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+      const session = await this._getSession();
+
+      if (session) {
+        return session;
+      }
+
+      return getSession();
+    };
+
+    const session = await this._acquires.add(getSession);
```
surbhigarg92, line 270: I didn't get this. Why do we need a queue if its a single session ? 

We could simply do `const session = await getSession();`

surbhigarg92: Where else are we using this queue ?

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  /**
+   * Retrieves the current multiplexed session.
+   *
+   * Returns the current multiplexed session associated with this instance.
+   *
+   * @returns {Session | null} The current multiplexed session if available, or `null` if no session is present.
+   */
+  _multiplexedSession(): Session | null {
```
surbhigarg92, line 282: You don't need this method, you could directly use the private property in the same class.

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+   *
+   * @private
+   */
+  async _getSession(): Promise<Session | null> {
+    // Check if the multiplexed session is already available
+    if (this._muxSession.multiplexedSession !== null) {
+      return this._multiplexedSession();
+    }
+
+    // Define event and promises to wait for the session to become available or for an error
```
surbhigarg92, line 304: We should also add a timeout logic here. Something like below . In MultiplexedSessionOptions get the value for `timeout` default would be infinity. Check this logic in other languages as well
```
let removeTimeoutListener = () => {};
    if (!is.infinite(timeout!)) {
      const elapsed = Date.now() - startTime!;
      const remaining = timeout! - elapsed;

      promises.push(
        new Promise((_, reject) => {
          const error = new Error(errors.Timeout);
          const timeoutFunction = setTimeout(
            reject.bind(null, error),
            remaining
          );
          removeTimeoutListener = () => clearTimeout(timeoutFunction);
        })
      );
    }
```

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+      .then(() => {
+        this._maintain();
+      })
+      .catch(err => {
+        if (
+          isDatabaseNotFoundError(err) ||
+          isInstanceNotFoundError(err) ||
+          isCreateSessionPermissionError(err) ||
+          isDefaultCredentialsNotSetError(err) ||
+          isProjectIdNotSetInEnvironmentError(err)
```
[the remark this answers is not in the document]
alkatrivedi: Done, PR link - https://github.com/googleapis/nodejs-spanner/pull/2194

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+import {Transaction} from './transaction';
+import {
+  isDatabaseNotFoundError,
+  isInstanceNotFoundError,
+  isCreateSessionPermissionError,
+  isDefaultCredentialsNotSetError,
+  isProjectIdNotSetInEnvironmentError,
+} from './session-pool';
+
+interface MuxSession {
```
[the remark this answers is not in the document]
alkatrivedi: Removed

src/multiplexed-session.ts
alkatrivedi: `this.database.createSession` is utilizing the database role same as the one passed via pool options, we have
```
if (typeof poolOptions === 'object') {
      this.databaseRole = poolOptions.databaseRole || null;
    }
```
in the database class constructor and this in
```
reqOpts.session.creatorRole =
      options.databaseRole || this.databaseRole || null;
```
      createSession method
      now since we don't have any such thing MultiplexedSessionOptions in other languages and are not exporting MultiplexedSessionOptions to our customers, hence I am removing the MultiplexedSessionOptions from this class as well

src/multiplexed-session.ts
alkatrivedi: Updated, thanks!

src/multiplexed-session.ts
alkatrivedi: updated

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * Multiplexed session configuration options.
+ *
+ * @typedef {object} MultiplexedSessionOptions
+ * @property {number} [refreshRate=7] How often to check for expiration of multiplexed session, in
+ *     days. Must not be greater than 7 days.
+ */
+
+export interface MultiplexedSessionOptions {
```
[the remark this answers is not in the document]
alkatrivedi: Yeah, in case of mux session we not exposing any options to our customers. I have removed the MultiplexedSessionOptions instead now I am using a refreshRate property inside the class and assign the default value to it in the constructor

src/multiplexed-session.ts
```diff
@@ -0,0 +1,320 @@
...
+  multiplexedSession: Session | null;
+}
+
+/**
+ * @callback GetSessionCallback
+ * @param {?Error} error Request error, if any.
+ * @param {Session} session The read-write session.
+ * @param {Transaction} transaction The transaction object.
+ */
+export interface GetSessionCallback {
```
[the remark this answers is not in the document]
alkatrivedi: For now I have abstracted the interface from the SessionPool class, will move it to our common class GetSession later in other PR

src/multiplexed-session.ts
alkatrivedi: Since, this method is part of SessionPool class and is private; to use it in this MultiplexedSession class we need to create an object of the SessionPool class. Hence, I will move this method in the GetSession class later in the other PR. So, that from that common class both the SessionPool and MultiplexedSession can use it

src/multiplexed-session.ts
alkatrivedi: we don't need it, since in case of unimplemented we are throwing error to the customers

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+      const error = {
+        code: Status.NOT_FOUND,
+        message: 'Database not found',
+      } as grpc.ServiceError;
+      sandbox.stub(multiplexedSession, '_createSession').rejects(error);
+      try {
+        multiplexedSession.createSession();
+      } catch (e) {
+        shouldNotBeCalled();
+      }
```
surbhigarg92, line 125: We could replace this with 

`assert.doesNotThrow(() => multiplexedSession.createSession());`

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+        code: Status.NOT_FOUND,
+        message: 'Unable to detect a Project Id in the current environment',
+      } as grpc.ServiceError;
+      sandbox.stub(multiplexedSession, '_createSession').rejects(error);
+      try {
+        multiplexedSession.createSession();
+      } catch (e) {
+        shouldNotBeCalled();
+      }
+    });
```
surbhigarg92, line 153: These tests have code duplication. You can have a single test with a loop for all message. For eg: https://github.com/googleapis/nodejs-spanner/blob/main/test/spanner.ts#L2769

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+
+      await errorPromise;
+    });
+  });
+
+  describe('_maintain', () => {
+    it('should set an interval to refresh mux sessions', done => {
+      const expectedInterval =
+        multiplexedSession.refreshRate! * 24 * 60 * 60000;
+      const clock = sandbox.useFakeTimers();
```
surbhigarg92, line 179: make sure to clear the clock after the test is over. Probably you can do below 
let clock;
beforeEach(() => {
  clock = sandbox.useFakeTimers();
});
afterEach(() => {
  clock.restore();
});

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+    it('should create a session object', () => {
+      assert.deepStrictEqual(multiplexedSession._multiplexedSession, null);
+    });
+
+    it('should inherit from EventEmitter', () => {
+      assert(multiplexedSession instanceof events.EventEmitter);
+    });
+  });
+
+  describe('createSession', () => {
```
surbhigarg92, line 86: We can add a extra edge case for test when createSession is called multiple times consecutively.

src/multiplexed-session.ts
```diff
@@ -0,0 +1,241 @@
...
+    this.refreshRate = 7;
+    this._multiplexedSession = null;
+  }
+
+  /**
+   * Creates a new multiplexed session and manages its maintenance.
+   *
+   * This method initiates the session creation process by calling the `_createSession` method, which returns a Promise.
+   */
+  createSession(): void {
```
surbhigarg92, line 88: A thought , can we make this method to return `Promise<void>` . This will allow us to use await with this method and if we dont want to await we can simply call this without await.

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+      assert.strictEqual(multiplexedSession.refreshRate, 7);
+    });
+
+    it('should create a session object', () => {
+      assert.deepStrictEqual(multiplexedSession._multiplexedSession, null);
+    });
+
+    it('should inherit from EventEmitter', () => {
+      assert(multiplexedSession instanceof events.EventEmitter);
+    });
```
surbhigarg92, line 83: nit: Club this instantiation into single test.

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+      _maintainStub = sandbox.stub(multiplexedSession, '_maintain');
+      _createSessionStub = sandbox
+        .stub(multiplexedSession, '_createSession')
+        .resolves();
+    });
+
+    it('should create mux session', () => {
+      multiplexedSession.createSession();
+      assert.strictEqual(_createSessionStub.callCount, 1);
+    });
```
surbhigarg92, line 100: You could stub `_createSession` to return a session not only resolves and verify the multiplexedSession property.

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+      const multiplexed = {multiplexed: true};
+      await multiplexedSession._createSession();
+      assert.deepStrictEqual(createSessionStub.lastCall.args[0], multiplexed);
+    });
+
+    it('should not make multiple requests', async () => {
+      createSessionStub.onCall(0).resolves([[{}, {}]]);
+      createSessionStub.onCall(1).resolves([[{}]]);
+
+      await multiplexedSession._createSession();
```
surbhigarg92, line 207: Did you mean to make multiple calls to `_createSession` here ?

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+    it('should create the mux sessions', async () => {
+      await multiplexedSession._createSession();
+      assert.strictEqual(createSessionStub.callCount, 1);
+    });
+
+    it('should pass the multiplexed option', async () => {
+      const multiplexed = {multiplexed: true};
+      await multiplexedSession._createSession();
+      assert.deepStrictEqual(createSessionStub.lastCall.args[0], multiplexed);
+    });
```
surbhigarg92, line 201:
```suggestion
    it('should create the mux sessions with multiplexed option', async () => {
      await multiplexedSession._createSession();
      assert.strictEqual(createSessionStub.callCount, 1);
      assert.deepStrictEqual(createSessionStub.lastCall.args[0], {multiplexed: true});
    });
```

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+    it('should acquire a session', done => {
+      sandbox.stub(multiplexedSession, '_acquire').resolves(fakeMuxSession);
+
+      multiplexedSession.getSession((err, session) => {
+        assert.ifError(err);
+        assert.strictEqual(session, fakeMuxSession);
+        done();
+      });
+    });
+    it('should pass any errors to the callback', done => {
```
surbhigarg92, line 235:
```suggestion

    it('should pass any errors to the callback', done => {
```

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+      const error = new Error('err');
+      sandbox.stub(multiplexedSession, '_acquire').rejects(error);
+      multiplexedSession.getSession(err => {
+        assert.strictEqual(err, error);
+        done();
+      });
+    });
+    it('should pass back the session and txn', done => {
```
surbhigarg92, line 243:
```suggestion

    it('should pass back the session and txn', done => {
```

test/multiplexed-session.ts
```diff
@@ -0,0 +1,326 @@
...
+
+    it('should have the multiplexed property set to true', async () => {
+      sandbox.stub(multiplexedSession, '_getSession').resolves(fakeMuxSession);
+      const session = await multiplexedSession._acquire();
+      assert.strictEqual((await session!.getMetadata()).multiplexed, true);
+      assert.strictEqual(
+        (await fakeMuxSession!.getMetadata()).multiplexed,
+        true
+      );
+    });
```
surbhigarg92, line 274: I feel this test case is not required, as we are just verifying the `getMetadata` stub here, nothing else.

alkatrivedi: certainly, but we are also verifying if the session is containing the multiplexed field enabled to true

src/multiplexed-session.ts
```diff
@@ -0,0 +1,241 @@
...
+   * `mux-session-available` will be emitted to signal that the session is ready.
+   *
+   * In case of error, an error will get emitted along with the erorr event.
+   *
+   * @private
+   */
+  async _createSession(): Promise<void> {
+    try {
+      const [createSessionResponse] = await this.database.createSession({
+        multiplexed: true,
```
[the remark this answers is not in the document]
harshachinta: Yes we will need them while creating any type of sessions. 
https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java#L243-L250

src/multiplexed-session.ts
```diff
@@ -0,0 +1,241 @@
...
+   * Every 7th day a new Multiplexed Session will get created. However, a Multiplexed Session will
+   * be alive for 30 days in the backend.
+   *
+   * @returns {Promise<void>} A Promise that resolves once the refresh process is completed.
+   *                          If the session is expired, a new session will be created.
+   *
+   * @throws {Error} If there is an issue with retrieving the session metadata or calculating the expiration time.
+   */
+  async _refresh(): Promise<void> {
+    const metadata = await this._multiplexedSession?.getMetadata();
```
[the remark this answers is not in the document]
harshachinta: Yes we can store it in multiplexed session class.
In Java we store the expiration time of session, instead of creation time to simplify checks
https://github.com/googleapis/java-spanner/blob/b545557b1a27868aeb5115b3947d42db015cc00e/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java#L183

src/multiplexed-session.ts
```diff
@@ -0,0 +1,241 @@
...
+    const createTime =
+      parseInt(metadata![0].createTime.seconds) * 1000 +
+      metadata![0].createTime.nanos / 1000000;
+
+    // Calculate expiration time (7 days after session creation)
+    const expireTime = createTime + this.refreshRate * 24 * 60 * 60 * 1000;
+
+    // If the current time exceeds the expiration time, create a new session
+    if (Date.now() > expireTime) {
```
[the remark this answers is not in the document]
alkatrivedi: Yeah, we don't need this check here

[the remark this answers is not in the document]
alkatrivedi: We don't even need this _refresh method since, we are sure that the interval is 7 days after the mux session creation, I have modified the _maintain method to call directly _createSession in a interval of 7 days

test/multiplexed-session.ts
alkatrivedi: Updated, thanks for the suggestion!

src/multiplexed-session.ts
```diff
@@ -0,0 +1,241 @@
...
+   * @private
+   */
+  async _getSession(): Promise<Session | null> {
+    // Check if the multiplexed session is already available
+    if (this._multiplexedSession !== null) {
+      return this._multiplexedSession;
+    }
+
+    // Define event and promises to wait for the session to become available
+    const availableEvent = 'mux-session-available';
```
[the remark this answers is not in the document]
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: done

alkatrivedi: there some modification needed in the test, I have modified it.
No, I mean the multiple calls to createSession method of the database class here, since _createSession is calling that method

test/multiplexed-session.ts
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: there is already a test where we are verifying the multiplexed property for the session
`should have the multiplexed property set to true`

test/multiplexed-session.ts
alkatrivedi: this test is removed based on discussion

test/multiplexed-session.ts
alkatrivedi: based on offline discussion, we are skipping this test

src/database.ts
```diff
@@ -994,24 +994,28 @@ class Database extends common.GrpcServiceObject {
...
+        },
+        (err, resp) => {
+          if (err) {
+            setSpanError(span, err);
+            span.end();
+            callback(err, null, resp!);
+            return;
+          }
+          const session = this.session(resp!.name!);
+          session.metadata = resp;
```
surbhigarg92, line 1014: add `span.end()` before callback

surbhigarg92 (approved):

src/database.ts
alkatrivedi: done

src/multiplexed-session.ts
```diff
@@ -0,0 +1,254 @@
...
+        span.addEvent('Requesting a multiplexed session');
+        try {
+          const [createSessionResponse] = await this.database.createSession({
+            multiplexed: true,
+          });
+          this._multiplexedSession = createSessionResponse;
+          span.addEvent(
+            `Created multiplexed session ${this._multiplexedSession.formattedName_}`
+          );
+          this.emit('mux-session-available');
```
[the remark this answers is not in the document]
alkatrivedi: done

src/multiplexed-session.ts
```diff
@@ -0,0 +1,254 @@
...
+   * @returns {Promise<void>} A Promise that resolves when the session has been successfully created and assigned, an event
+   * `mux-session-available` will be emitted to signal that the session is ready.
+   *
+   * In case of error, an error will get emitted along with the erorr event.
+   *
+   * @private
+   */
+  async _createSession(): Promise<void> {
+    const traceConfig = {
+      opts: this._observabilityOptions,
```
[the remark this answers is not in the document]
alkatrivedi: yeah, correct the value was not assigned. I have updated the code, now the value is getting assign in the constructor

src/multiplexed-session.ts
```diff
@@ -0,0 +1,254 @@
...
+ */
+export class MultiplexedSession
+  extends EventEmitter
+  implements MultiplexedSessionInterface
+{
+  database: Database;
+  // frequency to create new mux session
+  refreshRate: number;
+  _multiplexedSession: Session | null;
+  _pingHandle!: NodeJS.Timer;
```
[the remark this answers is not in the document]
alkatrivedi: no where, I have removed it. thanks!

src/multiplexed-session.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  /**
+   * Maintains the multiplexed session by periodically refreshing it.
+   *
+   * This method sets up a periodic refresh interval for maintaining the session. The interval duration
+   * is determined by the @param refreshRate option, which is provided in days.
+   * The default value is 7 days.
+   *
+   * @returns {void} This method does not return any value.
```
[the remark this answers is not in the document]
alkatrivedi: Updated the doc, thanks!

src/multiplexed-session.ts
```diff
@@ -0,0 +1,254 @@
...
+      this.once(MUX_SESSION_AVAILABLE, resolve);
+      removeListener = this.removeListener.bind(
+        this,
+        MUX_SESSION_AVAILABLE,
+        resolve
+      );
+    });
+
+    try {
+      await promise;
```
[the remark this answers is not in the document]
alkatrivedi: yes, the error will get surfaced to the customers

test/multiplexed-session.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    it('should have the multiplexed property set to true', async () => {
+      sandbox.stub(multiplexedSession, '_getSession').resolves(fakeMuxSession);
+      const session = await multiplexedSession._acquire();
+      assert.strictEqual(session.multiplexed, true);
+      assert.strictEqual(fakeMuxSession.multiplexed, true);
+    });
+  });
+
+  describe('_getSession', () => {
```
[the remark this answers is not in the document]
alkatrivedi: Since, we are listening to the error event in _getSession in case of the error from the _createSession, I have added a test for the case when the error event will get emitted, if we are able to listen to that event in _getSession

src/multiplexed-session.ts
```diff
@@ -0,0 +1,254 @@
...
+    if (this._multiplexedSession !== null) {
+      span.addEvent('Cache hit: has usable multiplexed session');
+      return this._multiplexedSession;
+    }
+
+    // Define event and promises to wait for the session to become available
+    span.addEvent('Waiting for a multiplexed session to become available');
+    let removeListener: Function;
+    const promise = new Promise(resolve => {
+      this.once(MUX_SESSION_AVAILABLE, resolve);
```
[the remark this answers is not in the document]
alkatrivedi: Thanks for pointing this out.
>> What happens if this event is never emitted?
In case MUX_SESSION_AVAILABLE event will not get emitted, an error event MUX_SESSION_CREATE_ERROR will get emitted.
Hence, in _getSession we either will wait for the MUX_SESSION_AVAILABLE or for the MUX_SESSION_CREATE_ERROR.
>> Is there any timeout?
There is no any timeout in the method, we are relying on the events

src/multiplexed-session.ts
```diff
@@ -0,0 +1,255 @@
...
+      traceConfig,
+      async span => {
+        span.addEvent('Requesting a multiplexed session');
+        try {
+          const [createSessionResponse] = await this.database.createSession({
+            multiplexed: true,
+          });
+          this._multiplexedSession = createSessionResponse;
+          span.addEvent(
+            `Created multiplexed session ${this._multiplexedSession.formattedName_}`
```
surbhigarg92, line 130: How this `${this._multiplexedSession.formattedName_}` will look like ? I don't think you should ad the name, instead add the `session.id` in span.

test/multiplexed-session.ts
```diff
@@ -0,0 +1,289 @@
...
+        try {
+          assert.strictEqual(_maintainStub.callCount, 1);
+          done();
+        } catch (err) {
+          done(err);
+        }
+      });
+    });
+
+    it('should propagate errors for Multiplexed Session which gets emitted', async () => {
```
surbhigarg92, line 105: There is no assert in this test ?

test/multiplexed-session.ts
```diff
@@ -0,0 +1,289 @@
...
+  describe('_maintain', () => {
+    let clock;
+    let createSessionStub;
+    beforeEach(() => {
+      createSessionStub = sandbox
+        .stub(multiplexedSession, '_createSession')
+        .resolves();
+      clock = sandbox.useFakeTimers();
+    });
+    afterEach(() => {
```
surbhigarg92, line 134:
```suggestion

    afterEach(() => {
```

test/multiplexed-session.ts
```diff
@@ -0,0 +1,289 @@
...
+      clock.restore();
+    });
+    it('should set an interval to refresh mux sessions', () => {
```
surbhigarg92, line 137:
```suggestion

    it('should set an interval to refresh mux sessions', () => {
```

test/multiplexed-session.ts
```diff
@@ -0,0 +1,289 @@
...
+      await multiplexedSession._createSession();
+      assert.strictEqual(createSessionStub.callCount, 1);
+      assert.deepStrictEqual(createSessionStub.lastCall.args[0], {
+        multiplexed: true,
+      });
+    });
+
+    it('should emit the MUX_SESSION_AVAILABLE event on successfully creating mux session', done => {
+      multiplexedSession.on(MUX_SESSION_AVAILABLE, () => {
+        done();
```
surbhigarg92, line 158: Add assert like below 
`assert.strictEqual(multiplexedSession._multiplexedSession, "whatever session is stubbed")` 

or maybe check that its not null

test/multiplexed-session.ts
```diff
@@ -0,0 +1,289 @@
...
+      }
+    });
+
+    it('should emit the error event on failed creation of mux session', done => {
+      const error = new Error(MUX_SESSION_CREATE_ERROR);
+      createSessionStub.rejects(error);
+      multiplexedSession.on(MUX_SESSION_CREATE_ERROR, () => {
+        done();
+      });
+      multiplexedSession._createSession().catch(err => {
```
surbhigarg92, line 181: Here will .on event called first ? In that case you are calling `done` here and then .catch will never be called ?

alkatrivedi: it will call the done as soon as the error event will get emitted, after that the pointer will go to `throw e` in getSession hence, we need to catch that error here. 

 >> .catch will never be called ?
 
 it will get called, on not calling .catch here will result into the unhandled exception error

harshachinta (approved):

harshachinta:
Marking it as `chore:` since we are not explicitly providing any functionality here for the customers to use?
Please feel free to modify this if that is not true.
@surbhigarg92 @alkatrivedi

src/multiplexed-session.ts
alkatrivedi: >>How this ${this._multiplexedSession.formattedName_} will look like ?

it looks like projects/my-project/instances/my-instance/databases/my-database/sessions/my-session

>> How this ${this._multiplexedSession.formattedName_} will look like ? I don't think you should ad the name, instead add the session.id in span.

sure, I have updated the `${this._multiplexedSession.formattedName_}` to `${this._multiplexedSession.id}`

src/multiplexed-session.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    // Define event and promises to wait for the session to become available
+    span.addEvent('Waiting for a multiplexed session to become available');
+    let removeAvailableListener: Function;
+    let removeErrorListener: Function;
+    const promises = [
+      new Promise((_, reject) => {
+        this.once(MUX_SESSION_CREATE_ERROR, reject);
+        removeErrorListener = this.removeListener.bind(
```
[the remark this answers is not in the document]
alkatrivedi: No, the _createSession call will emit the error along with the event, which the listener will listen in _getSession. Now, we have two promises in _getSession, for the error case the promise for error event will be there which on begin await will resolve into a rejected promise. hence, an error will get propagated to the parent call

src/multiplexed-session.ts
```diff
@@ -0,0 +1,255 @@
...
+    );
+  }
+
+  /**
+   * Maintains the multiplexed session by periodically refreshing it.
+   *
+   * This method sets up a periodic refresh interval for maintaining the session. The interval duration
+   * is determined by the @param refreshRate option, which is provided in days.
+   * The default value is 7 days.
+   * @throws {Error} In case the multiplexed session creation will get fail, and an error occurs within `_createSession`,
```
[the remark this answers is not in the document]
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: done

src/multiplexed-session.ts
```diff
@@ -0,0 +1,255 @@
...
+  /**
+   * Creates a new multiplexed session.
+   *
+   * This method sends a request to the database to create a new session with multiplexing enabled.
+   * The response from the database would be an array, the first value of the array will be containing the multiplexed session.
+   *
+   * @returns {Promise<void>} A Promise that resolves when the session has been successfully created and assigned, an event
+   * `mux-session-available` will be emitted to signal that the session is ready.
+   *
+   * In case of error, an error will get emitted along with the erorr event.
```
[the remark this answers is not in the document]
alkatrivedi: done

src/multiplexed-session.ts
```diff
@@ -0,0 +1,255 @@
...
+    );
+    return session;
+  }
+
+  /**
+   * Attempts to get a session, waiting for it to become available if necessary.
+   *
+   * Waits for the `MUX_SESSION_AVAILABLE` event or for the `MUX_SESSION_CREATE_ERROR`
+   * to be emitted if the multiplexed session is not yet available. The method listens
+   * for these events, and once `mux-session-available` is emitted, it resolves and returns
```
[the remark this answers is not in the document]
alkatrivedi: done

test/multiplexed-session.ts
alkatrivedi: The assertion part is in the if condition
```
if (err === fakeError) {
  resolve();
}
```
if both the errors are same the promise will get resolve and test will get pass
```
else {
  reject(new Error('Unexpected error emitted'));
}
```
otherwise the promise will fail with the rejected promise

alkatrivedi: Based on our 1:1. I have updated the test, and have used assert instead of if condition

alkatrivedi:
Closing this PR, since we are facing issues in presubmit check for Owlbot Processor. I have moved the changes into another branch which has been created from the main repo. Link to the new [PR](https://github.com/googleapis/nodejs-spanner/pull/2201).

PR #2192, opened 2024-11-15
feat: Add support for Interval

sagarwaal:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-spanner/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass

Fixes #<issue_number_goes_here> 🦕


system-test/spanner.ts
```diff
@@ -6523,6 +6530,264 @@ describe('Spanner', () => {
...
+          const intervalQuery = (done, database, query, value) => {
+            database.run(query, (err, rows) => {
+              assert.ifError(err);
+              const queriedValue = rows[0][0].value;
+              assert.deepStrictEqual(queriedValue, value);
+              done();
+            });
+          };
+
+          it('GOOGLE_STANDARD_SQL should bind the value when param type interval is used', done => {
```
surbhigarg92, line 6551: Please add a test for `GOOGLE_STANDARD_SQL should interval value is passed as string for both typed and untyped params.

var value = (new Interval(19, 768, BigInt('123456789123'))).toISO8601()

system-test/spanner.ts
```diff
@@ -367,6 +368,7 @@ describe('Spanner', () => {
       googleSqlTable = DATABASE.table(TABLE_NAME);
       postgreSqlTable = PG_DATABASE.table(TABLE_NAME);
       if (IS_EMULATOR_ENABLED) {
+        // TODO: add columns using Interval Value and Interval Array Value.
```
surbhigarg92, line 371: Why this is skipped ?

system-test/spanner.ts
```diff
@@ -6523,6 +6530,264 @@ describe('Spanner', () => {
             });
           });
         });
+
+        describe('interval', () => {
```
surbhigarg92, line 6534: Test cases for insert are missing. These test cases are only for query. Please add a block in describe('types')

test/codec.ts
```diff
@@ -308,6 +308,515 @@ describe('codec', () => {
...
+            ),
+          },
+        ];
+
+        testCases.forEach(({input, expected}) => {
+          assert.deepStrictEqual(codec.Interval.fromISO8601(input), expected);
+        });
+      });
+
+      it('should throw error for invalid ISO8601 strings', () => {
```
surbhigarg92, line 563: Please also add test for null, undefined here and elsewhere

src/codec.ts
```diff
@@ -394,6 +394,251 @@ export class PGOid extends WrappedNumber {
...
+
+  getDays(): number {
+    return this.days;
+  }
+
+  getNanoseconds(): bigint {
+    return this.nanoseconds;
+  }
+
+  static fromMonths(months: number): Interval {
```
surbhigarg92, line 464: Please add documentation for all public methods.

src/codec.ts
```diff
@@ -394,6 +394,251 @@ export class PGOid extends WrappedNumber {
...
+ * @typedef Interval
+ * @see Spanner.interval
+ */
+export class Interval {
+  private months: number;
+  private days: number;
+  private nanoseconds: bigint;
+
+  private static readonly ISO8601_PATTERN: RegExp =
+    /^P(?!$)(-?\d+Y)?(-?\d+M)?(-?\d+D)?(T(?=-?[.,]?\d)(-?\d+H)?(-?\d+M)?(-?(((\d+)([.,]\d{1,9})?)|([.,]\d{1,9}))S)?)?$/;
```
surbhigarg92, line 407: Add code comments to explain the regex. Also elsewhere where its not self explanatory.

system-test/spanner.ts
sagarwaal: done, added test for typed-param, skipped untyped param as library doesn't support untyped param.

src/codec.ts
```diff
@@ -394,6 +394,251 @@ export class PGOid extends WrappedNumber {
...
+/**
+ * @typedef Interval
+ * @see Spanner.interval
+ */
+export class Interval {
+  private months: number;
+  private days: number;
+  private nanoseconds: bigint;
+
+  private static readonly ISO8601_PATTERN: RegExp =
```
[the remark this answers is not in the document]
sagarwaal: precise-date parsing will convert string to Date object, from that extracting original Interval is not possible. 

Interval is a duration without any fix start/end whereas Date object is a specified date.

system-test/spanner.ts
sagarwaal: Removed this comment, Interval is not a stored type, so we can't have columns of Interval type.

system-test/spanner.ts
sagarwaal: This is intended,  Interval is not a stored type, so we can't have columns of Interval type.

test/codec.ts
sagarwaal: Added tests for null and undefined.

src/codec.ts
sagarwaal: added.

src/codec.ts
sagarwaal: Added.

surbhigarg92 (approved):

rahul2393 (approved):

PR #2204, opened 2024-12-12
chore: support for common interface to get a session to support multiplexed session

alkatrivedi:
This PR contains the common interface class SessionFactory which will be responsible for the creation of the Session Pool and the Multiplexed Session(if the env variable would be set to true) upon client initialization.

This PR also contains the getSession method which will return the session(multiplexed/regular) based upon the env variable value.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -102,6 +102,7 @@ import Policy = google.iam.v1.Policy;
 import FieldMask = google.protobuf.FieldMask;
 import IDatabase = google.spanner.admin.database.v1.IDatabase;
 import snakeCase = require('lodash.snakecase');
+import {SessionFactory, SessionFactoryInterface} from './session-factory';
 import {
   ObservabilityOptions,
   Span,
@@ -339,6 +340,7 @@ class Database extends common.GrpcServiceObject {
   private instance: Instance;
   formattedName_: string;
   pool_: SessionPoolInterface;
+  sessionFactory_: SessionFactoryInterface;
   queryOptions_?: spannerClient.spanner.v1.ExecuteSqlRequest.IQueryOptions;
   resourceHeader_: {[k: string]: string};
   request: DatabaseRequest;
@@ -450,22 +452,12 @@ class Database extends common.GrpcServiceObject {
       },
     } as {} as ServiceObjectConfig);
 
-    this.pool_ =
-      typeof poolOptions === 'function'
-        ? new (poolOptions as SessionPoolConstructor)(this, null)
-        : new SessionPool(this, poolOptions);
-    const sessionPoolInstance = this.pool_ as SessionPool;
-    if (sessionPoolInstance) {
-      sessionPoolInstance._observabilityOptions =
-        instance._observabilityOptions;
-    }
     if (typeof poolOptions === 'object') {
       this.databaseRole = poolOptions.databaseRole || null;
       this.labels = poolOptions.labels || null;
     }
     this.formattedName_ = formattedName_;
     this.instance = instance;
-    this._observabilityOptions = instance._observabilityOptions;
     this._traceConfig = {
       opts: this._observabilityOptions,
       dbName: this.formattedName_,
@@ -478,8 +470,14 @@ class Database extends common.GrpcServiceObject {
     this._observabilityOptions = instance._observabilityOptions;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     this.requestStream = instance.requestStream as any;
-    this.pool_.on('error', this.emit.bind(this, 'error'));
-    this.pool_.open();
+    this.sessionFactory_ = new SessionFactory(this, name, poolOptions);
+    this.pool_ = this.sessionFactory_.getPool();
+    this.multiplexedSession_ = this.sessionFactory_.getMultiplexedSession();
+    const sessionPoolInstance = this.pool_ as SessionPool;
+    if (sessionPoolInstance) {
+      sessionPoolInstance._observabilityOptions =
+        instance._observabilityOptions;
+    }
     this.queryOptions_ = Object.assign(
       Object.assign({}, queryOptions),
       Database.getEnvironmentQueryOptions()
--- a/src/database.ts
+++ b/src/database.ts
@@ -92,6 +92,7 @@ import {
   ResourceCallback,
   Schema,
   addLeaderAwareRoutingHeader,
+  getCommonHeaders,
 } from './common';
 import {finished, Duplex, Readable, Transform} from 'stream';
 import {PreciseDate} from '@google-cloud/precise-date';
@@ -341,7 +342,7 @@ class Database extends common.GrpcServiceObject {
   pool_: SessionPoolInterface;
   sessionFactory_: SessionFactoryInterface;
   queryOptions_?: spannerClient.spanner.v1.ExecuteSqlRequest.IQueryOptions;
-  resourceHeader_: {[k: string]: string};
+  commonHeaders_: {[k: string]: string};
   request: DatabaseRequest;
   databaseRole?: string | null;
   labels?: {[k: string]: string} | null;
@@ -463,11 +464,13 @@ class Database extends common.GrpcServiceObject {
       dbName: this.formattedName_,
     };
 
-    this.resourceHeader_ = {
-      [CLOUD_RESOURCE_HEADER]: this.formattedName_,
-    };
     this.request = instance.request;
     this._observabilityOptions = instance._observabilityOptions;
+    this.commonHeaders_ = getCommonHeaders(
+      this.formattedName_,
+      this._observabilityOptions?.enableEndToEndTracing
+    );
+
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     this.requestStream = instance.requestStream as any;
     this.sessionFactory_ = new SessionFactory(this, name, poolOptions);
@@ -579,7 +582,7 @@ class Database extends common.GrpcServiceObject {
         method: 'updateDatabase',
         reqOpts,
         gaxOpts,
-        headers: this.resourceHeader_,
+        headers: this.commonHeaders_,
       },
       callback!
     );
@@ -685,7 +688,7 @@ class Database extends common.GrpcServiceObject {
       sessionCount: count,
     };
 
-    const headers = this.resourceHeader_;
+    const headers = this.commonHeaders_;
     if (this._getSpanner().routeToLeaderEnabled) {
       addLeaderAwareRoutingHeader(headers);
     }
@@ -986,7 +989,7 @@ class Database extends common.GrpcServiceObject {
     reqOpts.session.creatorRole =
       options.databaseRole || this.databaseRole || null;
 
-    const headers = this.resourceHeader_;
+    const headers = this.commonHeaders_;
     if (this._getSpanner().routeToLeaderEnabled) {
       addLeaderAwareRoutingHeader(headers);
     }
@@ -1212,7 +1215,7 @@ class Database extends common.GrpcServiceObject {
           method: 'dropDatabase',
           reqOpts,
           gaxOpts,
-          headers: this.resourceHeader_,
+          headers: this.commonHeaders_,
         },
         callback!
       );
@@ -1444,7 +1447,7 @@ class Database extends common.GrpcServiceObject {
         method: 'getDatabase',
         reqOpts,
         gaxOpts,
-        headers: this.resourceHeader_,
+        headers: this.commonHeaders_,
       },
       (err, resp) => {
         if (resp) {
@@ -1698,7 +1701,7 @@ class Database extends common.GrpcServiceObject {
         method: 'getDatabaseDdl',
         reqOpts,
         gaxOpts,
-        headers: this.resourceHeader_,
+        headers: this.commonHeaders_,
       },
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       (err, statements, ...args: any[]) => {
@@ -1778,7 +1781,7 @@ class Database extends common.GrpcServiceObject {
         method: 'getIamPolicy',
         reqOpts,
         gaxOpts: options.gaxOptions,
-        headers: this.resourceHeader_,
+        headers: this.commonHeaders_,
       },
       (err, resp) => {
         callback!(err, resp);
@@ -1917,7 +1920,7 @@ class Database extends common.GrpcServiceObject {
           method: 'listSessions',
           reqOpts,
           gaxOpts,
-          headers: this.resourceHeader_,
+          headers: this.commonHeaders_,
         },
         (err, sessions, nextPageRequest, ...args) => {
           if (err) {
@@ -2010,7 +2013,7 @@ class Database extends common.GrpcServiceObject {
       method: 'listSessionsStream',
       reqOpts,
       gaxOpts,
-      headers: this.resourceHeader_,
+      headers: this.commonHeaders_,
     });
   }
 
@@ -2402,7 +2405,7 @@ class Database extends common.GrpcServiceObject {
         method: 'listDatabaseRoles',
         reqOpts,
         gaxOpts,
-        headers: this.resourceHeader_,
+        headers: this.commonHeaders_,
       },
       (err, roles, nextPageRequest, ...args) => {
         const nextQuery = nextPageRequest!
@@ -2612,7 +2615,7 @@ class Database extends common.GrpcServiceObject {
         method: 'restoreDatabase',
         reqOpts,
         gaxOpts,
-        headers: this.resourceHeader_,
+        headers: this.commonHeaders_,
       },
       (err, operation, resp) => {
         if (err) {
@@ -3511,7 +3514,7 @@ class Database extends common.GrpcServiceObject {
             method: 'batchWrite',
             reqOpts,
             gaxOpts,
-            headers: this.resourceHeader_,
+            headers: this.commonHeaders_,
           });
           dataStream
             .once('data', () => (dataReceived = true))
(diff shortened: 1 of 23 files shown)


src/database.ts
```diff
@@ -478,8 +470,13 @@ class Database extends common.GrpcServiceObject {
     this._observabilityOptions = instance._observabilityOptions;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     this.requestStream = instance.requestStream as any;
-    this.pool_.on('error', this.emit.bind(this, 'error'));
-    this.pool_.open();
+    this.sessionFactory_ = new SessionFactory(this, name, poolOptions);
+    this.pool_ = this.sessionFactory_.getPool();
```
surbhigarg92, line 474: Why do we need pool_ object here. This should be replaced with `sessionFactory_` everywhere right ?

src/session-factory.ts
```diff
@@ -0,0 +1,131 @@
...
+  ): void;
+}
+
+/**
+ * Interface for implementing session-factory logic.
+ *
+ * @interface SessionFactoryInterface
+ */
+export interface SessionFactoryInterface {
+  getSession(callback: GetSessionCallback): void;
```
surbhigarg92, line 51: There should also be a `release` method in this ?

src/session-factory.ts
```diff
@@ -0,0 +1,131 @@
...
+  getSession(callback: GetSessionCallback): void {
+    if (process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS === 'true') {
+      this.multiplexedSession_?.getSession((err, session) => {
+        err ? callback(err, null) : callback(null, session);
+      });
+    } else {
+      this.pool_?.getSession((err, session) => {
+        err ? callback(err, null) : callback(null, session);
+      });
+    }
```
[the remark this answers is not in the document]
alkatrivedi: done

src/session-factory.ts
```diff
@@ -0,0 +1,131 @@
...
+    this.pool_ =
+      typeof poolOptions === 'function'
+        ? new (poolOptions as SessionPoolConstructor)(database, null)
+        : new SessionPool(database, poolOptions);
+    this.multiplexedSession_ = new MultiplexedSession(database);
+    this.pool_.on('error', this.emit.bind(database, 'error'));
+
+    // create session pool
+    this.pool_.open();
+    // multiplexed session should only get created if the env varaible is enabled
```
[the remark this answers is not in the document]
alkatrivedi: done

src/session-factory.ts
```diff
@@ -0,0 +1,131 @@
...
+  ) {
+    super({
+      parent: database,
+      id: name,
+    } as {} as ServiceObjectConfig);
+    this.pool_ =
+      typeof poolOptions === 'function'
+        ? new (poolOptions as SessionPoolConstructor)(database, null)
+        : new SessionPool(database, poolOptions);
+    this.multiplexedSession_ = new MultiplexedSession(database);
```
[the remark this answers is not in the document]
alkatrivedi: updated

src/session-factory.ts
```diff
@@ -0,0 +1,131 @@
...
+    this.pool_.on('error', this.emit.bind(database, 'error'));
+
+    // create session pool
+    this.pool_.open();
+    // multiplexed session should only get created if the env varaible is enabled
+    if (process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS === 'true') {
+      this.multiplexedSession_.on('error', this.emit.bind(database, 'error'));
+
+      // create multiplexed session
```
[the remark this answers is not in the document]
alkatrivedi: done

src/database.ts
alkatrivedi: some RW methods like `runTransaction`, `batchWriteAtLeastOnce` are getting the session from the Session Pool. Since, the changes to these RW methods is out of the scope in the first phase of the development, hence we will need to ensure the pool_ object for their use

surbhigarg92:
What would happen if `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` is not set to `true` during the app startup but set when the app is running ? 
IMU, this would return an error. Could you please add a test case for it also verify this behavior in other languages .

src/session-factory.ts
alkatrivedi: yeah, updated the class with the release method

harshachinta:
> What would happen if `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` is not set to `true` during the app startup but set when the app is running ? IMU, this would return an error. Could you please add a test case for it also verify this behavior in other languages .

@surbhigarg92 
IIn Java and Golang, the environment variable GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS is read only during client initialization, and its value is stored in memory. For subsequent operations, such as creating multiplexed sessions or executing read-only transactions, the in-memory state is used instead of re-reading the environment variable.

Scenarios:
`Environment Variable Set to true Before Client Initialization:`
1. The in-memory state is initialized as `true` during client initialization.
2. When the customer invokes a read-only transaction, the in-memory state is read as true, and a multiplexed session is used.

`Environment Variable Set to true After Client Initialization:`
1. The in-memory state remains false (as it was captured during initialization).
2. When the customer invokes a read-only transaction, the in-memory state is read as false, and a regular session is used.

I understand the confusion arises because, in Node.js, the environment variable is read dynamically each time `getsession` is invoked. It might be more simpler if we can compute the value once during client initialization and then relying on the in-memory state throughout.

cc: @alkatrivedi

src/session-factory.ts
```diff
@@ -0,0 +1,143 @@
...
+    } as {} as ServiceObjectConfig);
+    this.pool_ =
+      typeof poolOptions === 'function'
+        ? new (poolOptions as SessionPoolConstructor)(database, null)
+        : new SessionPool(database, poolOptions);
+    this.pool_.on('error', this.emit.bind(database, 'error'));
+    this.pool_.open();
+    // multiplexed session should only get created if the env variable is enabled
+    if (process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS === 'true') {
+      this.multiplexedSession_ = new MultiplexedSession(database);
```
harshachinta, line 90: @alkatrivedi 
You can store this env computation in a variable and use it in `getSession`, instead of recomputing again.

src/session-factory.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.pool_ =
+      typeof poolOptions === 'function'
+        ? new (poolOptions as SessionPoolConstructor)(database, null)
+        : new SessionPool(database, poolOptions);
+    this.pool_.on('error', this.emit.bind(database, 'error'));
+    this.pool_.open();
+    this.isMuxCreated = false;
+    // multiplexed session should only get created if the env variable is enabled
+    if (process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS === 'true') {
+      this.isMuxCreated = true;
```
harshachinta, line 92:
```suggestion
      this.isMultiplexedSessionEnabled = true;
```

alkatrivedi: Have made the suggested change also, I have moved the in-memory variable to the Multiplexed Session class

src/session-factory.ts
alkatrivedi: Thanks for the suggestion, I have made the changes accordingly

src/session-factory.ts
```diff
@@ -0,0 +1,144 @@
...
+      id: name,
+    } as {} as ServiceObjectConfig);
+    this.pool_ =
+      typeof poolOptions === 'function'
+        ? new (poolOptions as SessionPoolConstructor)(database, null)
+        : new SessionPool(database, poolOptions);
+    this.pool_.on('error', this.emit.bind(database, 'error'));
+    this.pool_.open();
+    this.multiplexedSession_ = new MultiplexedSession(database);
+    // multiplexed session should only get created if the env variable is enabled
```
surbhigarg92, line 89:
```suggestion
    // Multiplexed sessions should only be created if its enabled.
```

test/database.ts
```diff
@@ -120,6 +120,37 @@ class FakeSessionPool extends EventEmitter {
...
+}
+
+export class FakeSessionFactory extends EventEmitter {
+  calledWith_: IArguments;
+  constructor() {
+    super();
+    this.calledWith_ = arguments;
+  }
+  getSession(): FakeSession | FakeMultiplexedSession {
+    if (process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS) {
```
surbhigarg92, line 140: This condition should be reversed ?

test/database.ts
```diff
@@ -295,43 +327,30 @@ describe('Database', () => {
       assert(database.formattedName_, formattedName);
     });
 
-    it('should create a SessionPool object', () => {
-      assert(database.pool_ instanceof FakeSessionPool);
-      assert.strictEqual(database.pool_.calledWith_[0], database);
-      assert.strictEqual(database.pool_.calledWith_[1], POOL_OPTIONS);
-    });
```
surbhigarg92, line 302: Why do we need to delete this test case, database class still has "pool_" object. Can we modify this test to use sessionfactory to instantiate pool object

test/database.ts
```diff
@@ -295,43 +327,30 @@ describe('Database', () => {
...
-
-    it('should accept a custom Pool class', () => {
```
surbhigarg92, line 304: Again there is no change in the constructor of database. So why this is deleted ?

test/spanner.ts
```diff
@@ -5075,6 +5076,23 @@ describe('Spanner with mock server', () => {
...
+      const database = newTestDatabase();
+      // enable env after database creation
+      process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS = 'true';
+      const sessionFactory = database.sessionFactory_ as SessionFactory;
+      sessionFactory.getSession((err, _) => {
+        assert.ifError(err);
+        done();
+      });
+    });
+  });
```
surbhigarg92, line 5095: This should go to database class

test/session-factory.ts
```diff
@@ -0,0 +1,237 @@
...
+import * as sinon from 'sinon';
+import * as assert from 'assert';
+import {MultiplexedSession} from '../src/multiplexed-session';
+import {util} from '@google-cloud/common';
+import * as db from '../src/database';
+import {FakeTransaction} from './session-pool';
+import {ReleaseError} from '../src/session-pool';
+
+describe('SessionFactory', () => {
+  process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS = 'true';
```
surbhigarg92, line 28: Can we initialize this in `before` block and clear in `after` block

test/database.ts
alkatrivedi: Now, the creation of session pool object is happening in the SessionFactory class, hence I have added a test in the tests for Session Factory and we need to delete this from here

test/database.ts
alkatrivedi: Yeah, this test can be for the both the classes Database and Session Factory. Thanks for noticing it

src/session-factory.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  /**
+   * Retrieves the session, either the regular session or the multiplexed session based upon the environment varibale
+   * If the environment variable GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS is set to `true` the method will attempt to
+   * retrieve the multiplexed session. Otherwise it will retrieve the session from the pool.
+   *
+   * The session is returned asynchronously via the provided callback, which will receive either an error or the session object.
+   * @param callback
+   */
```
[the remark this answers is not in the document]
alkatrivedi: done

src/session-factory.ts
alkatrivedi: done

src/session-factory.ts
```diff
@@ -0,0 +1,144 @@
...
+   * This method is used to return the session back to the pool after it is no longer needed.
+   *
+   * It only performs the operation if the Multiplexed Session is disabled which is controlled via the
+   * env variable GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS.
+   *
+   * @param session The session to be released. This should be an instance of `Session` that was previously
+   * acquired from the session pool.
+   *
+   * @throws {Error} Throws an error if the session is invalid or cannot be released.
+   */
```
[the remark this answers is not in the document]
alkatrivedi: done

src/session-factory.ts
```diff
@@ -0,0 +1,144 @@
...
+  release(session: Session): void {
+    if (!this.isMuxCreated) {
```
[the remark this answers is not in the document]
alkatrivedi: done

src/session-factory.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  /**
+   * Returns the SessionPoolInterface used by the current instance, which provide access to the session pool
+   * for obtaining database sessions.
+   *
+   * @returns {SessionPoolInterface} The session pool used by current instance.
+   * This object allows interaction with the pool for acquiring and managing sessions.
+   */
+
```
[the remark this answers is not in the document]
alkatrivedi: done

test/database.ts
alkatrivedi: yeah, thanks for pointing this out
updated

test/database.ts
alkatrivedi: updated

test/session-factory.ts
```diff
@@ -0,0 +1,237 @@
...
+      const openStub = sandbox
+        .stub(SessionPool.prototype, 'open')
+        .callsFake(() => {});
+
+      new SessionFactory(DATABASE, NAME, POOL_OPTIONS);
+
+      assert.strictEqual(openStub.callCount, 1);
+    });
+
+    describe('when env GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS is enabled', () => {
```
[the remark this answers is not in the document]
alkatrivedi: done

test/session-factory.ts
```diff
@@ -0,0 +1,237 @@
...
+    });
+  });
+
+  describe('getSession', () => {
+    describe('for regular session', () => {
+      before(() => {
+        process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS = 'false';
+      });
+
+      it('should return the regular session if GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS env is disabled', done => {
```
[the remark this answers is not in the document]
alkatrivedi: done

test/session-factory.ts
```diff
@@ -0,0 +1,237 @@
...
+          sandbox.stub(sessionFactory.pool_, 'getSession') as sinon.SinonStub
+        ).callsFake(callback => callback(null, fakeSession));
+        sessionFactory.getSession((err, resp) => {
+          assert.strictEqual(err, null);
+          assert.strictEqual(resp, fakeSession);
+          done();
+        });
+      });
+
+      it('should return the error from getSession if GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS env is disabled and regular session creation get failed', done => {
```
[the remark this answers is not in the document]
alkatrivedi: done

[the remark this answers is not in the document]
alkatrivedi: modified other tests as well

test/spanner.ts
alkatrivedi: since, we are getting sessionFactory as an instance of FakeSessionFactory instead of SessionFactory, 
in FakeSessionFactory, we are unable to use in-memory variable - https://github.com/googleapis/nodejs-spanner/blob/get-session-class/test/database.ts#L133-#L152 instead we are dynamically fetching the env variable configuration using `process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS `

Hence, even though we are enabling GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS after client initialization, still getSession is returning FakeMultiplexedSession which is not an expected behavior (expected behavior is to return FakeSession if enabling GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS after initialization)

alkatrivedi: We need to keep this test in spanner test, so that we can test the desired behavior after changing the GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS after client initialization

test/session-factory.ts
alkatrivedi: we need to keep this in after block so that others test which will be running after this test don't fail on because of the change in env configuration

src/session-factory.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Creates a SessionFactory object to manage the creation of
+ *      session-pool and multiplexed session.
+ *
+ * @class
+ *
+ * @param {Database} database Database object.
+ * @param {String} name Name of the database.
+ * @param {SessionPoolOptions|SessionPoolInterface} options Session pool
```
surbhigarg92, line 83: SessionPoolConstructor right
```suggestion
 * @param {SessionPoolOptions| SessionPoolConstructor} options Session pool
```

surbhigarg92: Please let me know if my understanding is incorrect ?

surbhigarg92 (approved):

src/session-factory.ts
alkatrivedi: Since, the `SessionPoolInterface` is eventually creating `SessionPoolConstructor` object, hence we can keep SessionPoolInterface here in the documentation

PR #2205, opened 2024-12-21
feat(debugging): implement x-goog-spanner-request-id propagation per request

### LINKED ISSUE #2200: Implement propagation of x-goog-spanner-request-id gRPC header on every call
### A screenshot that you have tested with "Try this API".

This is a debugging feature request

### What would you like to see in the library?

This is a feature request to implement propagation of x-goog-spanner-request-id gRPC header on every call.

## Reasoning
x-goog-spanner-request-id allows propagation of an ID that'll be correlated between client and server calls. We can't rely on trace headers because they are sampled, convoluted through a bunch of infrastructure so Google Engineering asked that we craft up a better mechanism. The value will allow tracking the operations of an RPC even through retries

## Requirements
Each x-goog-spanner-request-id requires the following fields
- [ ] Version of the specification being implemented: ordinal number: 1 for version 1
- [ ] random Uint64 for the process' lifetime
- [ ] ID of the SpannerClient being used
- [ ] channelID for the gRPC channel being used
- [ ] Nth Request Count: nthReq
- [ ] RPC attempt number for nthReq

## Specification
Please reference https://orijtech.notion.site/x-goog-spanner-request-id-always-on-gRPC-header-to-aid-in-quick-debugging-of-errors-14aba6bc91348091a58fca7a505c9827?pvs=4

## Reference implementation
I built a Go reference implementation at https://github.com/googleapis/google-cloud-go/pull/11048

/cc @tharoldD

### Describe alternatives you've considered

_No response_

### Additional context/notes

_No response_

odeke-em:
Implements propagation of the x-goog-spanner-request-id that'll be propagated for every call. Once an error has been encountered, that error will have `.requestId` set.

Fixes #2200

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -113,6 +113,13 @@ import {
   setSpanErrorAndException,
   traceConfig,
 } from './instrument';
+import {
+  AtomicCounter,
+  X_GOOG_SPANNER_REQUEST_ID_HEADER,
+  craftRequestId,
+  newAtomicCounter,
+} from './request_id_header';
+
 export type GetDatabaseRolesCallback = RequestCallback<
   IDatabaseRole,
   databaseAdmin.spanner.admin.database.v1.IListDatabaseRolesResponse
@@ -355,6 +362,8 @@ class Database extends common.GrpcServiceObject {
   > | null;
   _observabilityOptions?: ObservabilityOptions; // TODO: exmaine if we can remove it
   private _traceConfig: traceConfig;
+  private _nthRequest: AtomicCounter;
+  public _clientId: number;
   constructor(
     instance: Instance,
     name: string,
@@ -469,6 +478,12 @@ class Database extends common.GrpcServiceObject {
     };
 
     this.request = instance.request;
+    this._nthRequest = newAtomicCounter(0);
+    if (this.parent && this.parent.parent) {
+      this._clientId = (this.parent.parent as Spanner)._nthClientId;
+    } else {
+      this._clientId = instance._nthClientId;
+    }
     this._observabilityOptions = instance._observabilityOptions;
     this.commonHeaders_ = getCommonHeaders(
       this.formattedName_,
@@ -489,6 +504,11 @@ class Database extends common.GrpcServiceObject {
       Database.getEnvironmentQueryOptions()
     );
   }
+
+  _nextNthRequest(): number {
+    return this._nthRequest.increment();
+  }
+
   /**
    * @typedef {array} SetDatabaseMetadataResponse
    * @property {object} 0 The {@link Database} metadata.
@@ -697,6 +717,12 @@ class Database extends common.GrpcServiceObject {
       addLeaderAwareRoutingHeader(headers);
     }
 
+    const allHeaders = this._metadataWithRequestId(
+      this._nextNthRequest(),
+      1,
+      headers
+    );
+
     startTrace('Database.batchCreateSessions', this._traceConfig, span => {
       this.request<google.spanner.v1.IBatchCreateSessionsResponse>(
         {
@@ -704,7 +730,7 @@ class Database extends common.GrpcServiceObject {
           method: 'batchCreateSessions',
           reqOpts,
           gaxOpts: options.gaxOptions,
-          headers: headers,
+          headers: allHeaders,
         },
         (err, resp) => {
           if (err) {
@@ -726,6 +752,26 @@ class Database extends common.GrpcServiceObject {
         }
       );
     });
+  }
+
+  public _metadataWithRequestId(
+    nthRequest: number,
+    attempt: number,
+    priorMetadata?: {[k: string]: string}
+  ): {[k: string]: string} {
+    if (!priorMetadata) {
+      priorMetadata = {};
+    }
+    const withReqId = {
+      ...priorMetadata,
+    };
+    withReqId[X_GOOG_SPANNER_REQUEST_ID_HEADER] = craftRequestId(
+      this._clientId || 1,
+      1, // TODO: Properly infer the channelId
+      nthRequest,
+      attempt
+    );
+    return withReqId;
   }
 
   /**
@@ -993,7 +1039,11 @@ class Database extends common.GrpcServiceObject {
     reqOpts.session.creatorRole =
       options.databaseRole || this.databaseRole || null;
 
-    const headers = this.commonHeaders_;
+    const headers = this._metadataWithRequestId(
+      this._nextNthRequest(),
+      1,
+      this.commonHeaders_
+    );
     if (this._getSpanner().routeToLeaderEnabled) {
       addLeaderAwareRoutingHeader(headers);
     }
@@ -1915,6 +1965,12 @@ class Database extends common.GrpcServiceObject {
       delete (gaxOpts as GetSessionsOptions).pageToken;
     }
 
+    const headers = this._metadataWithRequestId(
+      this._nextNthRequest(),
+      1,
+      this.commonHeaders_
+    );
+
     return startTrace('Database.getSessions', this._traceConfig, span => {
       this.request<
         google.spanner.v1.ISession,
@@ -1925,7 +1981,7 @@ class Database extends common.GrpcServiceObject {
           method: 'listSessions',
           reqOpts,
           gaxOpts,
-          headers: this.commonHeaders_,
+          headers: headers,
         },
         (err, sessions, nextPageRequest, ...args) => {
           if (err) {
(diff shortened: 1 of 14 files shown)


odeke-em:
@olavloite @surbhigarg92 kindly help me take a look at this PR, it is ready for review. Thank you!

src/index.ts
```diff
@@ -321,6 +334,10 @@ class Spanner extends GrpcService {
         'grpc.callInvocationTransformer': grpcGcp.gcpCallInvocationTransformer,
         'grpc.channelFactoryOverride': grpcGcp.gcpChannelFactoryOverride,
         'grpc.gcpApiConfig': grpcGcp.createGcpApiConfig(gcpApiConfig),
+
+        // TODO: Negotiate with the Google team to plumb gRPC
+        // settings such as interceptors to the gRPC client.
+        // 'grpc.interceptors': interceptors,
```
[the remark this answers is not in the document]
odeke-em: No action at all sadly, I reported to the maintainers of this library but out of scope currently and not a priority for them.

test/spanner.ts
```diff
@@ -2690,6 +2877,7 @@ describe('Spanner with mock server', () => {
             (e as ServiceError).message,
             'No resources available.'
           );
+          // assert.deepStrictEqual((e as RequestIDError).requestID,`1.${randIdForProcess}.1.1.1.1`);
```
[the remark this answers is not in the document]
odeke-em: It is because that error isn't sent via RPC but really just thrown when `fail: true` is set.

test/spanner.ts
```diff
@@ -2764,6 +2952,7 @@ describe('Spanner with mock server', () => {
         assert.fail('missing expected error');
       } catch (err) {
         assert.strictEqual((err as ServiceError).code, Status.NOT_FOUND);
+        // assert.deepStrictEqual((err as RequestIDError).requestID,`1.${randIdForProcess}.1.1.1.1`);
```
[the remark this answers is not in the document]
odeke-em: done, thanks!

src/database.ts
```diff
@@ -723,6 +749,26 @@ class Database extends common.GrpcServiceObject {
...
+  ): {[k: string]: string} {
+    if (!priorMetadata) {
+      priorMetadata = {};
+    }
+    const withReqId = {
+      ...priorMetadata,
+    };
+    withReqId[X_GOOG_SPANNER_REQUEST_ID_HEADER] = craftRequestId(
+      this._clientId || 1,
+      1, // TODO: Properly infer the channelId
```
odeke-em, line 765: @olavloite I filed a TODO for retrieving the channelId which I believe can be done much later in a phased series of PRs instead of at once.

src/index.ts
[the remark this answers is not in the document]
olavloite: Then let's remove the TODO if the action has already been taken, even though the outcome was not what was hoped for.

src/index.ts
```diff
@@ -309,6 +316,12 @@ class Spanner extends GrpcService {
         }
       }
     }
+
+    let interceptors: any[] = [];
+    if (options) {
+      interceptors = options.interceptors || [];
+    }
```
olavloite, line 323: Can we also remove this for now, as it is not being used?

src/index.ts
```diff
@@ -248,6 +254,7 @@ class Spanner extends GrpcService {
   routeToLeaderEnabled = true;
   directedReadOptions: google.spanner.v1.IDirectedReadOptions | null;
   _observabilityOptions: ObservabilityOptions | undefined;
+  _nthClientId: number;
```
olavloite, line 257: nit: can this be `readonly`?

src/index.ts
```diff
@@ -1551,7 +1569,55 @@ class Spanner extends GrpcService {
           },
         })
       );
-      callback(null, requestFn);
+
+      // Wrap requestFn so as to inject the spanner request id into
+      // every returned error, so that users can have debugging continuity.
```
olavloite, line 1574: nit: Simplify comment

```suggestion
      // Wrap requestFn to inject the spanner request id into every returned error.
```

src/session.ts
```diff
@@ -534,6 +554,11 @@ export class Session extends common.GrpcServiceObject {
   private _getSpanner(): Spanner {
     return this.parent.parent.parent as Spanner;
   }
+
+  private channelId(): number {
+    // TODO: Infer channelId from the actual gRPC channel.
+    return 1;
```
olavloite, line 560: The Node.js client does not use a channel pool, meaning that returning 1 for `channelId` seems to be the correct thing to do in all cases. Meaning: We should replace the TODO with something like this:

```suggestion
    // The Node.js client does not use a gRPC channel pool, so this function always returns 1.
    return 1;
```

test/gapic_spanner_v1.ts
```diff
@@ -26,6 +26,10 @@ import * as spannerModule from '../src';
 import {PassThrough} from 'stream';
 
 import {protobuf} from 'google-gax';
+import {
```
olavloite, line 29: This file is auto-generated and should not be modified manually

test/spanner.ts
```diff
@@ -106,7 +240,13 @@ describe('Spanner with mock server', () => {
   const fooNotFoundErr = Object.assign(new Error('Table FOO not found'), {
     code: grpc.status.NOT_FOUND,
   });
-  const server = new grpc.Server();
+  const xGoogReqIDInterceptor = new XGoogRequestHeaderInterceptor([
+    '/google.spanner.admin',
+    '/google.spanner.admin.database.v1.DatabaseAdmin',
```
olavloite, line 244: nit: isn't this second prefix redundant due to the prefix above (`/google.spanner.admin`)?

test/spanner.ts
```diff
@@ -1047,6 +1199,18 @@ describe('Spanner with mock server', () => {
           .catch(done)
           .then(() => done());
       });
+      const gotStreamingCalls = xGoogReqIDInterceptor.getStreamingCalls();
```
olavloite, line 1202: I added this verification to check that the headers are actually being sent and received. Can you add more of these to the various tests here, because:
1. Currently, the attempt number is never increased. If we want to do that in a follow-up PR, then that is fine, but we should then add a TODO to these verifications.
2. The current 'global' verification for the request headers has a limitation, and that is that it does not seem to catch missing headers for streaming RPCs. It only verifies that if the header is set, that it has the correct format.

src/index.ts
odeke-em: Done, thanks!

src/index.ts
odeke-em: Done, thanks!

src/index.ts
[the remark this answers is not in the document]
olavloite: Well, if that is something that could happen, then can we reword the TODO to reflect that? Now it seems like we are waiting for an answer from a negotiation. So maybe something like this:

```
// TODO: Use interceptors directly on the gRPC client when that is supported.
```

test/spanner.ts
```diff
@@ -170,6 +313,7 @@ describe('Spanner with mock server', () => {
       servicePath: 'localhost',
       port,
       sslCreds: grpc.credentials.createInsecure(),
+      // interceptors: [xGoogReqIDInterceptor.generateLoggingClientInterceptor()],
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

test/spanner.ts
odeke-em: I've filed a TODO to send a follow-up PR.

odeke-em:
@olavloite sorry I had just 1 TODO left which I have sent up; kindly help me re-run the bots.

test/spanner.ts
```diff
@@ -5441,6 +5655,81 @@ describe('Spanner with mock server', () => {
...
+          spannerMock.abortTransaction(transaction!);
+        }
+        const [rows] = await transaction!.run(selectSql);
+        rows.forEach(() => rowCount++);
+        assert.strictEqual(rowCount, 3);
+        assert.strictEqual(attempts, 4);
+        await transaction!.commit();
+      });
+
+      // TODO: introduce tests for incremented attempts to verify
```
olavloite, line 5677: This test checks the behavior when a transaction is retried due to Spanner aborting a read/write transaction. That should not lead to the attempt number increasing, meaning that there is no pending TODO in this test. The current behavior should remain exactly the same as what is verified in this test now, also when support for increasing the attempt number has been implemented.

The attempt number should only be bumped when a single RPC is retried by Gax.

odeke-em: Oh yes, I just put the TODO inside the test I had included blindly. Let me make an update then ping. Thanks for spotting that.

olavloite (approved):

PR #2207, opened 2024-12-24
chore: Improve error message for Read-only transaction with bounded staleness

surbhigarg92:
This PR will improve error message for Read Only transactions when `maxStaleness` or `minReadTimestamp` is passed.
 
```
database.getSnapshot({maxStaleness: 100}, async (err, transaction) => {
    if (err) {
      console.error(err); // 'maxStaleness / minReadTimestamp is not supported for multi-use read-only transactions.'
      return;
    }
    try {
      const [qOneRows] = await transaction.run('SELECT SingerId, AlbumId, AlbumTitle FROM Albums');
    } finally {
      transaction.end();
      // Close the database when finished.
      await database.close();
    }
  });
 ```

### CODE REVIEW

File: test/database.ts [hunk 1]
```diff
@@ -2307,6 +2307,30 @@ describe('Database', () => {
       assert.strictEqual(bounds, fakeTimestampBounds);
     });
 
+    it('should throw error if maxStaleness is passed in the timestamp bounds to the snapshot', () => {
+      const fakeTimestampBounds = {maxStaleness: 10};
```
sakthivelmanii, line 2311: what will happen if someone assigns them with `null/undefined`?

surbhigarg92: Thanks for identifying this. I missed this check earlier.

sakthivelmanii (approved):

PR #2211, opened 2025-01-07
feat(x-goog-spanner-request-id): add bases

odeke-em:
This change brings in the bases for x-goog-spanner-request-id by commit AtomicCounter and the various utilities plus some unit tests.

Ripped out of PR #2205
Updates  #2200


odeke-em:
Kindly pinging @olavloite @surbhigarg92 to help bring this in as it is carved out of the larger PR #2205 and is isolated away from main code.

src/request_id_header.ts
```diff
@@ -0,0 +1,327 @@
+/*!
```
olavloite, line 1: nit:
```suggestion
/**
```

src/request_id_header.ts
```diff
@@ -0,0 +1,327 @@
...
+ * limitations under the License.
+ */
+
+import {randomBytes} from 'crypto';
+import * as grpc from '@grpc/grpc-js';
+const randIdForProcess = randomBytes(8).readBigUint64LE(0).toString();
+const X_GOOG_SPANNER_REQUEST_ID_HEADER = 'x-goog-spanner-request-id';
+
+class AtomicCounter {
+  private backingBuffer: Uint32Array;
```
olavloite, line 23:
```suggestion
  private readonly backingBuffer: Uint32Array;
```

src/request_id_header.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  constructor(initialValue?: number) {
+    this.backingBuffer = new Uint32Array(
+      new SharedArrayBuffer(Uint32Array.BYTES_PER_ELEMENT)
```
olavloite, line 27: I am not familiar with this in Node.js, but while looking into what this is, I ran into this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements

Does this affect the usability of this in Node.js? Or is this an issue that is only relevant to code running in a browser?

src/request_id_header.ts
```diff
@@ -0,0 +1,327 @@
...
+  }
+}
+
+function craftRequestId(
+  nthClientId: number,
+  channelId: number,
+  nthRequest: number,
+  attempt: number
+) {
+  return `1.${randIdForProcess}.${nthClientId}.${channelId}.${nthRequest}.${attempt}`;
```
olavloite, line 61: nit: Can we add a comment that explains why there is a '1' at the start of this string, e.g.:

// The first digit in this string is the version number of the header type that is used.

src/request_id_header.ts
```diff
@@ -0,0 +1,327 @@
...
+  return nthClientId.value();
+}
+
+function newAtomicCounter(n?: number): AtomicCounter {
+  return new AtomicCounter(n);
+}
+
+const X_GOOG_REQ_ID_REGEX = /(\d+\.){5}\d+/;
+
+class XGoogRequestHeaderInterceptor {
```
olavloite, line 87: If I understand this correctly, this is a server interceptor, meaning that it will be used for testing, right? Could we then move this to a test file?

src/request_id_header.ts
```diff
@@ -0,0 +1,327 @@
...
+    const method = call.handler.path;
+    const isUnary = call.handler.type === 'unary';
+    const that = this;
+    const listener = new grpc.ServerListenerBuilder()
+      .withOnReceiveMetadata((metadata, next) => {
+        let i = 0;
+        const prefixesToIgnore: string[] = that.prefixesToIgnore || [];
+        for (i = 0; i < prefixesToIgnore.length; i++) {
+          const prefix = prefixesToIgnore[i];
+          // console.log(`prefix: ${prefix}\nmethod: ${method}`);
```
olavloite, line 165: nit: remove?

src/request_id_header.ts
```diff
@@ -0,0 +1,327 @@
...
+}
+
+interface withMetadataWithRequestId {
+  _nthClientId: number;
+  _channelId: number;
+}
+
+function injectRequestIDIntoHeaders(
+  headers: {[k: string]: string},
+  session: any,
```
olavloite, line 257: Could we make this typed?

test/request_id_header.ts
```diff
@@ -0,0 +1,162 @@
+/*!
```
olavloite, line 1: nit:
```suggestion
/**
```

src/request_id_header.ts
odeke-em: Yeah, that's for browsers.

src/request_id_header.ts
odeke-em: That would create an import cycle because Session also then uses injectRequestIDIntoHeaders in the follow-up PR.

test/request_id_header.ts
odeke-em: Done, thanks!

src/request_id_header.ts
odeke-em: Done, thanks!

src/request_id_header.ts
odeke-em: Let's keep it in here for now but in the follow-up PR I shall move it there. Makes for easier rebasing.

odeke-em: Done and thankyou @olavloite, please take a look.

src/request_id_header.ts
odeke-em: Done.

odeke-em:
@surbhigarg92 @alkatrivedi @sakthivelmanii @harshachinta kindly help me trigger the bot runs.

odeke-em:
Kind ping @surbhigarg92 @olavloite

src/request_id_header.ts
olavloite: Ack

olavloite (approved):

odeke-em:
Awesome and thank you @olavloite 🤘🏾👏🏿👏🏿

PR #2235, opened 2025-02-08
feat(spanner): support for type UUID

alkatrivedi:
This PR contains support for UUID type on Cloud Spanner.

### DIFF
--- a/src/codec.ts
+++ b/src/codec.ts
@@ -33,6 +33,7 @@ import {Big} from 'big.js';
 import {common as p} from 'protobufjs';
 import {google as spannerClient} from '../protos/protos';
 import {GoogleError} from 'google-gax';
+import * as uuid from 'uuid';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type Value = any;
@@ -1034,6 +1035,7 @@ const TypeCode: {
   bool: 'BOOL',
   int64: 'INT64',
   pgOid: 'INT64',
+  uuid: 'UUID',
   float32: 'FLOAT32',
   float64: 'FLOAT64',
   numeric: 'NUMERIC',
@@ -1075,6 +1077,7 @@ interface FieldType extends Type {
 /**
  * @typedef {object} ParamType
  * @property {string} type The param type. Must be one of the following:
+ *     - uuid
  *     - float32
  *     - float64
  *     - int64
@@ -1157,6 +1160,10 @@ function getType(value: Value): Type {
     return {type: 'bool'};
   }
 
+  if (uuid.validate(value)) {
+    return {type: 'unspecified'};
+  }
+  
   if (isString(value)) {
     return {type: 'string'};
   }
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -1728,7 +1728,15 @@ export class Snapshot extends EventEmitter {
     if (!isEmpty(typeMap)) {
       Object.keys(typeMap).forEach(param => {
         const type = typeMap[param];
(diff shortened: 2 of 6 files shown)


system-test/spanner.ts
```diff
@@ -5170,6 +5245,196 @@ describe('Spanner', () => {
           });
         });
 
+        // TODO: Enable when the uuid feature has been released.
+        describe.skip('uuid', () => {
```
sakthivelmanii, line 5249: have you tested this against cloud-devel?

alkatrivedi: yeah

src/codec.ts
```diff
@@ -814,6 +817,10 @@ function getType(value: Value): Type {
     return {type: 'bool'};
   }
 
+  if (uuid.validate(value)) {
+    return {type: 'unspecified'};
```
sakthivelmanii, line 821: why unspecified here?

alkatrivedi: cause we are not sure if the customer has inserted the UUID value in the column of type UUID/String. Hence, we are returning type as unspecified implicitly, but while making gRPC call we are making sure that if the type is unspecified we don't pass param type with the gaxOptions, this way are letting backend to figure out the type

system-test/spanner.ts
surbhigarg92: @sakthivelmanii We will be merging this when the feature is available in prod right ? So ideally we will remove this "skip" from the PR ?

src/transaction.ts
```diff
@@ -1521,7 +1521,17 @@ export class Snapshot extends EventEmitter {
...
+        if (type.child) {
+          if (
+            typeObject.code === 'ARRAY' &&
+            typeObject.arrayElementType?.code !== 'TYPE_CODE_UNSPECIFIED'
+          ) {
+            paramTypes[param] = codec.createTypeObject(type);
+          }
+        } else if (typeObject.code !== 'TYPE_CODE_UNSPECIFIED') {
+          paramTypes[param] = codec.createTypeObject(type);
+        }
```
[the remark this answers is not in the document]
alkatrivedi: done

system-test/spanner.ts
```diff
@@ -427,6 +427,7 @@ describe('Spanner', () => {
     before(async () => {
       if (IS_EMULATOR_ENABLED) {
         // TODO: add column Float32Value FLOAT32 and FLOAT32Array Array<FLOAT32> while using float32 feature.
```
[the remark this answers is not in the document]
alkatrivedi: sure

src/codec.ts
harshachinta: But will this work for parameterized queries. I think parameterized queries does not support UNSPECIFIED type.
@alkatrivedi can you check if parameterized query is working when passing a UUID value?

harshachinta: The case that we are talking about here is when customer triggers a following parameterized query without an explicit param type
const query = {
              sql: 'SELECT @v',
              params: {
                v: values,
              },
            };
I think this fails if value is of UUID type  and param type is not mentioned because client now sends a UNSPECIFIED type to backend

harshachinta: My suggestion here is
Ideally, if param type is not mentioned then treat the value as string.
If customer wants to treat that value as UUID then they can explicitly specify the param type.

surbhigarg92 (dismissed):

src/codec.ts
alkatrivedi: @flovouin Thanks for flagging this! I will take a look

PR #2236, opened 2025-02-18
fix: adding span attributes for request tag and transaction tag

surbhigarg92:


src/database.ts
```diff
@@ -3448,9 +3476,13 @@ class Database extends common.GrpcServiceObject {
 
     let sessionId = '';
     const getSession = this.pool_.getSession.bind(this.pool_);
+
     return startTrace(
       'Database.runTransactionAsync',
-      this._traceConfig,
+      {
+        ...this._traceConfig,
+        transactionTag: options?.requestOptions?.transactionTag,
```
alkatrivedi, line 3484: should we add a test for this in observability-test/database.ts ?

surbhigarg92: done

alkatrivedi (approved):

PR #2245, opened 2025-03-06
feat(spanner): add support for snapshot isolation

alkatrivedi:
This PR contains code changes to add support for option IsolationLevel at the client level and at the transaction level.
supported methods are(RW and Blind Write):

```
- writeAtLeastOnce
- runTransactionAsync
- runTransaction
- getTransaction
- async getTransaction(from transaction runner class)
```


src/table.ts
```diff
@@ -52,7 +54,10 @@ export type DropTableCallback = UpdateSchemaCallback;
 
 interface MutateRowsOptions extends CommitOptions {
   requestOptions?: Omit<IRequestOptions, 'requestTag'>;
-  excludeTxnFromChangeStreams?: boolean;
+  defaultTransactionOptions?: Pick<
```
surbhigarg92, line 57: This is a breaking change

src/database.ts
```diff
@@ -316,6 +316,10 @@ export interface RestoreOptions {
   gaxOptions?: CallOptions;
 }
 
+export interface WriteAtLeastOnceOptions extends CallOptions {
+  defaultTransactionOptions?: Pick<RunTransactionOptions, 'isolationLevel'>;
```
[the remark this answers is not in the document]
alkatrivedi: will add that option in a separate PR, it got skipped earlier

src/transaction.ts
```diff
@@ -2697,6 +2697,13 @@ export class Transaction extends Dml {
   excludeTxnFromChangeStreams(): void {
     this._options.excludeTxnFromChangeStreams = true;
   }
+
+  /**
+   * Use option isolationLevel to add the isolation level in the transaction.
+   */
+  setIsolationLevel(isolationLevel: any): void {
```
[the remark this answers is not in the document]
alkatrivedi: updated, thanks for the suggestion!

src/database.ts
```diff
@@ -2228,6 +2235,14 @@ class Database extends common.GrpcServiceObject {
           transaction!.excludeTxnFromChangeStreams();
         }
 
+        if (options.isolationLevel) {
+          transaction!.setIsolationLevel(options.isolationLevel);
+        } else if (defaultTransactionOptions) {
+          transaction!.setIsolationLevel(
+            defaultTransactionOptions.isolationLevel
+          );
+        }
```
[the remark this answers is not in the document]
alkatrivedi: modified

src/table.ts
alkatrivedi: updated

src/database.ts
```diff
@@ -2221,12 +2225,11 @@ class Database extends common.GrpcServiceObject {
...
           );
         }
-        if (options.optimisticLock) {
-          transaction!.useOptimisticLock();
-        }
-        if (options.excludeTxnFromChangeStreams) {
-          transaction!.excludeTxnFromChangeStreams();
-        }
+        transaction?.setReadWriteTransactionOptions(
+          options && Object.keys(options).length
```
surbhigarg92, line 2229: How would this handle the case, when `TransactionOptions` only has `excludeTxnFromChangeStreams` set and `isolation_level` needs to be taken from `defaultTransactionOptions`

src/transaction-runner.ts
```diff
@@ -26,6 +26,7 @@ import {isSessionNotFoundError} from './session-pool';
 import {Database} from './database';
 import {google} from '../protos/protos';
 import IRequestOptions = google.spanner.v1.IRequestOptions;
+import {protos} from '.';
```
surbhigarg92, line 29: `import {google} from '../protos/protos'` is already imported. 

You could import IsolationLevel on top like IRequestOptions

`import IsolationLevel = google.spanner.v1.TransactionOptions.IsolationLevel`

src/transaction-runner.ts
```diff
@@ -46,6 +47,7 @@ export interface RunTransactionOptions {
   requestOptions?: Pick<IRequestOptions, 'transactionTag'>;
   optimisticLock?: boolean;
   excludeTxnFromChangeStreams?: boolean;
+  isolationLevel?: protos.google.spanner.v1.TransactionOptions.IsolationLevel;
```
surbhigarg92, line 50: Use the `IsolationLevel` import instead of the full syntax.

src/transaction-runner.ts
```diff
@@ -124,7 +126,12 @@ export abstract class Runner<T> {
     this.transaction = transaction;
     this.transaction.useInRunner();
 
-    const defaults = {timeout: 3600000};
+    const defaults = {
+      timeout: 3600000,
+      isolationLevel:
+        protos.google.spanner.v1.TransactionOptions.IsolationLevel
```
surbhigarg92, line 132: Change it here and all other places.

src/transaction-runner.ts
```diff
@@ -202,12 +209,11 @@ export abstract class Runner<T> {
...
-    if (this.options.optimisticLock) {
-      transaction.useOptimisticLock();
-    }
-    if (this.options.excludeTxnFromChangeStreams) {
-      transaction.excludeTxnFromChangeStreams();
-    }
+    transaction!.setReadWriteTransactionOptions(
+      this.options && Object.keys(this.options).length
+        ? this.options
+        : this.session.parent._getSpanner().defaultTransactionOptions
```
surbhigarg92, line 215: Do not use _getSpanner() private method . Instead create a private method `_getSpanner` in this class and use it. Refer this method in all other classes

src/transaction.ts
```diff
@@ -1822,9 +1823,33 @@ export class Transaction extends Dml {
 
     this._queuedMutations = [];
     this._options = {readWrite: options};
+    this._options.isolationLevel =
```
surbhigarg92, line 1826: Similarly change the import and protos full path

src/transaction.ts
```diff
@@ -1822,9 +1823,33 @@ export class Transaction extends Dml {
...
+     * Set option excludeTxnFromChangeStreams=true to exclude read/write transactions
+     * from being tracked in change streams.
+     */
+    if (options?.excludeTxnFromChangeStreams) {
+      this._options.excludeTxnFromChangeStreams = true;
+    }
+    /**
+     * Set isolation level .
+     */
+    if (options?.isolationLevel) {
```
surbhigarg92, line 1848: how would this use the default isolation level if available ?

src/transaction.ts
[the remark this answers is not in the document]
surbhigarg92: Is this method used anywhere now ?

test/database.ts
```diff
@@ -3166,6 +3180,57 @@ describe('Database', () => {
...
+            .REPEATABLE_READ,
+      };
+
+      await database.runTransaction(fakeOptions, assert.ifError);
+
+      const options = fakeTransactionRunner.calledWith_[3];
+      assert.strictEqual(options, fakeOptions);
+    });
+
+    it('should optionally accept `option` isolationLevel when passed with Spanner options', () => {
```
surbhigarg92, line 3196: Add test cases for below 

1. Should override `isolationLevel` from runTransactionOptions, when passed in both `defaultTransactionOptions` and `runTransactionOptions`

2. Should use isolation level from `defaultTransactionOptions` when `runTransactionOptions` are passed but does not consists of isolation level and contains other properties. The resulting options should be a combination of `defaultTransactionOptions` and `runTransactionOptions`

surbhigarg92: 3. If `defaultTransactionOptions` has more properties except, isolation level, then other properties should be ignored.

src/transaction-runner.ts
alkatrivedi: done, thanks for the suggestion!

src/transaction-runner.ts
alkatrivedi: done

src/transaction-runner.ts
alkatrivedi: done

src/transaction-runner.ts
alkatrivedi: have refactored the code, please check

src/transaction.ts
alkatrivedi: I have updated the code
in the `Transaction` class constructor we are assigning the default isolationLevel value as `this._options.isolationLevel = IsolationLevel.ISOLATION_LEVEL_UNSPECIFIED;`

now here in this case `this._options.isolationLevel = options?.isolationLevel? options?.isolationLevel:this._getSpanner().defaultTransactionOptions.isolationLevel;`

if the `isolationLevel` will get pass at the transaction level it will assign that value to the `this._options.isolationLevel` otherwise this value `this._getSpanner().defaultTransactionOptions.isolationLevel ` will get assign which if being passed from spanner options will use that value otherwise it will also be containing default value of unspecified

test/database.ts
alkatrivedi: have added the tests 
1. 'Should override isolationLevel from runner `option`, when passed in both Spanner Options and runner `option`'
2. 'Should use isolationLevel from spanner `option`, when isolationLevel is passed at client `options` level and other options are passed at transaction `options` level'

src/database.ts
```diff
@@ -3683,6 +3674,7 @@ class Database extends common.GrpcServiceObject {
         span.addEvent('Using Session', {'session.id': session?.id});
         this._releaseOnEnd(session!, transaction!, span);
         try {
+          transaction!.setReadWriteTransactionOptions(options);
```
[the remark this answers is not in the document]
surbhigarg92: Do this for all other places

surbhigarg92 (approved):

PR #2284, opened 2025-04-20
feat: add promise based signatures for createQueryPartitions

### LINKED ISSUE #1899: Missing Promise functions on BatchTransaction
#### Environment details

  - OS: OSX
  - Node.js version: v18.16.0
  - npm version: 9.5.1
  - `@google-cloud/spanner` version: 6.12.0

#### Steps to reproduce

  1. Use Typescript
  2. Attempt to use batch transactions as shown in `samples/batch.js`
```
    const query = {
      sql: `
        select
          *
        from
          ${POST_TABLE_NAME}
      `,
    };

    const [transaction] = await this.database.createBatchTransaction();
    const partitions = transaction.createQueryPartitions(query);
```
3. Get `tsc` error:
```
    domains/archive/post-service/src/post-service-spanner.repository.ts:274:11 - error TS2488: Type 'void' must have a '[Symbol.iterator]()' method that returns an iterator.

    274     const [partitions] = await transaction.createQueryPartitions(query);
                  ~~~~~~~~~~~~
    domains/archive/post-service/src/post-service-spanner.repository.ts:274:44 - error TS2554: Expected 2 arguments, but got 1.

    274     const [partitions] = await transaction.createQueryPartitions(query);
                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

      node_modules/@google-cloud/spanner/build/src/batch-transaction.d.ts:111:39
        111     createQueryPartitions(query: any, callback: any): void;
                                                  ~~~~~~~~~~~~~
        An argument for 'callback' was not provided.
```

The problem appears to be that your use of `promisifyAll` doesn't cause the appropriate method types to be included in the `.d.ts` file.

alkatrivedi:
fixes: #1899


src/batch-transaction.ts
```diff
@@ -126,20 +137,34 @@ class BatchTransaction extends Snapshot {
...
+  ): Promise<CreateQueryPartitionsResponse>;
+  createQueryPartitions(
+    query: string | ExecuteSqlRequest,
+    callback: CreateQueryPartitionsCallback,
+  ): void;
+  createQueryPartitions(
+    query: string | ExecuteSqlRequest,
+    cb?: CreateQueryPartitionsCallback,
+  ): void | Promise<CreateQueryPartitionsResponse> {
     if (is.string(query)) {
```
surbhigarg92, line 151: instead use 

```
const request: ExecuteSqlRequest =
  typeof query === 'string' ? { sql: query } : query;
```

surbhigarg92: You can use any other variable name if required. In the current implementation we are doing the type cast for every reference `query as ExecuteSqlRequest`

src/batch-transaction.ts
```diff
@@ -157,7 +182,7 @@ class BatchTransaction extends Snapshot {
             client: 'SpannerClient',
             method: 'partitionQuery',
             reqOpts,
-            gaxOpts: query.gaxOptions,
+            gaxOpts: (query as ExecuteSqlRequest).gaxOptions,
```
surbhigarg92, line 185: remove the type cast once above recommendation is implemented

surbhigarg92:
`execute` method should also be modified to support both promise and callback. https://github.com/googleapis/nodejs-spanner/blob/main/src/batch-transaction.ts#L329

src/batch-transaction.ts
alkatrivedi: thanks for the suggestion, I have updated the code!

src/batch-transaction.ts
alkatrivedi: done

surbhigarg92 (approved):

PR #2293, opened 2025-04-29
Try deleting bad region tag in spanner_executor_proxy_client.ts

suztomo:
Dry deleting the bad region tag.


suztomo:
It turned out that this code is part of generated code.

suztomo:
b/414789367

PR #2296, opened 2025-05-02
test: skip backup tests to run in presubmits

alkatrivedi:
this code change will skip the backup tests to run against presubmit kokoro pipeline


owlbot.py
```diff
@@ -64,7 +64,7 @@
 
 common_templates = gcp.CommonTemplates()
 templates = common_templates.node_library(source_location='build/src')
-s.copy(templates, excludes=[".kokoro/samples-test.sh", ".kokoro/trampoline_v2.sh", ".github/release-trigger.yml", ".github/sync-repo-settings.yaml"])
+s.copy(templates, excludes=[".kokoro/samples-test.sh", ".kokoro/trampoline_v2.sh", ".github/release-trigger.yml", ".github/sync-repo-settings.yaml", ".kokoro/presubmit/node18/system-test.cfg"])
```
surbhigarg92, line 67: Why are we skipping this ? If you have made changes in google3 , the changes should be allowed ?

sakthivelmanii (approved):

system-test/spanner.ts
```diff
@@ -55,6 +55,11 @@ import Long = util.Long;
 const fs = require('fs');
 
 const SKIP_BACKUPS = process.env.SKIP_BACKUPS;
+const KOKORO_JOB = process.env.KOKORO_JOB_NAME;
+const presubmit_system_test =
```
sakthivelmanii, line 59: just check if KOKORO_JOB_NAME includes `presubmit` and not null

alkatrivedi: done, thanks for the suggestion!

owlbot.py
alkatrivedi: updated how we should skip backup tests

system-test/spanner.ts
```diff
@@ -55,6 +55,7 @@ import Long = util.Long;
 const fs = require('fs');
 
 const SKIP_BACKUPS = process.env.SKIP_BACKUPS;
+const KOKORO_JOB = process.env.KOKORO_JOB_NAME;
```
sakthivelmanii, line 58: `KOKORO_JOB_NAME` as variable name makes more sense

surbhigarg92 (approved):

sakthivelmanii (approved):

PR #2299, opened 2025-05-04
chore: remove execa

alkatrivedi:


alkatrivedi:
Closing this PR since

The `spawn()` returns a child process, not a promise, which means it is not directly await-able. If we want to use `await` with `spawn()` - similar to how we use it with `execa()` - we need to manually wrap it in a promise to handle asynchronous behavior and process compilation.

Alternatively, we could use `spawnSync()` for synchronous execution, but this comes with a major drawback: it blocks the main thread, including the Nodejs event loop, which can severely impact performance and responsiveness.

In summary:

- `spawn()` -> Non-blocking, runs in background, but not await-able without a wrapper.
- `spawnSync()` -> Synchronous and blocks the main thread, including timers, I/O and all async operations.

cc: @surbhigarg92

PR #2300, opened 2025-05-05
feat: add promise based signatures on createReadPartitions

alkatrivedi:


src/batch-transaction.ts
```diff
@@ -284,7 +292,17 @@ class BatchTransaction extends Snapshot {
...
+  createReadPartitions(
+    options: ReadRequest,
+  ): Promise<CreateReadPartitionsResponse>;
+  createReadPartitions(
+    options: ReadRequest,
+    callback: CreateReadPartitionsCallback,
+  ): void;
+  createReadPartitions(
+    options: ReadRequest,
+    cb?: CreateReadPartitionsCallback,
```
surbhigarg92, line 304: why the parameter name is different here. In above signature its callback , here its cb. How do we know in method which variable has to be used ?
Same is applicable for createQueryPartitions

src/batch-transaction.ts
```diff
@@ -321,7 +339,7 @@ class BatchTransaction extends Snapshot {
             }
 
             span.end();
-            callback(err, partitions, resp);
+            cb!(err, partitions, resp);
```
surbhigarg92, line 342: Wny cb , why not callback ?

alkatrivedi: it can be callback as well,  just to make code cleaner I used cb here

we already are using such idiomatic in our existing code as well [ref](https://github.com/googleapis/nodejs-spanner/blob/main/src/database.ts#L2886)

src/batch-transaction.ts
surbhigarg92: Understood, thanks for clarification

src/batch-transaction.ts
surbhigarg92: got it.

surbhigarg92 (approved):

PR #2322, opened 2025-06-12
chore: add commit retry logic

alkatrivedi:
This PR contains code changes for adding commit retry logic.

Use Case - Read/Queries + Mutations

In case mux session is enabled and a read/write transaction is executed with the above use case. commit could return a commit response containing 'MultiplexedSessionRetry' field. In this scenario, commit will get retry only once with the help of implemented retry logic.


surbhigarg92 (approved):
@rahul2393 Please want a second pair of eyes on this PR, I might have missed any MUX case.

rahul2393 (approved):

PR #2366, opened 2025-07-30
chore: test backups systes tests

surbhigarg92:


system-test/spanner.ts
```diff
@@ -3498,10 +3461,12 @@ describe('Spanner', () => {
     };
 
     it('GOOGLE_STANDARD_SQL should return error for backup expiration time in the past', async () => {
+      assert.ok(googleSqlBackupCreated);
```
alkatrivedi, line 3464: why this assertion here?

system-test/spanner.ts
```diff
@@ -3460,6 +3410,13 @@ describe('Spanner', () => {
     };
 
     it('GOOGLE_STANDARD_SQL should have completed a backup', async () => {
+      // Create backups.
+      await createBackup(
```
alkatrivedi, line 3414: backup creation can be in before block?

system-test/spanner.ts
```diff
@@ -3743,14 +3697,6 @@ describe('Spanner', () => {
...
     };
 
-    it('GOOGLE_STANDARD_SQL should delete backup', async () => {
-      await deleteBackup(googleSqlBackup2);
-    });
-
-    it.skip('POSTGRESQL should delete backup', async () => {
-      await deleteBackup(postgreSqlBackup2);
-    });
-
```
alkatrivedi, line 3753: instead of deleting, we can skip pushing backups in RESOURCES_TO_CLEAN and enable this test?

system-test/spanner.ts
```diff
@@ -3560,24 +3525,13 @@ describe('Spanner', () => {
           backup => backup.formattedName_ === googleSqlBackup1.formattedName_,
         ),
       );
-      assert.ok(
-        page3.find(
-          backup => backup.formattedName_ === googleSqlBackup2.formattedName_,
-        ),
-      );
       if (!IS_EMULATOR_ENABLED && !SKIP_POSTGRESQL_BACKUP_TESTS) {
```
alkatrivedi, line 3528: can we remove this condition (!IS_EMULATOR_ENABLED) from here? we anyway have below check in before block
```
if (IS_EMULATOR_ENABLED) {
  this.skip();
}
```
hence, if this test is running that means emulator is not enabled?

system-test/spanner.ts
surbhigarg92: no, we can't do that as these backups are in use by restore operations and databases created with restore operations, so we would have to delete those databases first

system-test/spanner.ts
```diff
@@ -3375,58 +3364,17 @@ describe('Spanner', () => {
...
-
+        await postgreSqlDatabase1.table(TABLE_NAME).insert({
+          SingerId: generateName('id'),
+          Name: generateName('name'),
+        });
         postgreSqlBackup1 = instance.backup(postgreSqlBackup1Name);
-        postgreSqlBackup2 = instance.backup(postgreSqlBackup2Name);
-
-        RESOURCES_TO_CLEAN.push(...[postgreSqlBackup1, postgreSqlBackup2]);
+        RESOURCES_TO_CLEAN.push(...[postgreSqlBackup1]);
```
alkatrivedi, line 3377: similary, we can delete this as well

alkatrivedi (approved):

PR #2369, opened 2025-08-04
fix: race condition among transactions when running parallely

### LINKED ISSUE #2368: Race Condition in Multiplexed Session
Transaction gets assigned to the session as per this [ref](https://github.com/googleapis/nodejs-spanner/blob/main/src/multiplexed-session.ts#L194).

**Issue:**
In a multiplexed architecture, a single session is reused to handle multiple transactions, creating a one-to-many relationship. While efficient, this model introduces a critical problem in a multithreaded environment.

The core issue is a race condition. When multiple threads operate in parallel, they share the same session object. As each thread initiates a new transaction, it overwrites a common transaction field (txn) within that shared session object.

Consequently, operations from one thread can be mistakenly executed within the context of a transaction started by a different thread. This leads to severe errors, including:

**Incorrect transaction execution:** Logic intended for one transaction is applied to another.

**Erroneous commits:** Data is committed to the database under the wrong transaction, leading to data corruption and inconsistent state within the client.

**Steps to reproduce:**
Run below script on multiple threads in parallel

```
  const [rows] = await tx.run('SELECT * FROM Singers');
  rows.forEach(row => {
    const json = row.toJSON();
    console.log(`SingerId: ${json.SingerId}, FullName: ${json.FullName}`);
  });
  const id = Math.floor(Math.random() * 10000) + 1;
  const name = randomUUID();
  tx.upsert('Singers', [{SingerId: id, FirstName: name}]);
  await tx.commit();
  console.log('transaction done.');
}

```
cc: @surbhigarg92

alkatrivedi:
fixes: #2368

### DIFF
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -9191,7 +9191,7 @@ describe('Spanner', () => {
         commitTransaction(done, PG_DATABASE, postgreSqlTable);
       });
 
-      describe('when multiplexed session is enabled for read write', async () => {
+      describe('parallel transactions', async () => {
         async function insertAndCommitTransaction(database, sync, table, key) {
           await database.runTransactionAsync(async transaction => {
             // read from table TxnTable
@@ -9218,7 +9218,7 @@ describe('Spanner', () => {
           });
         }
 
-        it('GOOGLE_STANDARD_SQL should insert and commit transaction when running parallely', async () => {
+        it('should insert and commit transaction when running parallely', async () => {
           const promises: Promise<void>[] = [];
           let resolvePromise;
           const commitPromise = new Promise(
@@ -9236,40 +9236,6 @@ describe('Spanner', () => {
           );
           promises.push(
             insertAndCommitTransaction(DATABASE, sync, googleSqlTable, 'k1101'),
-          );
-
-          // wait for both the transactions to complete their execution
-          await Promise.all(promises);
-        });
-
-        it('POSTGRESQL should insert and commit transactions when running parallely', async () => {
-          const promises: Promise<void>[] = [];
-          let resolvePromise;
-          const commitPromise = new Promise(
-            resolve => (resolvePromise = resolve),
-          );
-          const sync = {
-            target: 2, // both the transactions to be ready
-            count: 0, // 0 transactions are ready so far
-            promise: commitPromise, // the promise both the transactions wait at
-            resolveCommitPromise: () => resolvePromise(), // the function to resolve the commit promise
-          };
-          // run the transactions in parallel
-          promises.push(
-            insertAndCommitTransaction(
-              PG_DATABASE,
-              sync,
-              postgreSqlTable,
-              'k1102',
-            ),
-          );
-          promises.push(
-            insertAndCommitTransaction(
-              PG_DATABASE,
-              sync,
-              postgreSqlTable,
-              'k1103',
-            ),
           );
 
           // wait for both the transactions to complete their execution
--- a/system-test/spanner.ts
+++ b/system-test/spanner.ts
@@ -9192,37 +9192,28 @@ describe('Spanner', () => {
       });
 
       describe('when multiplexed session is enabled for read write', async () => {
-        before(() => {
-          process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS = 'true';
-          process.env.GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS_FOR_RW = 'true';
-        });
-
-        after(() => {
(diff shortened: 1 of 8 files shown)


system-test/spanner.ts
```diff
@@ -9191,6 +9191,87 @@ describe('Spanner', () => {
...
+              StringValue: 'v6',
+            });
+
+            // increment the shared counter
+            sync.count++;
+
+            while (sync.count !== sync.target) {
+              // wait for 500 milliseconds
+              await new Promise(resolve => setTimeout(resolve, 500));
+            }
```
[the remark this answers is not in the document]
alkatrivedi: thanks for this catch, I have updated the test to execute a read before mutations. now we will have a precommit token while committing the transactions

test/multiplexed-session.ts
```diff
@@ -213,12 +213,11 @@ describe('MultiplexedSession', () => {
 
     it('should pass back the session and txn', done => {
       const fakeTxn = new FakeTransaction() as unknown as Transaction;
```
sakthivelmanii, line 215: where are we mocking `session.!transaction` method? Ideally this test should fail since we removed the assignment right?

system-test/spanner.ts
```diff
@@ -9191,6 +9191,92 @@ describe('Spanner', () => {
         commitTransaction(done, PG_DATABASE, postgreSqlTable);
       });
 
+      describe('when multiplexed session is enabled for read write', async () => {
```
surbhigarg92, line 9194: This need not be specific to multiplexed session, the scenario we are trying to test here is that parallel transaction should work , for both mux and non-mux. So describe should mention that.

test/spanner.ts
```diff
@@ -4734,6 +4734,65 @@ describe('Spanner with mock server', () => {
...
+          );
+
+          // assert that precommitToken is different in both the commit request
+          assert.notEqual(
+            commitRequest[0].precommitToken.precommitToken,
+            commitRequest[1].precommitToken.precommitToken,
+          );
+
+          // assert that seqNum is different in both the commit request
+          assert.notEqual(
```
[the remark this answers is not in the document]
surbhigarg92: I guess because mockspanner return a random number for sequence number , randomInt(1, 1000),

But thats not a valid assertion. SeqNum may or maynot be same for 2 transactions running in parallel. 

I am not sure if the `seqNum` is supposed to be an incremental number from 1...n . If yes, then mockSpanner should have followed the same logic.

test/multiplexed-session.ts
alkatrivedi: we have removed the assignment but we are returning the transaction using `transaction` method defined in Session class.

session!.transaction((session!.parent as Database).queryOptions_),

test/spanner.ts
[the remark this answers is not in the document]
alkatrivedi: >> I am not sure if the seqNum is supposed to be an incremental number from 1...n

yes, I have updated the mock spanner logic and hence, the test case to run read before performing mutations. now we will be sure that precommit token is there for both the transactions while making commit request.

test/spanner.ts
```diff
@@ -4671,7 +4671,7 @@ describe('Spanner with mock server', () => {
             MultiplexedSessionRetry: 'precommitToken',
             precommitToken: {
               precommitToken: fakeRetryToken,
-              seqNum: 1,
+              seqNum: 100,
```
surbhigarg92, line 4674: I dont understand why are there changes in this test, why should code changes impact this?

test/spanner.ts
```diff
@@ -4715,13 +4715,7 @@ describe('Spanner with mock server', () => {
               2,
               'The mock commit method should have been called exactly twice.',
             );
-            const firstRequest = capturedCommitRequests[0];
```
surbhigarg92, line 4718: Why is this removed ?

test/mockserver/mockspanner.ts
```diff
@@ -1022,7 +1025,7 @@ export class MockSpanner {
             const precommitToken = session?.multiplexed
               ? protobuf.MultiplexedSessionPrecommitToken.create({
                   precommitToken: Buffer.from('mock-precommit-token'),
-                  seqNum: randomInt(1, 1000),
+                  seqNum: this._seqNum++,
```
surbhigarg92, line 1028: seqNum should increase per transaction not per mockSpanner object

test/spanner.ts
alkatrivedi: since, now we are returning precommit token in the response of executeStreamingSql. the precommit token is not null here

surbhigarg92 (approved):

Issue #2373, opened 2025-08-08
Provide a way to programmatically disable builtin metrics


flovouin:
### A screenshot that you have tested with "Try this API".

N/A

### What would you like to see in the library?

From what I understand, the feature introduced by https://github.com/googleapis/nodejs-spanner/pull/2272 can only be disabled using either the `SPANNER_DISABLE_BUILTIN_METRICS` or `SPANNER_EMULATOR_HOST` environment variable is set.

I'm programmatically creating a client to an emulator, using something like:

```typescript
const spanner = new Spanner({
  servicePath: 'some-address',
  port: SPANNER_GRPC_PORT,
  projectId: 'demo-some-project',
  sslCreds: grpc.credentials.createInsecure(),
});
```

This doesn't get detected by the metrics initialisation logic, and I end with logs like:

```
Need monitoring metric writer permission on project demo-some-project. Follow https://cloud.google.com/spanner/docs/view-manage-client-side-metrics#access-client-side-metrics to set up permissions
ERROR: Send TimeSeries failed: 7 PERMISSION_DENIED: Permission monitoring.timeSeries.create denied (or the resource may not exist).
```

I'm not sure if there's a reliable way of determining that the metrics should not be exported based on the configuration I'm passing to `Spanner`. If there is, then great, but I don't mind passing an explicit parameter like `disableBuiltinMetrics: true` either.

### Describe alternatives you've considered

Using environment variables, which in my case would require manipulating `process.env` as I'm configuring the client programmatically.

PR #2380, opened 2025-08-13
fix: provide option to disable built in metrics

surbhigarg92:
#2373

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -150,6 +150,7 @@ export type GetInstanceConfigOperationsCallback = PagedCallback<
  * Indicates which replicas or regions should be used for non-transactional reads or queries.
  * DirectedReadOptions won't be set for readWrite transactions"
  * @property {ObservabilityOptions} [observabilityOptions] Sets the observability options to be used for OpenTelemetry tracing
+ * @property {boolean} [disableBuiltInMetrics=True] If set to true, built-in metrics will be disabled.
  */
 export interface SpannerOptions extends GrpcClientOptions {
   apiEndpoint?: string;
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -1614,10 +1616,11 @@ class Spanner extends GrpcService {
   /**
    * Setup the OpenTelemetry metrics capturing for service metrics to Google Cloud Monitoring.
    */
-  configureMetrics_() {
+  configureMetrics_(disableBuiltInMetrics?: boolean) {
     const metricsEnabled =
       process.env.SPANNER_DISABLE_BUILTIN_METRICS !== 'true' &&
-      !this._isEmulatorEnabled;
```
alkatrivedi, line 1620: can we remove this variable?

surbhigarg92: Changed the logic a bit, PTAL

src/index.ts
```diff
@@ -443,14 +445,12 @@ class Spanner extends GrpcService {
       );
     }
 
-    let isEmulatorEnabled = false;
```
sakthivelmanii, line 446: why are we changing it?

test/metrics/metrics.ts
```diff
@@ -209,6 +209,7 @@ describe('Test metrics with mock server', () => {
       spannerMock.resetRequests();
       spannerMock.removeExecutionTimes();
       // Reset the MetricsFactoryReader to an in-memory reader for the tests
+      MetricsTracerFactory.enabled = true;
```
alkatrivedi, line 212: why are enabling it here now? we already are enabling it to true in before block of parent

surbhigarg92: because in after each block we are resetting MetricsTracerFactory, earlier it was always default enabled , but now with mockSpanner since it is using inSecureCredentials MetricsTracerFactory.enabled is false.

alkatrivedi (approved):

PR #2480, opened 2025-12-09
feat: add option to opt out of  SPANNER_EMULATOR_HOST

alkatrivedi:
This adds a new `ignoreEnvSpannerEmulatorHost` option to the `SpannerOptions`. This new option allows users to explicitly disable the automatic detection of the emulator environment variable, which is useful when integrating with configuration providers or running hybrid tests


alkatrivedi:
closing this PR as we are not moving forward with this approach

